PR #50, opened 2020-05-20
Show backtrace by default if CLI crashes

### LINKED ISSUE #47: Always show full tracebacks if a cli command is crashed
I'm not sure if this has been resolved yet or not, but we should always show a full traceback in case a CLI command is crashed. At least until the CLI is stable.

1st1:
Fixes #47


1st1:
Not sure if this is the most elegant way but it works on my machine

Issue #15, opened 2020-03-24
In repl implement variables of non-string types


tailhook:
Currenly, only strings are accepted

*(hint: you can cast strings to other types in the edgeql itself)*

1st1:
> (hint: you can cast strings to other types in the edgeql itself)

Maybe this should be the preferred way of how we implement the feature -- add a compiler flag to accept all arguments as text, basically implicitly compiling `SELECT <int32>$0` as `SELECT <int32><str>$0`

elprans:
We already have `json_parameters`, wouldn't that work?

1st1:
Another question: how do we input strings?  Currently there's no requirement for quotes, e.g.

`Parameter <str>$foo>  blah blah`

translates into `"blah blah"` string.

This is problematic as it's unclear how to enter an array of strings.

IMO, while it's slightly inconvenient, we should require the user to enter EdgeQL literals, so string/json/datetime quotes should be required.

elprans:
IMO, they should be proper literals.

1st1:
> IMO, they should be proper literals.

Yep. And then we can syntax highlight the input.

tailhook:
That sounds like over-engineering: the primary use case for the feature is to quickly past query from the code an fill in the gaps. It's unlikely that there would be any complex texts that can't be typed directly. If they are in 1% of cases, you can just edit the original query and put complex constant there.

That said, we can do something like this, if you think this is a crucial feature:
```
Parameter <str>$var (use Ctrl+P for quoted string):
```

tailhook:
> Maybe this should be the preferred way of how we implement the feature -- add a compiler flag to accept all arguments as text

As I've said in the zoom, we can have much shorter feedback loop if parser/validate things directly in repl: we can highlight invalid value as you type, and show error and repeat input right after return key. In case, you have a query with more than one parameter it could be a huge time saver.

1st1:
> That sounds like over-engineering: the primary use case for the feature is to quickly past query from the code an fill in the gaps.

Yes, but if the query has an `<array<str>>$param` it becomes unclear how to enter it. The uncertainty goes away if we stick with EdgeQL literals syntax.  It's not over-engineering, it's consistent design.

> Parameter <str>$var (use Ctrl+P for quoted string):

Not sure I understand the idea. Do you want to add quotes/escapes once Ctrl+P is pressed?  I'm not sure we need this unless a lot of users request that.

[...]

PR #51, opened 2020-05-20
repl: Implement int16/32/64 input types for query parameters

1st1:


1st1:
Maybe we should disable this feature as many datatypes aren't implemented

src/variables.rs
```diff
@@ -55,16 +55,45 @@ async fn input_item(name: &str, mut item: &Descriptor, all: &InputTypedesc,
...
+            };
+
+            match s.id {
+                codec::STD_STR => {
+                    Ok(Value::Str(val))
+                }
+                codec::STD_INT16 => {
+                    match val.parse::<i16>() {
+                        Ok(v) => Ok(Value::Int16(v)),
+                        Err(_) => Err(anyhow::anyhow!("invalid int16 value"))
```
tailhook, line 80: This is not helpful enough, rust is able to show more specific error. It should work by providing context:
```use anyhow::Context;
val.parse::<i16>().context("invalid int16 value")?;
```

1st1: I've implemented the suggestion, however the error output doesn't necessarily look better now:

```
yury> select <int16>$a;
Parameter <int16>$a: 1231231231
Error {
    context: "invalid int16 value",
    source: ParseIntError {
        kind: Overflow,
    },
}
```

Is that what you expected?

tailhook: The error is fine. The following should fix error formatting:
```diff
diff --git a/src/client.rs b/src/client.rs
index 80c2acd..64967d3 100644
--- a/src/client.rs
+++ b/src/client.rs
@@ -521,7 +521,7 @@ async fn _interactive_main(
             let input = match input_variables(&indesc, state).await {
                 Ok(input) => input,
                 Err(e) => {
-                    eprintln!("{:#?}", e);
+                    eprintln!("{:#}", e);
                     state.last_error = Some(e);
                     continue 'statement_loop;
                 }
```
Also we might want to loop inside this function until the value is valid or Ctrl+C. But this can be in another commit.

1st1: Nice! The error output is definitely better now. I'll merge the PR.

> Also we might want to loop inside this function until the value is valid or Ctrl+C. But this can be in another commit.

I agree. Let's first decide on https://github.com/edgedb/edgedb-cli/issues/15.

Thanks for the review

PR #58, opened 2020-05-27
Read password from EDGEDB_PASSWORD

### LINKED ISSUE #57: Add ability to get password from EDGEDB_PASSWORD env var
This is required for things like launching interactive repl from Python processes and will also simplify deployment in some cases. Our edgedb-python & edgedb-js drivers have this.

1st1:
Fixes #57.


1st1:
@tailhook The CI seems to be broken for PRs.

@elprans 
> (also, perhaps we should replace --password-from-stdin with a more generic --password-from-fd to allow passing passwords via an alternate file descriptor).

Can you open an issue for A4?

PR #73, opened 2020-06-02
Server refactoring and partial implementation of `server init`

tailhook:


Cargo.toml.orig
```diff
@@ -0,0 +1,83 @@
+[package]
```
elprans, line 1: You probably didn't intend to commit this.

tailhook: Nice catch. Thanks!

src/server/linux.rs
```diff
@@ -0,0 +1,268 @@
...
+Type=notify
+
+Environment=EDGEDATA={directory}
+RuntimeDirectory=edgedb-{instance_name}
+
+ExecStart={server_path} -D ${{EDGEDATA}} --runstate-dir=%t/edgedb-{instance_name}
+ExecReload=/bin/kill -HUP ${{MAINPID}}
+KillMode=mixed
+KillSignal=SIGINT
+TimeoutSec=0
```
fantix, line 244: Looks like this will be a problem on distros like CentOS 8 with [systemd >= 229](https://github.com/systemd/systemd/blob/a2031de849da52aa85b7e4326c0112ed7e5b5672/NEWS#L6440-L6444) if you meant to turn off the timeout here.

tailhook: > To maintain backwards compatibility, "0" continues turn off previously existing timeout settings.

I think this is previously existing timeout, so fine. I think we need to support older systemd for now, but I did no research (i.e. we only need to support as old as systemd on a system we have packages for).

Issue #74, opened 2020-06-02
Implement `rustup-init.sh`-style installer script for the CLI.


elprans:
The CLI itself should be trivially installable via a simple command.  `rustup-init` (and `rustup-init.sh`) is a great example of bootstrapping a CLI.

PR #75, opened 2020-06-03
First version of edgedb-init

tailhook:
Mostly similar to Rust implementation, except:
1. On linuxes we install directly into `~/.local/bin` since we have only one binary
2. When run as root CLI is installed directly into `/usr/bin` instead of bailing out
3. On MacOS we install into `/usr/local` (untested yet)
4. Windows support is incomplete yet

#74

### DIFF
--- a/edgedb-init.sh
+++ b/edgedb-init.sh
@@ -1,8 +1,11 @@
 #!/bin/bash
 
-# This is just a little script that can be downloaded from the internet to
-# install eddgedb CLI tools. It just does platform detection, downloads the
-# installer and runs it.
+# Portions Copyright (c) 2020 MagicStack Inc.
+# Portions Copyright (c) 2016 The Rust Project Developers.
+#
+# This is a simple script that can be downloaded from https://edgedb.com to
+# install `edgedb` CLI tools. Its job is to detect the host platform and to 
+# download and run the relevant installer.
 
 set -u
 
--- a/edgedb-init.sh
+++ b/edgedb-init.sh
@@ -96,7 +96,7 @@ main() {
     ensure chmod u+x "$_file"
     if [ ! -x "$_file" ]; then
         printf '%s\n' "Cannot execute $_file (likely because of mounting /tmp as noexec)." 1>&2
-        printf '%s\n' "Please copy the file to a location where you can execute binaries and run ./edgedb-cli${_ext} _init." 1>&2
+        printf '%s\n' "Please copy the file to a location where you can execute binaries and run ./edgedb-cli${_ext} _self_install." 1>&2
         exit 1
     fi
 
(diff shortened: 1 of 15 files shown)


edgedb-init.sh
```diff
@@ -0,0 +1,269 @@
+#!/bin/bash
```
elprans, line 1: Please don't forget to credit `rustup-init`, since this is derived from `rustup-init.sh`.

edgedb-init.sh
```diff
@@ -0,0 +1,269 @@
...
+
+# This is just a little script that can be downloaded from the internet to
+# install eddgedb CLI tools. It just does platform detection, downloads the
+# installer and runs it.
```
1st1, line 5: I'd adjust this comment to something like:

```suggestion
# This is a simple script that can be downloaded from https://edgedb.com to
# install `edgedb` CLI tools. Its job is to detect the host platform and to 
# download and run the relevant installer.
#
# (Some parts of this script are inspired by the great rustup-init.sh.)
```

edgedb-init.sh
1st1: I think something like what I suggested in https://github.com/edgedb/edgedb-cli/pull/75/files#r435523120 would be enough.

edgedb-init.sh
```diff
@@ -0,0 +1,269 @@
...
+
+set -u
+
+EDGEDB_PKG_ROOT="${EDGEDB_PKG_ROOT:-https://packages.edgedb.com}"
+
+usage() {
+    cat 1>&2 <<EOF
+edgedb-init
+The installer for edgedb command-line tools
```
1st1, line 14:
```suggestion
The installer for edgedb CLI.
```

edgedb-init.sh
elprans: Well, some things here are copied verbatim, including some comments, so it's not "inspired", it's literally based on (parts of) their code, which is totally OK, but we must be diligent in placing attribution, hence I'd put an explicit "Portions Copyright (c) 2016 The Rust Project Developers"

1st1: Makes sense. I've amended my suggestion below.

edgedb-init.sh
tailhook: Why? I see general attitude to avoid abbreviations in open-source community. We're not limited in the space here.

edgedb-init.sh
tailhook: Sure. Just wasn't sure how to mention this.

edgedb-init.sh
1st1: Simply because I used "CLI" in other suggested edit for the same file. You can disregard this, the current text works too.

PR #85, opened 2020-07-09
Bump version to 1.0.0-alpha.4

elprans:
This is mostly to unbreak the nighlies


tailhook:
Also added Cargo.lock

PR #86, opened 2020-07-10
repl: Implement `--instance` argument

tailhook:


src/options.rs
```diff
@@ -193,7 +198,15 @@ impl Options {
         tmp.wait_until_available.map(|w| conn_params.wait_until_available(w));
 
         let unix_host = host.contains("/");
-        if admin || unix_host {
+        if let Some(name) = tmp.instance {
+            let path = server::get_instance(&name)
+                .and_then(|inst| inst.get_socket(admin))
+                .unwrap_or_else(|e| {
+                    eprintln!("Failed to find instance {:?}: {:#}", name, e);
```
elprans, line 205: s/Failed/Could not/ and I'd also add a hint that instances are expected to be created using `edgedb server init`.

elprans:
Having thought about this a bit more, I withdraw my initial objection expressed in the chat. As long as there's sufficient explanation of what `-I` is and how it differs from explicit connection options, I'm fine with adding it.

PR #98, opened 2020-07-15
Move determine-version.py to edgedb-pkg

ambv:
This is a better central place for this, and local scripts work only when a repository is checked out anyway.


chrisrap232:
Hello

PR #99, opened 2020-07-16
Rephrase CLI post-install message

ambv:


src/self_install.rs
```diff
@@ -145,28 +145,29 @@ fn ensure_line(path: &PathBuf, line: &str) -> anyhow::Result<()> {
 
 fn print_post_install_message(settings: &Settings) {
     if cfg!(windows) {
-        print!(r###"# EdgeDB command-line tool is installed now. Great!
+        print!(r###"# EdgeDB command-line tool is installed now!
 
-To get started you need installation directory ({dir}) in your `PATH`
-environment variable. Future applications will automatically have the
-correct environment, but you may need to restart your current shell.
+We've updated your bashrc to have {dir} in your `PATH`
```
elprans, line 150: s/bashrc/shell configuration/, because we're not actually updating .bashrc

ambv: "shell profile"?

ambv: Right, that's Windows. So, "environment configuration"?

1st1: +1 for "environment configuration"

PR #103, opened 2020-07-17
Use revision when comparing versions

### LINKED ISSUE #100: `edgedb server install` ignores the `revision` field in .jsonindexes
- EdgeDB Version: 1.0a4
- EdgeDB CLI Version: 1.0a4
- OS Version: macOS 10.15

`edgedb server install` ignores the `revision` field in the .jsonindexes file, causing it to choose an outdated 1.0a4 build instead of the newest one.

Attached jsonindexes demonstrate the issue: [jsonindexes.zip](https://github.com/edgedb/edgedb-cli/files/4933602/jsonindexes.zip)

I'm interested in solving this with Paul's help.

tailhook:
Fixes #100

### DIFF
--- a/src/server/package.rs
+++ b/src/server/package.rs
@@ -96,19 +96,20 @@ fn version_matches(package: &PackageInfo, version: &VersionQuery) -> bool {
 pub fn find_version(haystack: &RepositoryInfo, ver: &VersionQuery)
     -> Result<VersionResult, anyhow::Error>
 {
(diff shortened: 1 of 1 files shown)


elprans:
Could we add some tests for this?

tailhook:
We can probably feed some synthetic data to `find_version` function. So yes. But I'm out of time slots to do that before the vacation :palm_tree:

ambv:
I'll add tests for this in a separate PR.

PR #105, opened 2020-07-21
Install explicit Rust toolchain also during nightlies and releases

ambv:
This is an equivalent of edgedb/edgedb#1564

### DIFF
--- a/.github/workflows.src/nightly.tpl.yml
+++ b/.github/workflows.src/nightly.tpl.yml
@@ -15,6 +15,13 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Install Rust toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        default: true
+
     - name: Build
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -18,7 +18,6 @@ jobs:
         with:
           profile: minimal
           toolchain: stable
-          target: x86_64-unknown-linux-musl
```
ambv, line 21: I also removed this line since it's clearly a bad copypasta. There is a musl-specific test below in the file which needs a musl target. This one here runs on a matrix of operating systems, not all of which are Linux.

.github/workflows.src/nightly.tpl.yml
```diff
@@ -15,6 +15,13 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Install Rust toolchain
```
elprans, line 18: Technically, we don't need the up-to-date rust toolchain for Linux, because we run the build in a container, which already has it.

ambv: That's right, I'll remove that then for clarity. The edgedb/edgedb equivalent isn't installing the Rust toolchain on Linux either.

PR #106, opened 2020-07-21
Explain how to run without Vagga (e.g. on macOS)

ambv:


tailhook:
Maybe we should default EDGEDB_MAJOR_VERSION? Other variables are optional, we can make a section `environment` to describe them, including both linux and macos example values.

elprans:
Question: why are the environment variables resolved at compile-time and not run-time?

tailhook:
> Question: why are the environment variables resolved at compile-time and not run-time?

For PSQL_DEFAULT_PATH that is intention, you build it using `setup.py` and which sets the env variable and then run the tool without caring of environment.

For `EDGEDB_MAJOR_VERSION` that's probably was lazy implementation, as generally I'm assuming that build-and-run is a single step using `cargo test` in 99% use cases.

(`PYTHONWARNINGS` is obviously a runtime thing)

1st1:
FWIW I failed to run local tests for https://github.com/edgedb/edgedb-cli/pull/108. I'll take another look soon. But I'd appreciate if you could add a `make` target for poor souls like me who are on macOS and have no vagga.

> For EDGEDB_MAJOR_VERSION that's probably was lazy implementation

Maybe we should make it check at runtime.

tailhook:
> > For EDGEDB_MAJOR_VERSION that's probably was lazy implementation
> 
> Maybe we should make it check at runtime.

I've changed that. Also added default value.

> FWIW I failed to run local tests for #108. I'll take another look soon. But I'd appreciate if you could add a `make` target for poor souls like me who are on macOS and have no vagga.
> 

`cargo test` is the command line. You don't need Makefile. It would be helpful if you show what error you get.

1st1:
Closing this, as developing without vagga now doesn't require anything special (that's how I develop).

PR #107, opened 2020-07-21
Add test for find_version

ambv:
I have no idea what I'm doing wrt the string ownership here. I also didn't know what the idiomatic way to fail on an Err match would be in Rusts tests.

### DIFF
--- a/src/server/package.rs
+++ b/src/server/package.rs
@@ -132,3 +132,55 @@ impl PackageInfo {
         Version(format!("{}-{}", self.version, self.revision))
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::RepositoryInfo;
+    use super::Version;
+    use super::VersionQuery;
+    use super::VersionResult;
+    use super::find_version;
+
+    #[test]
+    fn test_find_version() {
+        let json_contents = r#"
+            {"packages": [{"architecture": "x86_64",
+                           "basename": "edgedb-server",
+                           "installref": "/archive/macos-x86_64/edgedb-server-1-alpha3_1.0a3_2020060201.pkg",
+                           "name": "edgedb-server-1-alpha3",
+                           "revision": "2020060201",
+                           "slot": "1-alpha3",
+                           "version": "1.0a3"},
+                          {"architecture": "x86_64",
+                           "basename": "edgedb-server",
+                           "installref": "/archive/macos-x86_64/edgedb-server-1-alpha4_1.0a4_2020071512.pkg",
+                           "name": "edgedb-server-1-alpha4",
+                           "revision": "2020071512",
+                           "slot": "1-alpha4",
+                           "version": "1.0a4"},
+                          {"architecture": "x86_64",
+                           "basename": "edgedb-server",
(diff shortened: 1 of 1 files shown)


src/server/package.rs
```diff
@@ -132,3 +132,55 @@ impl PackageInfo {
...
+    use super::VersionQuery;
+    use super::VersionResult;
+    use super::find_version;
+
+    #[test]
+    fn test_find_version() {
+        let json_contents = r#"
+            {"packages": [{"architecture": "x86_64",
+                           "basename": "edgedb-server",
+                           "installref": "/archive/macos-x86_64/edgedb-server-1-alpha3_1.0a3_2020060201.pkg",
```
1st1, line 149: Please avoid breaking 79 characters per line when possible, like it is here.

src/server/package.rs
```diff
@@ -132,3 +132,55 @@ impl PackageInfo {
...
+                           "installref": "/archive/macos-x86_64/edgedb-server-1-alpha4_1.0a4_2020071614.pkg",
+                           "name": "edgedb-server-1-alpha4",
+                           "revision": "2020071614",
+                           "slot": "1-alpha4",
+                           "version": "1.0a4"}]}
+        "#;
+        let repository_info: RepositoryInfo = serde_json::from_str(json_contents).unwrap();
+        let version_str = format!("{}-{}", "1.0a4", "2020071614");
+        let query = VersionQuery::Stable(Some(Version(version_str.to_owned())));
+        let result = find_version(&repository_info, &query);
```
tailhook, line 172: `let actual = find_version(...).unwrap();` returns value inside `Ok(..)` and panics otherwise (panic is an expected way to fail a test). Actually you already do the same on `let repository_info` line.

src/server/package.rs
```diff
@@ -132,3 +132,55 @@ impl PackageInfo {
...
+        let expected = VersionResult {
+            package_name: String::from("edgedb-server"),
+            major_version: Version(version_str.to_owned()),
+            version: Version(version_str.to_owned()),
+            revision: String::from("2020071614"),
+        };
+        match result {
+            Ok(actual) => assert!(actual.package_name == expected.package_name &&
```
tailhook, line 180: You don't need a structure `expected`. Do you use it for type inference?

Also `assert_eq!` is better, than `assert` (i.e. shows a value). And you should be able to compare version to a `&str`, so:
```
{
assert_eq!(actual.version, "1.0a4-2020071614");
assert_eq!(actual.revision, "2020071614");
}
```
Should work.

src/server/package.rs
```diff
@@ -132,3 +132,55 @@ impl PackageInfo {
...
+                                  actual.version == expected.version &&
+                                  actual.revision == expected.revision),
+            Err(_) => (),
```
tailhook, line 183: Should be a panic with a formatted message:
```
Err(e) => panic!("error returned: {}", e),
```
But `unwrap` as mentioned above is enough in tests.

1st1:
@tailhook this one is fairly outdated -- should we close it or you can rebase it?

tailhook:
Looks like too off the current codebase. And I don't remember which bug this was supposed to be regression test for :(

PR #108, opened 2020-07-22
Use the new Prepare message header to render object type names.

1st1:
This obsoletes the `\set introspect_types on` command as object types
will always have their type names displayed.

See https://github.com/edgedb/edgedb/pull/1567 and
https://github.com/edgedb/edgedb/issues/1350 for details.


tailhook:
This is what I get with older server:
```
edgedb> SELECT schema::Object { name };
ERROR: BinaryProtocolError: unexpected message header: 65282
```
We promised not to break backwards compaitibility.

elprans:
> We promised not to break backwards compaitibility.

Yes, we should bump the protocol version.  There would me more changes (support for arguments in `execute`), so the timing is good.

1st1:
@tailhook @elprans I've rebased this branch and it seems to work fine with https://github.com/edgedb/edgedb/pull/1567.

Few things left to do:

* we need to implement protocol negotiation in edgedb-rust: https://github.com/edgedb/edgedb-rust/blob/07ebe8023308402e369484c3d0053a904872dbb6/edgedb-client/src/builder.rs#L285-L289. The idea is that the client should request the protocol version -- 0.9 from now on -- and the server will reply with the version it supports. We need to store that version somewhere to access it later.

* we then need to modify my PR to only inject the `QUERY_OPT_INLINE_TYPENAMES` header when the server support protocol 0.9.

@tailhook Paul, maybe you could help me with this? I'd like to switch to fixing "ports".

tailhook:
> @tailhook Paul, maybe you could help me with this? I'd like to switch to fixing "ports".

Sure

PR #122, opened 2020-08-12
Fix compiler errors after client extraction

CodesInChaos:
Not sure why it broke. Perhaps files you changed after I created my branch. Anyways, with these fixes it'll compile.


CodesInChaos:
Looks like you already fixed it

PR #123, opened 2020-08-12
run tests on pull request

CodesInChaos:
It'd be useful to run the CI on pull requests to prevent merging broken code.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,4 +1,4 @@
-on: [push]
+on: [push, pull_request]
 
 name: Tests
 env:


CodesInChaos:
Not sure if it's necessary to test the full OS matrix on pull requests

tailhook:
Tests do currently run on PRs. Where do you see the problem?

CodesInChaos:
I think they only run on pushes, not pull requests so they only run after you merged the pull request but not while it's open.

tailhook:
I see tests on your PR too, don't you have that:
image

CodesInChaos:
I think it uses the workflow config defined in the pull request.

Take a look at my pull requests to edgedb-rust:
* https://github.com/edgedb/edgedb-rust/pull/34 ran tests because it added the trigger to the workflow
* https://github.com/edgedb/edgedb-rust/pull/35 and https://github.com/edgedb/edgedb-rust/pull/33 didn't

(btw I don't see any CLA checks on edgedb-cli, is that set up correctly?)

tailhook:
Ah, right. I see now. Thanks!

tailhook:
> (btw I don't see any CLA checks on edgedb-cli, is that set up correctly?)

Probably not. Thanks for noticing it.

CodesInChaos:
Looks like this has the side effect that if the feature branch is in the main repo, the tests run twice (once triggered by `push`, once by `pull_request`. Perhaps it'd make sense to limit the push trigger to the `master` branch.

Issue #91, opened 2020-07-10
Initialize auth in `edgedb server init`


elprans:
`edgedb server init` should allow initializing the authentication configuration by asking for the superuser password interactively. The interactive prompt maybe turned off by passing `--no-pwprompt` (in which case `Trust` auth is configured) or substituted by a password from a file using `--pwfile=<file>`).

We can also add a way to reset the superuser password (using `--admin` connection) like `edgedb server auth` or something like that.

tailhook:
1. What is the name of that user? Is this `edgedb`?
2. What if stdin is not tty?
3. `--no-pwprompt` doesn't encompass `trust`, we should name it `--trust-without-password` or whatever, that shows that not only password prompt is not there on init, but we also allow everybody to login with any password
4. I still see a conflict with `--interactive`. Since `--interactive` is opt-in, I assume that non-interactive, scripting mode is default. This options ruins the assumption.

And a tiny nitpick: we have `--password` and even `--password-from-stdin`. So it should be `--password-file` (not `--pwfile`) for consistency.

elprans:
> What is the name of that user? Is this `edgedb`?

Yes, `edgedb`.

> What if stdin is not tty?

Assume `--no-pwprompt`

>  `--no-pwprompt` doesn't encompass `trust`, we should name it `--trust-without-password` or whatever, that shows that not only password prompt is not there on init, but we also allow everybody to login with any password

+1

> I assume that non-interactive, scripting mode is default

Perhaps we should make the interactive mode to be the default?  These commands are intended to be used interactively first and foremost.

> So it should be `--password-file` (not `--pwfile`) for consistency.

+1

tailhook:
> > What if stdin is not tty?
> 
> Assume `--no-pwprompt`

This is a very common use case for automatic installers to have stdin from dev null. And skipping auth in this case leaves servers open from outside world. (Relying that nobody will use `edgedb server init` on production looks like a mistake, while we don't recommend that, we can't make sure nobody uses the most easy way to do the task).

> Perhaps we should make the interactive mode to be the default? These commands are intended to be used interactively first and foremost.

We discussed that in zoom. The problem is that many commands like  `start/stop/status` are rarely useful in interactive mode. And remembering which ones are interactive and which ones aren't is ugly.

elprans:
> And skipping auth in this case leaves servers open from outside world

[...]

PR #124, opened 2020-08-13
New credentials implementation for `server init`

tailhook:
Based on what was discussed here: https://github.com/edgedb/edgedb-cli/issues/91#issuecomment-673383943
(`save-credentials` will be in follow-up PR)


tailhook:
@edgedb/engineering please review command names and arguments.

Issue #125, opened 2020-08-14
Crash on line wrapping query responses


msullivan:
- EdgeDB Version: latest from git
- EdgeDB CLI Version: latest from git
- OS Version: Ubuntu 20.04

I have a very specific crasher, but it's certainly more general.

Steps to Reproduce:

1. `CREATE TYPE Note { CREATE PROPERTY  name -> str; }; CREATE TYPE Person { CREATE PROPERTY name -> str { CREATE CONSTRAINT exclusive; }; CREATE MULTI link notes -> Note; };`

2. `FOR noob in {"Phil Emarg", "Madeline Hatch"} UNION (INSERT Person {name := noob});`

3. In an 83-character wide terminal, `SELECT Person {name, notes};`. It will work if the terminal is 82 or 84 characters wide.

The crash:
```
sully> SELECT Person {name, notes};
{
  Object {name: 'Phil Emarg', notes: {}},
  Object {name: 'Madeline Hatch', notes: {}},
thread 'async-std/runtime' panicked at 'poll after end of stream', /home/sully/.cargo/git/checkouts/edgedb-rust-8bc8936b1acdc166/1dde8bc/edgedb-client/src/reader.rs:196:19
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1076
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1537
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:218
[...]

PR #126, opened 2020-08-15
Fix refs to edgedb_client::Builder

1st1:
Related to https://github.com/edgedb/edgedb-cli/issues/125 and https://github.com/edgedb/edgedb-rust/pull/36


CodesInChaos:
I don't think it's a good idea to expose finely grained modules to the public. I'd prefer public re-exports. For the most important types at the crate root.

I'd also take the opportunity to rename `Builder` at the same time, since it's not clear what it's building.

PR #135, opened 2020-08-18
Interactive migrations

tailhook:


tailhook:
@edgedb/engineering take a look at keys: https://github.com/edgedb/edgedb-cli/pull/135/files#diff-03d96668e6b4251c1e17e46da8a7f003R91-R98 and messages as always.

src/migrations/create.rs
```diff
@@ -65,14 +75,57 @@ async fn read_schema_file(path: &Path) -> anyhow::Result<String> {
...
+
+    const HELP: &str = r###"
+y - confirm the prompt, use the DDL statements
+n - reject the prompt
+l - list the DDL statements associated with prompt
+c - list already confirmed EdgeQL statements
+b - revert back to previous save point, perhaps previous question
+s - stop and save changes (splits migration into multiple)
+q - quit without saving changes
+h or ? - print help
```
1st1, line 98: I like it.

src/migrations/create.rs
```diff
@@ -65,14 +75,57 @@ async fn read_schema_file(path: &Path) -> anyhow::Result<String> {
...
+            return Ok(Quit);
+        }
+        let val = match &input.trim().to_lowercase()[..] {
+            "y"|"yes" => Yes,
+            "n"|"no" => No,
+            "l"|"list" => List,
+            "c"|"confirmed" => Confirmed,
+            "b"|"back" => Back,
+            "s"|"stop"|"split" => Split,
+            "h"|"?"|"help" => {
```
1st1, line 115: Do you really need to match full words? You don't list them as an alternative to single characters.

tailhook: I like that we can use them in tutorials/interactive sessions listing:
```
Do you want to delete type A? [y,n,l,c,b,s,q]
no
Do you want to delete type B? [y,n,l,c,b,s,q]
back
Do you want to delete type A? [y,n,l,c,b,s,q]
list
The following DDL will be executed:
    DROP TYPE A;
Do you want to delete type A? [y,n,l,c,b,s,q]
yes
Do you want to delete type B? [y,n,l,c,b,s,q]
```
Since you have no help handy here, words are more readable I think (otherwise I would probably need to write comments near the letters to explain what it does to make the listing readable).

I'm not sure how often we will do that, but see no harm of accepting words.

1st1: Alright, I like it.

PR #144, opened 2020-09-08
Correct a couple of typos in server init command output

RobertoPrevato:
I noticed this typo when using `server init`:
_"Bootstrap complete. Server is up and ru**nnn**ing now."_

I came to make this small contribution, then noticed another typo in "Bootrapping" instead of "Bootstrapping". Changed to "Bootstrap complete" to be consistent with the output from other conditions.


1st1 (approved):

1st1:
Good catch!

PR #146, opened 2020-09-11
Sign Windows binaries

elprans:
This adds a build step that embeds a signature in the produced
executable.


ambv (approved):

ambv:
Migrations tests timed out on macOS. Unfortunate but unrelated to this PR.

PR #150, opened 2020-09-21
Docker

tailhook:
Discussion points:
1. We name containers `edgedb_<container_name>` and put owning user name into the `org.edgedb.metadata.user` label. This way we can add `--system` instances later, they will conflict with user instances by name, though. Which is fine I think as:
    a) We don't even know if we want system instances
    b) Probably users will have either system or user instances, not both
    c) Conflict between user and system instances are implied anyway (i.e. user instances override system instances with the same name). This scheme will create a conflict between multiple instances, though. But it's unlikely common scenario.
    d) Even if we eliminate name conflicts there would be conflicts on unused ports, which we can't (easily) coordinate for non-docker instances too
2. Currently we use `sh -c` in the edgedb container for setting permissions on the volume and `busybox` container for few utility functions working in volumes. We do that only on upgrade or on metadata failure (i.e. user dropped container, but not volume). Is it fine to rely on `busybox`? Is it fine to rely on `sh` being present in edgedb container?
3. We rely on `.edgedb/credentials` file when dumping the database on upgrade and store dump locally, so that we don't need any hacks with docker and CLI. This also makes in more similar to how we would upgrade cloud instance.
4. No image cleanup yet.


elprans:
> We name containers `edgedb_<container_name>` and put owning user name into

Why can't we use `edgedb_<username>_<container_name>` to distinguish between system/per-user instances and between instances of different users?

> 2\. Is it fine to rely on `sh` being present in edgedb container?

Yes.

tailhook:
> > We name containers `edgedb_<container_name>` and put owning user name into
> 
> Why can't we use `edgedb_<username>_<container_name>` to distinguish between system/per-user instances and between instances of different users?

Because using this container should look like:
```
docker run --link edgedb:edgedb_myproject ...
```
If this we add username we should add more explanation:
```
docker run --link edgedb:edgedb_<username>_myproject ...
# please replace <username> to your current user name
```
Or:
```
docker run --link edgedb:edgedb_$(id -un)_myproject ...
```
But that works only on linux.

Also on Windows user names can contain spaces (not sure this is a problem for docker container names, but fighting with windows arguments quoting is no pleasure).

And I even not sure how that may work with docker compose (everybody needs their own modified compose file?).

PR #157, opened 2020-10-05
Use the current database name when invoking psql

elprans:
Currently, `\psql` will always connect to `edgedb0`, which is
inconvenient, because you almost always want to work with the database
you currently have in edgedb-cli.

### DIFF
--- a/src/commands/psql.rs
+++ b/src/commands/psql.rs
@@ -8,7 +8,7 @@ use crate::commands::Options;
 use edgedb_client::server_params::PostgresAddress;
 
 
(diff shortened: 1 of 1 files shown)


tailhook:
I'm not against this. But why edgedb always sends `edgedb0` ? I would expect it to send respective database. This can matter in case we start encoding postgres database names or use uuids for them for some reason.

tailhook:
Ping. Should this be fixed in edgedb? Should I do the PR?

elprans:
This will be fixed in https://github.com/edgedb/edgedb/pull/2009.

PR #160, opened 2020-10-08
Use workflow_dispatch instead of repository_dispatch

ambv:
This is an equivalent of edgedb/edgedb#1809


ambv:
This requires edgedb/edgedb-pkg#28

tailhook (approved):

PR #168, opened 2020-10-14
Nicer message when we were unable to bootstrap a service

tailhook:
This is important on WSL or Centos7 (and potentially others like Alpine)

Now it prints something like that:
```
Failed to get D-Bus connection: No such file or directory
Error: cannot create service: process "systemctl" "--user" "daemon-reload" failed: exit code: 1
Bootrapping complete, but there was an error creating a service. You can run server manually via:
  edgedb server start --foreground test10
```
And still exits with erroneous code `2`. Still this is friendly enough
to be useful.

This PR also fixes bug of removing bootstrapped directory which popped up in 1-alpha6.

### DIFF
--- a/src/server/init.rs
+++ b/src/server/init.rs
@@ -12,10 +12,12 @@ use edgeql_parser::helpers::{quote_string, quote_name};
 use prettytable::{Table, Row, Cell};
 use fn_error_context::context;
 
+use crate::commands::ExitCode;
 use crate::platform::{config_dir, home_dir};
 use crate::server::reset_password::{generate_password, write_credentials};
 use crate::server::reset_password::{password_hash};
 use crate::server::detect::{self, VersionQuery};
+use crate::server::errors::CannotCreateService;
 use crate::server::metadata::Metadata;
 use crate::server::methods::{InstallMethod, Methods};
 use crate::server::options::{Init, StartConf};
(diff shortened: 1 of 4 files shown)


src/server/init.rs
```diff
@@ -246,14 +248,24 @@ pub fn init(options: &Init) -> anyhow::Result<()> {
...
+                        settings.name.escape_default());
+                    return Err(ExitCode::new(2))?;
+                } else {
+                    log::error!("Bootstrap error, cleaning up...");
+                    method.clean_storage(&settings.storage)
+                        .map_err(|e| {
+                            log::error!("Cannot clean up storage {}: {}",
+                                settings.storage.display(), e);
+                        }).ok();
+                    Err(e).context(format!("cannot bootstrap {}",
```
1st1, line 266:
```suggestion
                    Err(e).context(format!("Cannot bootstrap {}",
```

src/server/unix.rs
```diff
@@ -88,11 +82,14 @@ pub fn bootstrap(method: &dyn Method, settings: &init::Settings)
                     format!("error running server {:?}", cmd))?;
             init_credentials(&settings, &inst)?;
             drop(child);
-            println!("Bootstrap complete. To start a server:\n  \
-                      edgedb server start {}",
-                      settings.name.escape_default());
+            if settings.start_conf == StartConf::Manual {
+                println!("Bootstrap complete. To start a server:\n  \
```
1st1, line 86: I'm not sure if it's `a server` or `the server` in this context. @elprans ?

src/server/init.rs
```diff
@@ -246,14 +248,24 @@ pub fn init(options: &Init) -> anyhow::Result<()> {
...
-                    .map_err(|e| {
-                        log::error!("Cannot clean up storage {}: {}",
-                            settings.storage.display(), e);
-                    }).ok();
-                Err(e).context(format!("cannot bootstrap {}",
-                                       settings.storage.display()))?
+                if e.is::<CannotCreateService>() {
+                    eprintln!("Error: {:#}", e);
+                    eprintln!("Bootrapping complete, \
+                        but there was an error creating a service. \
```
1st1, line 254: Ditto, I think it should be `the server` in this context. @elprans ?

1st1 (approved):

src/server/init.rs
tailhook: I think we always have something like `Error: cannot bootstrap...`. And while we have some capitalized errors, I think this is a mistake and we generally tend to lowercase them. Capitalized messages should be only used when we directly output with `log/println/eprintln`

src/server/unix.rs
tailhook: IDK, tutorials usually use `a server`. This is kinda the same and not the same :(

src/server/init.rs
tailhook: I think we tried to create **a service** but failed to. So this there is no **the service** as it wasn't created. No?

1st1: Well, depends on what kind of service the tool failed to create. If it was a systemd service config I'd say we should use "the". If it was a temporary service just to check if everything was installed it should be "a". In this context it sounds like "the".

**That said**, using articles correctly, unfortunately, isn't my strong skill :/  So I'd fix both articles to "the" and let @Dhghomon do a post-commit review :)

src/server/unix.rs
tailhook: Done

src/server/init.rs
tailhook: Done

Dhghomon: Feels like "the" to me as well: it can still be multiple services but in the moment we're looking at "the" service in question. Also, using "a" feels (just a bit) like creating some new service that we didn't intend to.

PR #182, opened 2020-10-23
Add timeout to terminating interactive connections

### LINKED ISSUE #178: very long disconnect
I connected to a remote edgedb instance and disconnecting REPL from it took quite some time (I think around 60 seconds). 

In the end I was presented with this message:

```
[2020-10-22T19:36:33Z WARN  edgedb::repl] Termination error: error reading data: Operation timed out (os error 60): Operation timed out (os error 60)
```

First, I think we should cap the termination to 5 seconds and just forcefully close the socket if we don't receive response by then.

Second, the error message has "Operation timed out (os error 60)" repeated two times. Let's investigate why it's repeated and fix.

Lastly, it's not clear that 60 is an ERRNO. I'd change it to "Operation timed out (ETIMEOUT)"

tailhook:
Fixes #178

### DIFF
--- a/src/async_util.rs
+++ b/src/async_util.rs
@@ -1,0 +1,15 @@
+use std::io;
+use std::time::Duration;
+
+use async_std::future::Future;
+
+
+pub async fn timeout<F, T>(dur: Duration, f: F) -> anyhow::Result<T>
+    where F: Future<Output = anyhow::Result<T>>,
+{
+    use async_std::future::timeout;
+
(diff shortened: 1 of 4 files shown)


1st1 (approved):
I'm not sure if it should 5 seconds or 1 second like you've proposed. What do you think?

tailhook:
Fixed to 1 sec. The shorter is better.

PR #183, opened 2020-10-23
Better error handling

### LINKED ISSUE #177: REPL: reconnect on errors in \commands
Here's the scenario:

```
edgedb> \d Acess
Error executing command: ERROR: InvalidReferenceError: schema item 'Acess' does not exist
Reconnecting...
edgedb> \d Access
Error executing command: ERROR: InvalidReferenceError: schema item 'Access' does not exist
Reconnecting...
edgedb> \d metrics::Access
type metrics::Access {
```

We don't need to reconnect in these cases.

tailhook:
Fixes #177

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,6 +1,6 @@
 // We don't need to hunt of unused imports on windows, as they are harmless
 #![cfg_attr(windows, allow(unused_imports))]
-#![type_length_limit="2097152"]
+#![type_length_limit="2134214"]
 
 use std::env;
 use std::path::Path;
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -1,6 +1,6 @@
 // We don't need to hunt of unused imports on windows, as they are harmless
 #![cfg_attr(windows, allow(unused_imports))]
-#![type_length_limit="2097152"]
+#![type_length_limit="2134214"]
```
1st1, line 3: Just curious: is this for linting?

1st1 (approved):

src/main.rs
tailhook: Not exactly. Looks like new compiler exposed a limit (or just created a longer type) and we needed to bump it. Now we bump it again as we added some longer type (i.e. because of deeper nesting of futures is the most likely case)

PR #189, opened 2020-11-03
Upgrades dependencies

tailhook:
There are two important things:
1. Now `edgedb` user and database are default
2. Using migration hasher from edgeql-parser module (same used in edgedb itself)


1st1 (approved):

dmgolembiowski:
Thank you for this commit @tailhook

PR #190, opened 2020-11-03
Don't infer user and database name from unix user name

tailhook:
Second patch, previous removed inference code only in edgedb-client
itself

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -5,7 +5,6 @@ use anyhow::Context;
 use atty;
 use clap::{Clap, AppSettings};
 use edgedb_client::Builder;
-use whoami;
 
 use crate::credentials::get_connector;
 use crate::commands::parser::Common;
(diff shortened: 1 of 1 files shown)


src/options.rs
```diff
@@ -235,16 +234,8 @@ fn conn_params(tmp: &TmpOptions) -> anyhow::Result<Builder> {
...
-                String::from("edgedb")
-            } else {
-                whoami::username()
-            }
-        });
-        conn_params.user(user.clone());
-        conn_params.database(database.as_ref().map(|x| &x[..])
-                             .unwrap_or(&user));
+        user.as_ref().map(|user| conn_params.user(user));
+        database.as_ref().map(|db| conn_params.database(db));
```
1st1, line 238: Just for my education -- the code in the `if` branch appears almost identical, except there it's `user.map(...)` and here it's `user.as_ref().map(...)`. What's the difference?

1st1 (approved):

src/options.rs
tailhook: `as_ref()` converts `Option<String>` into `Option<&String>`. Without `as_ref()` the method `user.map(|user:String| { })` gets the string itself, not the reference. So it effectively moves out `user` variable and it can't be used later in code. In that if above it looks like we don't use `user` anymore. (we could write `as_ref()` above too, but not vice versa).

PR #192, opened 2020-11-06
Allow full range for dates and times in repl

### LINKED ISSUE #174: <datetime> that doesn't panic before 1970
I've been using cal::local_time and the rest in the tutorial a lot because the setting is the 19th century, and a <datetime> before 1970 leads to a panic that kicks me out of the server along with a big ol' Rust backtrace. I took a look at it though and I see it comes from the humantime crate at line 247:

https://docs.rs/crate/humantime/2.0.1/source/src/date.rs

which starts out with a duration_since and an .expect():

```
impl fmt::Display for Rfc3339Timestamp {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        use self::Precision::*;

        let dur = self.0.duration_since(UNIX_EPOCH)
            .expect("all times should be after the epoch");
```

Would be nice if the crate had something like a pub const FAKE_UNIX_EPOCH and an unwrap_or or a match where it uses UNIX_EPOCH in one case and FAKE_UNIX_EPOCH (a Unix epoch that starts at year 0 for example) in another.

tailhook:
Fixes #174

### DIFF
--- a/src/print/native.rs
+++ b/src/print/native.rs
@@ -110,9 +110,9 @@ impl FormatExt for Value {
             V::LocalDatetime(t)
             => prn.typed("cal::local_datetime", format!("{:?}", t)),
             V::LocalDate(d)
-            => prn.typed("std::local_date", format!("{:?}", d)),
+            => prn.typed("cal::local_date", format!("{:?}", d)),
             V::LocalTime(t)
-            => prn.typed("std::local_time", format!("{:?}", t)),
+            => prn.typed("cal::local_time", format!("{:?}", t)),
             V::Duration(d) => {
                 // TODO(tailhook) implement more DB-like duration display
                 prn.const_scalar(format_args!("{}{:?}",
--- a/src/print/native.rs
+++ b/src/print/native.rs
@@ -1,47 +1,13 @@
 use std::cmp::min;
-use std::convert::TryInto;
 
 use bigdecimal::BigDecimal;
 use colorful::Colorful;
-use chrono::format::{Item, Numeric, Pad, Fixed};
-use chrono::{NaiveDateTime, NaiveDate, NaiveTime};
-use humantime::format_rfc3339;
 use num_bigint::BigInt;
 
 use edgedb_protocol::value::Value;
 use crate::print::formatter::Formatter;
 use crate::print::buffer::Result;
 
-
-static DATETIME_FORMAT: &[Item<'static>] = &[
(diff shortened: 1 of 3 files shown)


src/print/native.rs
```diff
@@ -140,26 +106,13 @@ impl FormatExt for Value {
...
-                }
-            }
-            V::LocalTime(t) => {
-                prn.typed("cal::local_time",
-                    Into::<NaiveTime>::into(t)
-                        .format_with_items(TIME_FORMAT.iter()))
-            }
+            V::Datetime(t) => prn.typed("datetime", format!("{:?}", t)),
+            V::LocalDatetime(t)
+            => prn.typed("std::local_datetime", format!("{:?}", t)),
```
1st1, line 111:
```suggestion
            => prn.typed("cal::local_datetime", format!("{:?}", t)),
```

src/print/native.rs
```diff
@@ -140,26 +106,13 @@ impl FormatExt for Value {
...
+            V::LocalDate(d)
+            => prn.typed("std::local_date", format!("{:?}", d)),
```
1st1, line 113: ditto

src/print/native.rs
```diff
@@ -140,26 +106,13 @@ impl FormatExt for Value {
...
+            V::LocalTime(t)
+            => prn.typed("std::local_time", format!("{:?}", t)),
```
1st1, line 115: ditto

src/print/native.rs
tailhook: Yeah, thanks. That's some attention blindness. There are no prefixes for `std::` types here so I definitely meant `cal::` :)

src/print/tests.rs
```diff
@@ -105,6 +106,14 @@ fn bigint() {
     ]).unwrap(), "{10n, 10000n, 1e11n}");
 }
 
+#[test]
+fn datetime() {
+    assert_eq!(test_format(&[
+        Value::Datetime(Datetime::from_micros(-1000000000000000)),
+        Value::Datetime(Datetime::from_micros(1604506938347258)),
+    ]).unwrap(), "{<datetime>\'1968-04-23T22:13:20Z\', <datetime>\'2050-11-04T16:22:18.347258Z\'}");
```
1st1, line 114: nit: don't you try to keep line lengths under 79?

1st1 (approved):

PR #194, opened 2020-11-16
Implement operation_id for migrations

### LINKED ISSUE #193: Group migration operations by "prompt_id"
`DESCRIBE MIGRATION` may return a single logical migration operation as multiple pieces of DDL, which are connected by the `"operation_id"` attribute, which should be treated as an opaque string.  The `create-migration` _should not_ repeat the user query for the same `operation_id`.

tailhook:
Fixes #193


tailhook:
Updated PR. Waiting for edgedb/edgedb#2118 to land.

PR #195, opened 2020-11-19
Update github's set-env conventions

tailhook:


elprans (approved):
Oh. We need to do this across all our workflows. \cc  @ambv

tailhook:
@elprans update https://github.com/elprans/gha-setup-vcpp-build-tools (by merging upstream) to get windows building

elprans:
@tailhook There's only dependency bumps upstream, was that the problem? (I merged `ms-build/master`)

tailhook:
> @tailhook There's only dependency bumps upstream, was that the problem? (I merged `ms-build/master`)

Perhaps yes, I'm not sure. I've seen that there is a closed issue about that.

PR #199, opened 2020-12-06
Don't include builtin DBs in \list-databases or in "dumpall"

1st1:


src/commands/dump.rs
```diff
@@ -150,7 +150,10 @@ async fn get_databases(cli: &mut Connection)
     -> Result<Vec<String>, anyhow::Error>
 {
     let mut query = cli.query(
-        "SELECT sys::Database.name",
+        "WITH
```
elprans, line 153: You can write this as short as:
You could do shorter:

```suggestion
        "SELECT (sys::Database FILTER NOT .builtin).name"
```

elprans (approved):

src/commands/dump.rs
1st1: Sure, fixed

PR #212, opened 2020-12-15
Improving tests

tailhook:


tests/github-actions/install.rs
```diff
@@ -109,7 +109,7 @@ fn github_action_install() -> anyhow::Result<()> {
 
         println!("Install different version");
         Command::new(&edgedb)
-            .arg("server").arg("install").arg("--version=1-alpha5")
+            .arg("server").arg("install").arg("--version=1-alpha6")
```
1st1, line 112: Why not alpha7?

tailhook: I need two versions here

PR #222, opened 2021-02-05
Implement `print-stats` setting

tailhook:


src/commands/backslash.rs
```diff
@@ -539,6 +543,9 @@ pub async fn execute(cmd: &BackslashCmd, prompt: &mut repl::State)
                 ExpandStrings(b) => {
                     prompt.print.expand_strings = b.unwrap_value();
                 }
+                PrintStats(v) => {
+                    prompt.print_stats = v.value.expect("only writes here");
```
1st1, line 547: Not sure I understand the "only writes here" part

1st1 (approved):

src/commands/backslash.rs
tailhook: `v.value` is `None`, when on read command i.e. bare `\print-stats`. `expect` asserts on that `None`, and message in `expect` tries to explain that this should not happen, because read-only commands are processed elsewhere.

PR #231, opened 2021-02-16
Fix uninstall on MacOS and better tests

tailhook:
This adds `server list-versions --column=` parameter to the command-line, for easier tests and other scripting.


elprans (approved):

elprans:
Doesn't seem to compile in CI, though.

PR #234, opened 2021-02-16
Do not block Docker installation method on missing/inaccessible socket

elprans:
There is no Docker deamon Unix socket on Windows (there's a TCP port and
a named pipe, though [1]).  But, more importantly, the CLI itself
doesn't use the socket in any way, so the check doesn't seem to be
useful.

[1] https://docs.docker.com/desktop/faqs/#how-do-i-connect-to-the-remote-docker-engine-api

### DIFF
--- a/src/server/docker.rs
+++ b/src/server/docker.rs
@@ -1,7 +1,6 @@
 use std::ffi::OsStr;
 use std::fs;
 use std::fmt;
-use std::env;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Child};
 use std::time::{SystemTime, Duration, UNIX_EPOCH};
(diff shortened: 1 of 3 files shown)


tailhook:
We don't use socket, but mere presense of docker CLI doesn't mean docker daemon is running, so shouldn't assume docker works on the system. Note: if we detect docker as supported, but it doesn't actually work, using `--method=package` becomes nightmare too (i.e. `list-versions` checks both, `init` may hit both, `uninstall` checks both, and so on).

On windows we may suppress the check for now.

elprans:
> We don't use socket, but mere presense of docker CLI doesn't mean docker daemon is running, so shouldn't assume docker works on the system.

I still think checking the socket is not the best strategy to test if docker is working.  Why not run some CLI command as a test instead?

tailhook:
> > We don't use socket, but mere presense of docker CLI doesn't mean docker daemon is running, so shouldn't assume docker works on the system.
> 
> I still think checking the socket is not the best strategy to test if docker is working. Why not run some CLI command as a test instead?

Because it's slow, and a bunch of unexpected error codes and messages, which is unclear how to deal with. Checking for socket while is not perfect worked well so far.

And I think we may use docker API eventually anyway, so you don't need command-line and socket is the only thing we'll use.

tailhook:
Superseded by #242

PR #235, opened 2021-02-16
Editorial pass on self-install messages

elprans:
Update the wording of messages produced in the `edgedb-init` flow.

### DIFF
--- a/src/self_install.rs
+++ b/src/self_install.rs
@@ -46,9 +46,9 @@ fn print_long_description(settings: &Settings) {
     println!(r###"
 Welcome to EdgeDB!
 
-This will install the EdgeDB command-line tools.
+This will install the official EdgeDB command-line tools.
 
-It will add `edgedb` binary to the {dir_kind} bin directory located at:
+The `edgedb` binary will be placed in the {dir_kind} bin directory located at:
 
   {installation_path}
 {profile_update}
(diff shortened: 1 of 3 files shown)


src/server/methods.rs
```diff
@@ -79,11 +79,23 @@ impl InstallationMethods {
...
-                https://docs.docker.com/get-docker/");
-            buf.push_str("Or ask for native support at \
-                https://github.com/edgedb/edgedb-cli/issues/new\
-                ?template=install-unsupported.md");
+            if cfg!(windows) {
+                buf.push_str("EdgeDB server installation on Windows \
+                    requires Docker Desktop to be installed and running. \
+                    You can download Docker Desktop for Windows here: \
+                    https://hub.docker.com/editions/community/docker-ce-desktop-windows/ \
+                    Once Docker Desktop is installed and running, restart the \
```
1st1, line 87: "restart" is confnusing here. "perform"? "attempt"?

1st1 (approved):

src/server/methods.rs
elprans: Why is it confusing?  IMO it's the most appropriate verb here.  You started an install, it failed because you're missing a prerequisite, you fix the issue and then you restart the install.

src/self_install.rs
```diff
@@ -143,15 +143,15 @@ fn ensure_line(path: &PathBuf, line: &str) -> anyhow::Result<()> {
 
 fn print_post_install_message(settings: &Settings) {
     if cfg!(windows) {
-        print!(r###"# EdgeDB command-line tool is installed now!
+        print!(r###"# The EdgeDB command-line tool has been successfully installed!
```
tailhook, line 146: Rust also uses present simple here. This doesn't have to be past perfect passive.

tailhook (approved):

PR #236, opened 2021-02-16
Add en explicit option to inhibit the "Press Enter" message on Windows

elprans:
The new Powershell installation script runs `edgedb-init` from an
existing terminal window, and so the "Press Enter key to continue"
message is redundant.  This adds the new `--same-terminal` option
to inhibit the message.


tailhook:
I don't think it should be called `--same-terminal`. It should be something like `--suppress-prompt` or `--no-final-wait`, or something like that.

elprans:
How about `--wait-for-exit-prompt`?

tailhook (approved):

PR #237, opened 2021-02-17
Add `allow_dml_in_functions` param support

tailhook:


elprans (commented):
We made `allos_dml_in_functions` internal in edgedb/edgedb#240, so this should not be necessary anymore.

tailhook:
Okay

PR #241, opened 2021-02-17
Bump the default history-size to 10000

msullivan:
I find 100 to be much too low, and we don't have a persistent way to
configure it yet.


1st1:
> and we don't have a persistent way to configure it yet.

Perhaps we can make the REPL read `~/.edgedb-replrc` file or something like that on every run? @tailhook if that's too long/hard to implement at this moment then +1 to simply bumping the history size.

tailhook (approved):

tailhook:
Rc file should be too hard, but not great last minute change.

PR #244, opened 2021-02-18
Don't run tests one-too-many times and remove concurrency cap

elprans:


.github/workflows/tests.yml
```diff
@@ -69,10 +70,6 @@ jobs:
         with:
           server-version: ${{ matrix.edgedb-version }}
 
-      - uses: actions-rs/cargo@v1
-        with:
-          command: test
```
tailhook, line 74: Not sure about this one. Running with glibc is useful. We only need musl build for docker tests. We can run them in parallel or exclude non-docker tests from musl run, though.

elprans: But we already run glibc in a sibling job, how is this one different from that?

tailhook: Which one? I don't see any.

elprans: This?

https://github.com/edgedb/edgedb-cli/blob/496fb2f5b7e55636f130655f42fa2841cbdabe87/.github/workflows/tests.yml#L19-L44

tailhook (approved):

.github/workflows/tests.yml
tailhook: Ah, yes sorry. I've thought that should exclude ubuntu, but this way is fine.

PR #277, opened 2021-03-26
Implement `--json` for selected `edgedb server` commands

### LINKED ISSUE #275: Add a flag to output json for the 'edgedb server' commands
It would be useful for other tools using the cli (like edgedb studio) if there was a flag to output json instead of a table for the `server` informational commands, ie. `info`, `list-versions`, and `status`.

tailhook:
Fixes #275

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1619,6 +1619,10 @@ name = "linked-hash-map"
 version = "0.5.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7fb9b38af92608140b86b693604b9ffcc5824240a484d1ecd4795bacb2fe88f3"
+dependencies = [
+ "serde",
+ "serde_test",
+]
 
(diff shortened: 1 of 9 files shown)


elprans:
Perhaps `--fmt=json`?  This would leave the possibility of adding future formatting options without the need of adding new arguments.

tailhook:
> Perhaps `--fmt=json`? This would leave the possibility of adding future formatting options without the need of adding new arguments.

I don't think we ever need more formats. It's more like `--porcelain` in git, i.e. it means "programmatic output" rather than here are multiple formats of how we would output the data.

Also we have top-level arguments (mostly for repl):
```
    -j, --json                JSON output for the queries (single JSON list per query)
    -t, --tab-separated       Tab-separated output of the queries
```
Which should be short, as they might be used when running commands manually. So supporting `--json` here is a natural choice.

PR #278, opened 2021-03-29
Process more systemd messages on destroy

### LINKED ISSUE #254: unhelpful message when nonexistant server is destroyed
```
$ edgedb server destroy instance_that_doesnt_exists
edgedb error: Error running "systemctl" "--user" "stop" "edgedb-server@instance_that_doesnt_exists": Failed to stop edgedb-server@instance_that_doesnt_exists.service: Unit edgedb-server@instance_that_doesnt_exists.service not loaded.
```

- EdgeDB Version: package:1.0b2.dev5478+g25318fe0e-2021022200~focal.nightly
- EdgeDB CLI Version: edgedb-cli 1.0.0-beta.1
- OS Version: Linux 5.8.0-44-generic #50~20.04.1-Ubuntu SMP Wed Feb 10 21:07:30 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

tailhook:
Fixes #254

### DIFF
--- a/src/server/linux.rs
+++ b/src/server/linux.rs
@@ -444,7 +444,9 @@ pub fn get_instance<'x>(method: &'x dyn Method, name: &str)
 fn systemd_is_not_found_error(e: &str) -> bool {
     e.contains("Failed to get D-Bus connection") ||
     e.contains("Failed to connect to bus") ||
-    e.contains("No such file or directory")
+    e.contains("No such file or directory") ||
+    e.contains(".service not loaded") ||
+    e.contains(".service does not exist")
 }
 
 pub fn destroy(options: &Destroy) -> anyhow::Result<()> {


elprans (commented):
These appear to rely on an English system locale.  Is there a way to rely on exit codes instead?

tailhook:
Most of the time it's generic error code of 1:
```
$ systemctl --user stop test || echo $?
Failed to get D-Bus connection: No such file or directory
$ systemctl --user stop something || echo $?
Failed to stop something.service: Unit something.service not loaded.
5
$ systemctl --user disable something || echo $?
Failed to disable unit: Unit file something.service does not exist.
1
```
We can match this one on stop, which seems to be consistent now (https://github.com/systemd/systemd/blob/414f59dad7502b1ca0ea492c47a48fd6914e08b6/src/systemctl/systemctl-util.c#L102-L103), but I'm still not sure if this consistency will be kept in future, because the symbolic name of that is actually for different purposes: `5 | EXIT_NOTINSTALLED | The program is not installed.` (https://www.freedesktop.org/software/systemd/man/systemd.exec.html). Also it looks like that exit code can be propagated from the unit itself (I haven't verified the latter, as this being untrue doesn't fully fix the issue).

Currently I can't make systemd make locale-dependent messages, but I can't make sure that it will not happen. Do you think that setting locale env vars would be enough to ensure that messages are right?

PR #286, opened 2021-04-09
Project init implementation

tailhook:


src/server/docker.rs
```diff
@@ -1026,6 +1026,13 @@ impl<'os, O: CurrentOs + ?Sized> Method for DockerMethod<'os, O> {
             port: settings.port,
             start_conf: settings.start_conf,
         })?;
+        if settings.project {
+            println!("To connect run either of:\n  edgedb\n  edgedb -I {}",
```
elprans: I'd only offer the parameter-less option here to reduce the conceptual option noise: "To connect run `edgedb` from the project directory".

tailhook: I'm concerned that:
1. Bare `edgedb` doesn't look like something specified. I.e. use `edgedb` to run migrations, or use `edgedb` to connect (sure, what else I would run)
2. Running `edgedb project init --project-dir=/something` (we can obviously detect this, and add `edgedb --project-dir`

tailhook:
Merged. Messages can be fixed in the subsequent PRs.

PR #310, opened 2021-04-25
Apply `cargo fmt` and fixes for `cargo clippy`

Fogapod:
This PR includes 2 commits:
- `cargo fmt`
- fixes for `cargo clippy`

I can split it into 2 separate PRs if needed.

Not sure how much code is covered by tests, there could be bugs.


src/server/macos.rs
```diff
@@ -147,140 +157,157 @@ impl<'os> Method for PackageMethod<'os, Macos> {
...
+                        .split(|&b| b == b'\n')
+                        .filter_map(|line| str::from_utf8(line).ok());
+                    let mut maybe_version: Option<&str> = None;
+                    for line in lines {
+                        maybe_version = match line.strip_prefix("version:") {
+                            Some(version) => Some(version),
+                            None => continue,
+                        };
+
                         break;
```
Fogapod, line 281: This is probably not the most elegant solution but I'm new to Rust and I tried to stay close to existing logic.
`break VALUE;` is not possible in for loop.

src/prompt.rs
```diff
@@ -366,7 +374,7 @@ fn show_history(history: &History) -> Result<(), anyhow::Error> {
             }
         }
     }
-    drop(childin);
+
```
Fogapod, line 377: I'm not sure what's going on here and why this is needed. Clippy said that this drop call does nothing, so I removed it.

src/commands/dump.rs
```diff
@@ -135,47 +134,44 @@ async fn dump_db(cli: &mut Connection, _options: &Options, filename: &Path)
...
-        if let Some(text) = response.next().await.transpose()? {
-            break text;
-        } else {
-            anyhow::bail!("`{}` returns empty value", query.escape_default());
-        }
+    let text = if let Some(text) = response.next().await.transpose()? {
+        text
+    } else {
+        anyhow::bail!("`{}` returns empty value", query.escape_default());
     };
```
Fogapod, line 161: Not sure if this is correct.

src/migrations/create.rs
```diff
@@ -72,6 +72,11 @@ pub struct Proposal {
     pub required_user_input: Vec<RequiredUserInput>,
 }
 
+// NOTE: clippy emits warning about needless question mark inside Queryable.String
+// This could be a bug in clippy.
+//
+// Adding #[allow(clippy::needless_question_mark)] does not help.
+#[allow(clippy::needless_question_mark)]
 #[derive(Deserialize, Queryable, Debug)]
```
Fogapod, line 80: This is the only remaining clippy warning and it can't be silenced. Should be fixed in the original macros crate or I don't understand something.

Fogapod (commented):
I left a few comments in places where I was unsure. The rest of changes are simple things like removing `return` or folding `else if`.

src/main.rs
```diff
@@ -43,39 +43,37 @@ mod variables;
...
+                    "  Hint: {}",
+                    e.hint.lines().collect::<Vec<_>>().join("\n        ")
+                );
+            } else if item.is::<bug::Bug>() {
+                eprintln!(
+                    "  Hint: This is most likely a bug in EdgeDB \
                         or command-line tools. Please consider opening an \
                         issue ticket at \
                         https://github.com/edgedb/edgedb-cli/issues/new\
-                        ?template=bug_report.md");
```
Fogapod, line 70: This indentation might be wrong.

src/server/package.rs
```diff
@@ -45,19 +44,23 @@ pub struct Package {
...
         Package {
             major_version,
-            version: Version(format!("{}-{}", self.version, self.revision)),
+            version: Version(format!("{}-{}", pkg_info.version, pkg_info.revision)),
             slot: slot.as_ref().to_owned(),
-        }.into_ref()
+        }
+        .into_ref()
     }
 }
```
Fogapod, line 65: Clippy said that `From` is better than `Into` because `From` implements `Into` for free. Is this implementation correct?

PR #313, opened 2021-04-28
Init project in self install

tailhook:
This effectively does `project init` in `curl .. | sh` pipeline.

Note: we don't do project init when using `-y` flag. This latter is for scripting and it's better to use explicit `project init` in scripts.

### DIFF
--- a/src/print/native.rs
+++ b/src/print/native.rs
@@ -24,6 +24,7 @@ fn format_string(s: &str, expanded: bool) -> String {
             '\u{0080}'..='\u{009F}'
             => buf.push_str(&format!("\\u{:04x}", c as u32)),
             '\'' => buf.push_str("\\'"),
+            '\\' => buf.push_str("\\\\"),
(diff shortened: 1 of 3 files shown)


1st1:
I don't like running init by default.  If this is to simplify our tutorials, maybe we should add an explicit `-- --init` flag specifically to the install command we feature in our tutorials. But the default one should not init; I don't think that by default users will run this command in their specific project dir.

elprans:
> add an explicit `-- --init` flag

The installer is interactive by default, why can't we simply ask?  E.g "Detected an EdgeDB project directory, would you like to initialize a database instance for it?"

1st1:
> E.g "Detected an EdgeDB project directory, would you like to initialize a database instance for it?"

Is that what this PR is doing? I can't see that from the diff.

elprans:
> Is that what this PR is doing?

It's detecting the project dir, but doesn't ask currently.

PR #332, opened 2021-05-04
describe-schema: initial implementation

### LINKED ISSUE #327: `DESCRIBE SCHEMA AS SDL` shortcut
Now that we have the command `DESCRIBE SCHEMA AS SDL` [implemented](https://github.com/edgedb/edgedb/pull/2481), it would be nice to have a shortcut for it in REPL. Something like `\ds` perhaps.

tailhook:
Fixes #327

### DIFF
--- a/src/commands/backslash.rs
+++ b/src/commands/backslash.rs
@@ -30,6 +30,7 @@ const HELP: &str = r###"
 Introspection
   (options: -v = verbose, -s = show system objects, -I = case-sensitive match)
   \d [-v] NAME             describe schema object
+  \ds, \describe-schema    describe the whole schema
   \l, \list-databases      list databases
   \lT [-sI] [PATTERN]      list scalar types
                            (alias: \list-scalar-types)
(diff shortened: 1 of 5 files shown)


src/commands/backslash.rs
```diff
@@ -30,6 +30,7 @@ const HELP: &str = r###"
 Introspection
   (options: -v = verbose, -s = show system objects, -I = case-sensitive match)
   \d [-v] NAME             describe schema object
+  \ds                      describe the whole schema
```
elprans, line 33: No long variant?

1st1:
Please don't merge this until we discuss how this should be called. Hopefully tomorrow

src/commands/backslash.rs
tailhook: `\describe-schema`. Well, we list long variants here inconsistently :(

tailhook:
Merged in. Will fix to the RFC1006 together with other commands.

PR #312, opened 2021-04-26
Silence warning on launchctl load

tailhook:
Does that by using `enable` and `kickstart` separately

Fixes #287

PR #333, opened 2021-05-04
Fix Launchctl service on macOS

fantix:
Now we follow the way how Homebrew service works:

1. Use `launchctl bootstrap` instead of `launchctl kickstart`
2. Use `launchctl bootout` instead of `launchctl unload`
3. Only run `launchctl enable` for auto-starting services
4. Keep the unit.plist file in the data directory for manual-starting
   services
5. Check service status before start/stop/destroy

Assuming `launchctl bootout` is available, macOS 10.11 or above is
now required. We could add an additional check on macOS version and
fallback to `launchctl unload` like what Homebrew did, but I'm not
sure we want to support those macOS versions over 6 years old.

Refs #312.

Refs:

https://babodee.wordpress.com/2016/04/09/launchctl-2-0-syntax/
https://github.com/Homebrew/homebrew-services/blob/afa2516060bcf0594f4914d7af53e33a4f4c5af5/lib/services_cli.rb

### 1. `server init` or `project init` with `--start-conf=auto`

* A new instance should be created without error.
* The new instance should be in the status of `Running`.
* Restarting the computer or re-logging in the OS user, the instance should still be `Running`.
* Able to access the instance and run queries.

### 2. `server init` or `project init` with `--start-conf=manual`

* A new instance should be created without error.
* The new instance should be in the status of `Inactive`.
* Restarting the computer or re-logging in the OS user, the instance should still be `Inactive`.
* No server process should be found; should not be able to access the instance.

### 3. Start a manual-starting instance

1. Create an instance with `--start-conf=manual`.
2. Start the instance with `edgedb server start`.
3. The instance should be in the status of `Running`.
4. Running `edgedb server start` again shouldn't cause any error.
5. Kill the server process with `SIGKILL`, the supervisor should automatically restart the instance.
6. Restarting the computer or re-logging in the OS user, the instance should be `Inactive`.

### 4. Start a auto-starting instance

1. Create an instance with `--start-conf=auto`.
2. The instance should be in the status of `Running`.
3. Running `edgedb server start` shouldn't cause any error.
4. Kill the server process with `SIGKILL`, the supervisor should automatically restart the instance.

### 5. Start a manually-stopped instance

1. Start an instance (either auto or manual).
2. Kill the server process with `SIGTERM` - the server should exit normally.
3. Check `edgedb server status`, the instance should have status `not running` (`Failed`).
4. Start the instance with `edgedb server start`.
5. The instance is `Running` again.

### 6. Stop an instance

1. In all these cases:
  a. `Running` auto-starting instance.
  b. `Running` manual-starting instance.
  c. `Inactive` instance (already stopped instance).
  d. `Failed` (`not running`) instance (server process killed with `SIGTERM`).
2. `edgedb server stop` should run without errors.
3. The instance should be in the status of `Inactive`.
4. No server process should be found; should not be able to access the instance.

### 7. Restart an instance

1. Start an instance (either auto or manual).
2. Remember the server process PID.
3. Running `edgedb server restart` should succeed without errors.
4. The instance should still be `Running` and accessible.
5. The server process PID should be different than it was in step 2.

### 8. Restart an abnormal instance

Restarting an instance in the status of either `Inactive` or `Failed` (`not running`, process killed with `SIGTERM`) should have the same effect as running `start`, so the instance should be started `Running` without errors.

### 9. Check the service status

Running `edgedb server status --service` on instances in the status of either `Running` or `Failed` (`not running`, process killed with `SIGTERM`) should echo the launchd status, while `Inactive` instances should report an error because their launchd service is not loaded.

### 10. Destroy an instance

1. In all these cases:
  a. `Running` instance.
  b. `Inactive` instance (already stopped instance).
  c. `Failed` (`not running`) instance (server process killed with `SIGTERM`).
2. Running `edgedb server destory` should succeed without errors.
3. Instance is gone in `edgedb server status`.
4. No server process found in the OS.
5. `launchctl list` should not contain anything with the instance name.
6. Nothing about the instance can be found under either `~/Application Support/edgedb/data/` or `~/Library/LaunchAgents/`.

### DIFF
--- a/src/server/macos.rs
+++ b/src/server/macos.rs
@@ -334,17 +334,28 @@ impl<'os> Method for PackageMethod<'os, Macos> {
         unix::upgrade(todo, options, self)
     }
     fn destroy(&self, options: &Destroy) -> anyhow::Result<()> {
-        let mut found = false;
-        log::info!(target: "edgedb::server::destroy",
-            "Unloading service");
-        let unit_path = unit_path(&options.name)?;
-        process::run(&mut StdCommand::new("launchctl")
-            .arg("unload").arg(&unit_path))?;
-        if unit_path.exists() {
+        let mut found;
+        let mut unit_file = unit_path(&options.name, true)?;
+        if unit_file.exists() {
             found = true;
+        } else {
+            unit_file = unit_path(&options.name, false)?;
+            found = unit_file.exists();
+        }
+        if found {
+            match launchctl_status(&options.name, false, &StatusCache::new()) {
+                Service::Running {..} => {
+                    log::info!(target: "edgedb::server::destroy",
+                               "Unloading service");
+                    let domain_target = get_domain_target();
+                    process::run(&mut StdCommand::new("launchctl")
+                        .arg("bootout").arg(&domain_target).arg(&unit_file))?;
+                },
+                _ => {},
+            }
             log::info!(target: "edgedb::server::destroy",
-                "Removing unit file {}", unit_path.display());
-            fs::remove_file(unit_path)?;
+                       "Removing unit file {}", unit_file.display());
+            fs::remove_file(unit_file)?;
         }
         let dir = unix::storage_dir(&options.name)?;
         if dir.exists() {
@@ -385,7 +396,7 @@ impl LocalInstance<'_> {
         })
     }
     fn unit_path(&self) -> anyhow::Result<PathBuf> {
-        unit_path(&self.name)
+        unit_path(&self.name, self.get_meta()?.start_conf == StartConf::Auto)
     }
     fn socket_dir(&self) -> anyhow::Result<PathBuf> {
         Ok(runtime_dir(&self.name)?)
@@ -424,8 +435,12 @@ impl<'a> Instance for LocalInstance<'a> {
         let service = launchctl_status(&self.name, system,
             // TODO
             &StatusCache::new());
-        let service_exists = launchd_plist_path(&self.name, system)
-            .map(|p| p.exists())
+        let service_exists = self.get_start_conf()
+            .map(|start_conf| launchd_plist_path(
+                &self.name, system, start_conf == StartConf::Auto)
+                .map(|p| p.exists())
+                .unwrap_or(false)
+            )
             .unwrap_or(false);
         unix::status(&self.name, &self.path, service_exists, service)
     }
@@ -433,20 +448,36 @@ impl<'a> Instance for LocalInstance<'a> {
         if options.foreground {
             process::run(&mut self.get_command()?)?;
         } else {
-            let lname = self.launchd_name();
-            process::run(
-                StdCommand::new("launchctl").arg("enable").arg(&lname)
-            )?;
-            process::run(
-                StdCommand::new("launchctl").arg("kickstart").arg(&lname)
-            )?;
+            match launchctl_status(&options.name, false, &StatusCache::new()) {
+                Service::Running {..} => {
+                    log::error!(target: "edgedb::server::start",
+                               "Service is already running");
+                },
+                _ => {
+                    let domain_target = get_domain_target();
+                    process::run(
+                        StdCommand::new("launchctl").arg("bootstrap")
+                            .arg(&domain_target).arg(&self.unit_path()?)
+                    )?;
+                },
+            }
         }
         Ok(())
     }
     fn stop(&self, _options: &Stop) -> anyhow::Result<()> {
-        process::run(&mut StdCommand::new("launchctl")
-            .arg("unload")
-            .arg(&self.unit_path()?))?;
+        match launchctl_status(&_options.name, false, &StatusCache::new()) {
+            Service::Running {..} => {
+                let domain_target = get_domain_target();
+                process::run(&mut StdCommand::new("launchctl")
+                    .arg("bootout")
+                    .arg(&domain_target)
+                    .arg(&self.unit_path()?))?;
+            },
+            _ => {
+                log::error!(target: "edgedb::server::stop",
+                           "Service is not running");
+            },
+        }
         Ok(())
     }
     fn restart(&self, _options: &Restart) -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


src/server/macos.rs
```diff
@@ -334,17 +334,28 @@ impl<'os> Method for PackageMethod<'os, Macos> {
...
-        let mut found = false;
-        log::info!(target: "edgedb::server::destroy",
-            "Unloading service");
-        let unit_path = unit_path(&options.name)?;
-        process::run(&mut StdCommand::new("launchctl")
-            .arg("unload").arg(&unit_path))?;
-        if unit_path.exists() {
+        let mut found;
+        let mut unit_file = unit_path(&options.name, true)?;
+        if unit_file.exists() {
```
tailhook, line 339: Here we run `unload` regardless of the existing file, because it possible that file was deleted by someone, but service is still running in memory. So I prefer to run `unload` (or bootout, or whatever) command anyway, to make sure we cleaned out all leftovers.

src/server/macos.rs
```diff
@@ -424,29 +435,49 @@ impl<'a> Instance for LocalInstance<'a> {
...
-            let lname = self.launchd_name();
-            process::run(
-                StdCommand::new("launchctl").arg("enable").arg(&lname)
-            )?;
-            process::run(
-                StdCommand::new("launchctl").arg("kickstart").arg(&lname)
-            )?;
+            match launchctl_status(&options.name, false, &StatusCache::new()) {
+                Service::Running {..} => {
+                    log::error!(target: "edgedb::server::start",
```
tailhook, line 453: It's not an error running `start` if service is already started. We can write this on `info` level, I think even `warn` is too scary.

src/server/macos.rs
```diff
@@ -424,29 +435,49 @@ impl<'a> Instance for LocalInstance<'a> {
...
+        match launchctl_status(&_options.name, false, &StatusCache::new()) {
+            Service::Running {..} => {
+                let domain_target = get_domain_target();
+                process::run(&mut StdCommand::new("launchctl")
+                    .arg("bootout")
+                    .arg(&domain_target)
+                    .arg(&self.unit_path()?))?;
+            },
+            _ => {
+                log::error!(target: "edgedb::server::stop",
```
tailhook, line 477: It's also not an error to run stop if service is already stopped. `log::info` is good enough.

fantix:
Observed Launchctl facts:

1. `launchctl bootstrap` will report an error if the service is already running.
2. `launchctl bootout` will report an error if the service is not running.
3. `launchctl bootstrap` and `launchctl bootout` will report an error if the service is disabled (both in plist and through command).
4. The service will auto-start if `KeepAlive` is set in the unit.plist even without `RunAtLoad=true`.
5. The service will NOT auto-start if the unit.plist file is not placed in one of the `LaunchAgents` folders, even if the unit.plist file gets loaded manually with `KeepAlive` or `RunAtLoad` defined.

src/server/macos.rs
```diff
@@ -334,17 +334,28 @@ impl<'os> Method for PackageMethod<'os, Macos> {
...
+        let mut found;
+        let mut unit_file = unit_path(&options.name, true)?;
+        if unit_file.exists() {
             found = true;
+        } else {
+            unit_file = unit_path(&options.name, false)?;
+            found = unit_file.exists();
+        }
+        if found {
+            match launchctl_status(&options.name, false, &StatusCache::new()) {
```
fantix, line 346: I feel like I should do something to these `launchctl_status` calls, but I'm not sure about the direction. Please kindly advise.

src/server/macos.rs
```diff
@@ -525,22 +564,26 @@ pub fn get_server_path(slot: &str) -> PathBuf {
...
+fn plist_dir(system: bool, auto_start: bool, name: &str) -> anyhow::Result<PathBuf> {
+    if auto_start {
+        if system {
+            Ok(PathBuf::from("/Library/LaunchDaemons"))
+        } else {
+            Ok(home_dir()?.join("Library/LaunchAgents"))
+        }
     } else {
-        Ok(home_dir()?.join("Library/LaunchAgents"))
+        unix::storage_dir(name)
```
tailhook, line 575: How it is going to work? I don't think launchd recognize units in arbitrary dirs (or does it?). I that's the intention, let's not write it at all?

But the idea is that manually run services you can start via `edgedb server start xxx`. `--start-conf=manual` just means that service should not be started at system startup (or user login).

src/server/macos.rs
tailhook: Have you tested without them? Usually supervisors work like we want: don't do anything (i.e. no error) if service is already running. So are you sure that these status calls are required?

If you are, it looks like you always only look for `Running`. So you might wrap it into `is_running` function.

fantix: Right - will do, thanks!

src/server/macos.rs
fantix: You're right - looks like `launchctl bootout` supports taking a service target, I'll use that directly.

src/server/macos.rs
fantix: OK!

src/server/macos.rs
fantix: > But the idea is that manually run services you can start via edgedb server start xxx. --start-conf=manual just means that service should not be started at system startup (or user login).

Yes I think we're on the same page.

> How it is going to work? I don't think launchd recognize units in arbitrary dirs (or does it?). I that's the intention, let's not write it at all?

It turns out that launchd will load plist from arbitary dirs if given explicitly, and manages it normally as instructed (e.g. `KeepAlive + SuccessfulExit=false` will restart the service if it's killed with SIGKILL, but not for SIGTERM). However, launchd won't remember plist files from arbitary dirs, that means those manually-loaded services won't be automatically started after reboot or re-login, which is exactly what we need.

src/server/macos.rs
fantix: Hmm I found a failing situation where the server process exited normally (e.g. killed with SIGTERM) but the service was never unloaded (pid = `-`, `launchctl_status() == Failed`), in which case the service can be restarted with `kickstart`, but `bootstrap` will fail because the service is already loaded. So I'll keep using `launchctl_status()` in `start` (`kickstart` if `Failed` or `Error`, `bootstrap` if `Inactive`, log info if `Running`), and use `fn is_service_loaded() { launchctl_status() != Inactive }` in `stop`/`destroy`/`status`.

src/server/macos.rs
tailhook: Okay.

src/server/macos.rs
tailhook: Maybe we need to use `kill` for stop and `bootout` only on destroy? So that start is `kickstart` (except for manually started, which we `bootstrap` on start, but I don't think we have to `bootout` them on stop).

So we only need to check for status for start of manually started instances.
(hope this makes sense)

tailhook (approved):
Looks good. Thanks!

src/server/macos.rs
fantix: Oops sorry I missed this comment. Yes, that's actually much cleaner to me - `bootstrap` and `bootout` for loading/unloading the service with launchd in only `init` and `destroy`, and `kickstart` / `SIGTERM` for (re-)starting and stopping the registered service. That way, we don't need to keep the plist file for manually-started services as you mentioned - because we only need to bootstrap them into launchd once at `init`.

PR #340, opened 2021-05-06
New way to manage launchd service on macOS

fantix:
Based on #333, now we won't store a plist file for manually-starting
services. Instead, such service will be loaded into launchd from a
temporary plist file. At the same time, we won't unload services on
stop, instead we use `launchctl kill SIGTERM` to stop the server, so
that we could mostly use `launchctl kickstart` to start the service.

### DIFF
--- a/src/server/macos.rs
+++ b/src/server/macos.rs
@@ -1,6 +1,8 @@
 use std::collections::BTreeSet;
 use std::fs;
 use std::str;
+use std::thread;
+use std::time;
 use std::path::{Path, PathBuf};
 use std::process::{Command as StdCommand};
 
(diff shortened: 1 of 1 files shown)


tailhook (approved):

1st1:
Is this critical to have before the release?

tailhook:
We've discussed that privately with @elprans and decided that it's better to merge this.

fantix:
@1st1 according to my testings, we should be fine with #333 (already in master) without #340. However #333 is somewat different than #340 in implementation detail, so it would be nice to skip #333 and jump to #340 if possible. I'll update here more info about version upgrade compatibilies and drawbacks after more testing, but at the very least people could just destroy instances created in beta1 and recreate with beta2.

fantix:
> We've discussed that privately with @elprans and decided that it's better to merge this.

Oh okay, sorry only saw the update after previous comment. Will merge now, thanks!

fantix:
Also I think I should document the explanation of how EdgeDB CLI works on macOS with launchd somewhere.

PR #371, opened 2021-06-30
Use .edgeql as the suffix for the \e temp file

msullivan:
This makes it more likely that an editor integration will pick it up.


elprans (approved):

tailhook:
Hm, looks like a typo. Thanks!

PR #381, opened 2021-07-14
Fix self migrate instance recreation on macOS

fantix:
We need to `bootout` the service first to let launchd reload the new service file, so that it doesn't write to old log files. Also this is needed to run the `bootstrap`.

Also fixed a typo.

### DIFF
--- a/src/server/macos.rs
+++ b/src/server/macos.rs
@@ -342,10 +342,7 @@ impl<'os> Method for PackageMethod<'os, Macos> {
             // unnecessary to unload the service if it wasn't loaded.
             log::info!(target: "edgedb::server::destroy",
                        "Unloading service");
-            process::run(&mut StdCommand::new("launchctl")
-                .arg("bootout")
-                .arg(&launchd_name(&options.name))
-            )?;
+            bootout_launchctl_service(&options.name)?;
         }
 
         let mut found = false;
@@ -728,12 +725,10 @@ fn log_file(name: &str) -> anyhow::Result<PathBuf> {
 }
 
 fn bootout_launchctl_service(name: &str) -> anyhow::Result<()> {
-    if is_service_loaded(name) {
-        let unit_name = launchd_name(name);
-        process::run(
-            StdCommand::new("launchctl").arg("bootout").arg(&unit_name),
-        )?;
-    }
+    let unit_name = launchd_name(name);
+    process::run(
+        StdCommand::new("launchctl").arg("bootout").arg(&unit_name),
+    )?;
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


src/server/macos.rs
```diff
@@ -773,6 +783,17 @@ pub fn create_launchctl_service(name: &str, meta: &Metadata)
     }
 }
 
+pub fn recreate_launchctl_service(name: &str, meta: &Metadata)
+    -> anyhow::Result<()>
+{
+    bootout_launchctl_service(name)?;
+    if meta.start_conf == StartConf::Auto {
```
elprans, line 790: I think this should be conditional on `is_service_loaded` rather than `start_conf`.  If the service was loaded it should stay loaded after the upgrade.

fantix: Yes I agree! It's just a little bit more complex than that, because a loaded service may not be running. And I think we are expecting the running state to be reserved after the upgrade. So we have 6 situations:

| Auto?  | Loaded? | Running? | Expected Result                                           |
|--------|---------|----------|-----------------------------------------------------------|
| auto   | Yes     | Yes      | loaded + running                                          |
| auto   | Yes     | No       | loaded + stopped                                          |
| auto   | No      | No       | loaded + stopped (assuming auto service is always loaded) |
| manual | Yes     | Yes      | loaded + running                                          |
| manual | Yes     | No       | not running                                               |
| manual | No      | No       | not running                                               |

Means we shall also check if the instance was running before the upgrade. I'll push a fix for this.

fantix: Loading the new service will automatically start the server, and unfortunately we don't have a reliable-yet-easy way to wait to the right moment to send the SIGTERM to achieve the "loaded + stopped" result. I'm currently just sleeping 10s because it is about the time for startup, and this is a relatively rare case (auto-starting service manually stopped before migration).

src/server/macos.rs
```diff
@@ -773,6 +778,40 @@ pub fn create_launchctl_service(name: &str, meta: &Metadata)
...
+    };
+
+    let meta = instance.get_meta()?;
+    if is_running || meta.start_conf == StartConf::Auto {
+        log::info!("Updating service file for instance {:?}", name);
+        bootstrap_launchctl_service(name, meta)?;
+        if is_running {
+            log::info!("Started instance {:?}", name);
+        } else {
+            // Wait for the instance to start so that we can stop it,
```
elprans, line 806: What is the scenario where the instance wouldn't be running?

elprans (approved):

src/server/macos.rs
fantix: The table is updated - no more "loaded + stopped" results now.

PR #386, opened 2021-07-19
Use the updated Docker image tag names

fantix:
Refs edgedb/edgedb#2690

### DIFF
--- a/src/server/docker.rs
+++ b/src/server/docker.rs
@@ -355,7 +355,7 @@ impl Tag {
                 Some(rev) => Some(Tag::Stable(name.into(), rev[..7].into())),
                 None => None,
             }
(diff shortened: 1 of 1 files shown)


src/server/docker.rs
```diff
@@ -355,8 +355,8 @@ impl Tag {
                 Some(rev) => Some(Tag::Stable(name.into(), rev[..7].into())),
                 None => None,
             }
-        } else if name.starts_with("202") {
-            // example: `20200826052156c04ba5`
+        } else if name.starts_with("nightly_1") {
```
elprans, line 358:
```suggestion
        } else if name.starts_with("nightly_") {
```

We will not stay on `1` forever.

elprans (approved):

PR #389, opened 2021-07-22
Don't panic if Docker set EDGEDB_PORT=tcp://...

fantix:
Running the sample on https://github.com/edgedb/edgedb-docker will hti an error like:

```
edgedb error: Cannot parse environment variable "EDGEDB_PORT": invalid digit found in string
```

This is because Docker added environment variables like `EDGEDB_PORT=tcp://172.....:5656` when there is a link named `edgedb` like `docker run --link=edgedb edgedb/edgedb-cli`, and our CLI took that value and failed to parse it into an integer. This PR simply identifies the situation and ignores the wrong value.

Also updated edgedb-rust revision.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -480,6 +480,30 @@ fn env_fallback<T: FromStr>(value: Option<T>, name: &str)
     }
 }
 
+fn parse_port_env() -> anyhow::Result<(Option<String>, Option<u16>)> {
+    match get_env("EDGEDB_PORT") {
+        Ok(Some(value)) => if let Some(value) = value.strip_prefix("tcp://") {
+            // #389: `docker run --link=edgedb edgedb/edgedb-cli` will set env
+            // `EDGEDB_PORT` to `tcp://<host>:<port>` - we take it as default.
+            if let Some((host, port)) = value.split_once(":") {
+                let port = port.parse().with_context(|| {
+                    format!("Cannot parse port in Docker URI: EDGEDB_PORT")
+                })?;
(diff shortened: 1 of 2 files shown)


src/options.rs
```diff
@@ -518,7 +518,22 @@ fn conn_params(tmp: &ConnectionOptions) -> anyhow::Result<Builder> {
...
     let host = env_fallback(tmp.host.clone(), "EDGEDB_HOST")?;
-    let port = env_fallback(tmp.port, "EDGEDB_PORT")?;
+    let port = match env_fallback(tmp.port, "EDGEDB_PORT") {
+        Ok(port) => port,
+        Err(e) => {
+            if let Ok(Some(val)) = get_env("EDGEDB_PORT") {
+                if val.starts_with("tcp://") {
+                    // This is likely set by Docker for linked container which
+                    // happens to have the name "edgedb"
+                    None
```
elprans, line 528: Perhaps we should actually parse this instead of ignoring?

fantix: I thought about that - maybe it makes more sense to parse it in the Docker image `edgedb/edgedb-cli`?

elprans: Yeah, that's an even better idea.

fantix:
@tailhook is [this](https://github.com/edgedb/edgedb-cli/pull/389/commits/6a56874011ab8045627bec41673de4a804f8dd05) close to what you mentioned in the PM please?

tailhook (approved):
Yes. Looks good!

PR #390, opened 2021-07-26
Consistently check major_version for compare

fantix:
Also:
* Added `edgedb authenticate` successful message
* Fixed `1-beta*`/`1-rc*`/`1` major_version compare issue

### DIFF
--- a/src/server/version.rs
+++ b/src/server/version.rs
@@ -136,12 +136,14 @@ impl<T: AsRef<str>> Ord for Version<T> {
                 (None, Some(Numeric(_))) => Less,
                 (None, Some(String(x)))
                 if matches!(x, "a"|"b"|"c"|"rc"|"pre"|"dev"|"dirty")
+                || x.starts_with("beta") || x.starts_with("rc")
                 => Greater,
                 (None, Some(String(_))) => Less,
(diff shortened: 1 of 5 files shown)


src/server/version.rs
```diff
@@ -136,12 +136,14 @@ impl<T: AsRef<str>> Ord for Version<T> {
                 (None, Some(Numeric(_))) => Less,
                 (None, Some(String(x)))
                 if matches!(x, "a"|"b"|"c"|"rc"|"pre"|"dev"|"dirty")
+                || x.starts_with("beta") || x.starts_with("rc")
```
tailhook, line 139: I see this code was assuming that `beta` and `rc` are separate components. If they aren't we should probably change of component split. Otherwise it looks like `beta10 > beta2` will not hold.

tailhook (commented):
Looks like version comparison needs a bit more sophisticated fix.

src/server/version.rs
fantix: You're absolutely right. But I tried that - unless we also tries to match `beta` and `rc` in the component split, or else the `gfba00d7` case will fail. So considering we'll very unlikely get to beta 10, this PR is probably the easiest way out.

PR #391, opened 2021-07-26
Allow manually dispatch nightly event

fantix:
Although we had `repository_dispatch` with hook `nightly-build`


.github/workflows.src/nightly.tpl.yml
```diff
@@ -3,6 +3,8 @@ name: Build Test and Publish Nightly Packages
 on:
   schedule:
     - cron: "0 0 * * *"
+  workflow_dispatch:
+    inputs: {}
   repository_dispatch:
```
elprans, line 8: We can get rid of `repository_dispatch` now.

elprans (approved):

PR #392, opened 2021-07-26
Add `edgedb instance` subcommand

fantix:
Refs edgedb/rfcs#38, this PR only adds new calling paths for the
selected subcommands, deprecating old paths.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -177,6 +177,9 @@ pub enum Command {
     /// Manage local server installations
     #[edb(expand_help)]
     Server(server::options::ServerCommand),
+    /// Manage local server instances
+    #[edb(expand_help)]
+    Instance(server::options::ServerInstanceCommand),
     /// Manage project installation
     #[edb(expand_help)]
     Project(project::options::ProjectCommand),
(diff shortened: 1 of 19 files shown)


elprans (approved):
LGTM

fantix:
Merged first here - please feel free to leave further comments.

src/options.rs
```diff
@@ -318,53 +321,80 @@ fn get_matches(app: clap::App) -> clap::ArgMatches {
...
-                _ => {}
+                _ => e.exit(),
             }
-            e.exit();
         }
     }
 }
 
+fn get_deprecated_matches(mismatch_cmd: &str) -> Option<clap::ArgMatches> {
+    let mut args = env::args_os().skip(1);
```
tailhook, line 337: Unfortunately this may be a problem in case `edgedb --something instance init` or even `edgedb instance --something init`. I'm not sure how practical these uses currently, though.

src/server/control.rs
```diff
@@ -50,6 +50,7 @@ pub fn instance_command(cmd: &InstanceCommand) -> anyhow::Result<()> {
                 return status::print_status_all(c.extended, c.debug, c.json);
             }
         }
+        _ => panic!("Should not be here"),
```
tailhook, line 53: There is `unreachable!()` macro for exactly this case. But please add specific `Create` and `Destroy` cases here and mention where they were handled.

This is important that when a new command is added compiler highlights all the matches where it should be added.

src/server/control.rs
```diff
@@ -77,5 +78,6 @@ pub fn instance_command(cmd: &InstanceCommand) -> anyhow::Result<()> {
                 }
             }
         }
+        _ => panic!("Should not be here"),
```
tailhook, line 81: Same here.

src/options.rs
fantix: Ah you're right - let me see how to fix this.

PR #394, opened 2021-07-27
Implement `reset-password` for docker instances

tailhook:


elprans (approved):
LGTM.  Ideally, we need an integration test for this too.

fantix (approved):
Tested on macOS. `reset-password` should be `edgedb instance reset-password` though I think - I'll create separate PRs to undeprecate this command.

PR #399, opened 2021-07-28
Move `reset-password` to `instance` and undeprecate

tailhook:


fantix (approved):
Thanks! I think we were also discussing renaming `reset-password` to `reset-auth`

tailhook:
Ah, right! But I'm not sure how different the command will be. Feel free to submit follow up PR.

PR #400, opened 2021-07-28
Add `Method` column in `instance status`

### LINKED ISSUE #397: Make ` edgedb instance status` display install method
Add another column, "Installation"; can be set to "native" or "docker"

```
~ » edgedb instance status
┌──────┬───────┬─────────┬─────────────┐
│ Name │ Port  │ Version │   Status    │
├──────┼───────┼─────────┼─────────────┤
│ foo  │ 10700 │ 1-beta2 │ not running │
└──────┴───────┴─────────┴─────────────┘
```

While there, I think it's better to left-align text in headers of the ascii tables.

tailhook:
Fixes #397

### DIFF
--- a/src/server/status.rs
+++ b/src/server/status.rs
@@ -312,7 +312,7 @@ pub fn print_status_all(extended: bool, debug: bool, json: bool)
         let mut table = Table::new();
         table.set_format(*table::FORMAT);
         table.set_titles(Row::new(
-            ["Name", "Port", "Version", "Status"]
+            ["Name", "Port", "Version", "Installation", "Status"]
             .iter().map(|x| table::header_cell(x)).collect()));
         for status in statuses {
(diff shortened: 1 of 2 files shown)


src/server/status.rs
```diff
@@ -312,7 +312,7 @@ pub fn print_status_all(extended: bool, debug: bool, json: bool)
         let mut table = Table::new();
         table.set_format(*table::FORMAT);
         table.set_titles(Row::new(
-            ["Name", "Port", "Version", "Status"]
+            ["Name", "Port", "Version", "Method", "Status"]
```
1st1, line 315: I'm afraid "Method: native" might confuse users who are not using docker installs. That's why I suggested to use "Installation". Also, could you please left-align table headers?

fantix (approved):

PR #401, opened 2021-07-28
Move "existing instance" error earlier

### LINKED ISSUE #396: creating an instance with already taken name
Currently, when an instance is created with a name that's already taken, the error message is printed after a table with instance summary info is printed:

```
~ » edgedb instance create foo --nightly --method=docker
┌──────────────────────┬──────────────────────────────────────────────────────────┐
│ Instance Name        │ foo                                                      │
│ Mode                 │ Docker                                                   │
│ Data Directory       │ <docker volume "edgedb_foo">                             │
│ Credentials Path     │ /home/yury/.config/edgedb/credentials/foo.json           │
│ Database Server Port │ 10700                                                    │
│ Default User         │ edgedb                                                   │
│ Default Database     │ edgedb                                                   │
│ EdgeDB Version       │ nightly_1-beta3-dev202107231400_cv202107231400 (nightly) │
└──────────────────────┴──────────────────────────────────────────────────────────┘
Initializing EdgeDB instance...
edgedb error: Credential file /home/yury/.config/edgedb/credentials/foo.json already exists. This may mean that instance is already initialized. You may run `--overwrite` to overwrite the instance.
```

This is super confusing. The command should check if the name is taken right at the beginning and should print the error message right away, not displaying the summary table at all.

In general, the table should be displayed only when the instance was successfully created.

tailhook:
Fixes #396

### DIFF
--- a/src/server/create.rs
+++ b/src/server/create.rs
@@ -216,6 +216,15 @@ pub fn create(options: &Create) -> anyhow::Result<()> {
     let (distr, meth_name, method) = find_distribution(
         &*current_os, &avail_methods,
         &version_query, &options.method)?;
+    let credentials_path = credentials::path(&options.name)?;
+    if credentials_path.exists() && !options.overwrite {
+        anyhow::bail!("Credential file {} already exists. \
+            This may mean that instance is already initialized, \
+            or the name is taken as a link to a remote instance. \
+            Use different instance name or \
+            run `edgedb instance destroy {}` first",
+            credentials_path.display(), options.name);
+    }
     let port = allocate_port(&options.name)?;
     let settings = Settings {
         name: options.name.clone(),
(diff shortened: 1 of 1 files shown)


src/server/create.rs
```diff
@@ -216,6 +216,14 @@ pub fn create(options: &Create) -> anyhow::Result<()> {
     let (distr, meth_name, method) = find_distribution(
         &*current_os, &avail_methods,
         &version_query, &options.method)?;
+    let credentials_path = credentials::path(&options.name)?;
+    if credentials_path.exists() && !options.overwrite {
+        anyhow::bail!("Credential file {} already exists. \
+            This may mean that instance is already initialized. \
```
1st1, line 222: What else can it mean?

fantix: Probably we could say "This means the instance is already initialized, or the name is taken as a link to a remote instance"?

fantix (approved):
Other than Yury's question, I tested this branch and it works fine.

PR #404, opened 2021-07-28
Rename authenticate to instance link

fantix:
Also:

* Added connecting hint in link
* CRF #392: follow Rust best practice

TODO:

- [x] unlink
- [x] list links in `instance status`
- [x] update RFC 1006
- [x] update RFC 1008

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -377,6 +377,12 @@ impl Options {
             return Err(ExitCode::new(0).into());
         }
 
+        if tmp.subcommand.is_some() && tmp.query.is_some() {
+            anyhow::bail!(
+                "Option `-c` conflicts with specifying a subcommand"
+            );
+        }
+
         // TODO(pc) add option to force interactive mode not on a tty (tests)
         let interactive = tmp.query.is_none()
             && tmp.subcommand.is_none()
@@ -385,9 +391,7 @@ impl Options {
         let mut builder = conn_params(&tmp.conn);
 
         // Special case for `edgedb instance link`
-        if let (Some(Command::Instance(i)), None) = (
-            &tmp.subcommand, &tmp.query
-        ) {
+        if let Some(Command::Instance(i)) = &tmp.subcommand {
             if let server::options::InstanceCommand::Link(
                 link
             ) = &i.subcommand {
@@ -425,14 +429,9 @@ impl Options {
         });
 
         let subcommand = if let Some(query) = tmp.query {
-            if tmp.subcommand.is_some() {
-                anyhow::bail!(
-                    "Option `-c` conflicts with specifying a subcommand");
-            } else {
-                Some(Command::Query(Query {
-                    queries: vec![query],
-                }))
-            }
+            Some(Command::Query(Query {
+                queries: vec![query],
+            }))
         } else {
             tmp.subcommand
         };
--- a/src/options.rs
+++ b/src/options.rs
@@ -172,8 +172,6 @@ pub struct RawOptions {
 
 #[derive(EdbClap, Clone, Debug)]
 pub enum Command {
-    /// Authenticate to a remote instance
-    Authenticate(Authenticate),
     #[clap(flatten)]
     Common(Common),
     /// Execute EdgeQL query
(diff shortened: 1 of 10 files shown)


src/options.rs
```diff
@@ -409,15 +390,24 @@ impl Options {
             && atty::is(atty::Stream::Stdin);
 
         let mut builder = conn_params(&tmp.conn);
-        if let (Some(Command::Authenticate(auth)), None) = (&tmp.subcommand, &tmp.query) {
-            if builder.is_err() {
-                builder = Ok(Builder::new());
-                load_tls_options(&tmp.conn, builder.as_mut().unwrap())?;
+
+        // Special case for `edgedb instance link`
+        if let (Some(Command::Instance(i)), None) = (
```
tailhook, line 395: Oh, I've overlooked this previosly, but as we here: we don't need to check tuple here. `tmp.subcommand.is_some() && tmp.query.is_some()` is prohibited and it's checked below in `let subcommand =`.

src/options.rs
```diff
@@ -409,15 +390,24 @@ impl Options {
...
+            &tmp.subcommand, &tmp.query
+        ) {
+            if let server::options::InstanceCommand::Link(
+                link
+            ) = &i.subcommand {
+                if builder.is_err() {
```
tailhook, line 401: This should have more specific error check. So `conn_params` function should emit a specific error whan `edgedb.toml` not found. You can look at `InstanceNotFound` error as an example of custom error and it's matching.

src/options.rs
```diff
@@ -409,15 +390,24 @@ impl Options {
...
+                    builder = Ok(Builder::new());
+                    load_tls_options(&tmp.conn, builder.as_mut().unwrap())?;
+                }
+                server::link::prompt_conn_params(
```
tailhook, line 405: The only purpose of calling this function here is just to keep `tmp.conn` around, right? Let's just put `ConnectionOptions` into the final `Options` structure, and use it as a part of the normal command dispatch.

src/server/main.rs
```diff
@@ -26,13 +30,19 @@ pub fn main(cmd: &ServerCommand) -> Result<(), anyhow::Error> {
...
-pub fn instance_main(cmd: &ServerInstanceCommand) -> Result<(), anyhow::Error> {
+pub fn instance_main(cmd: &ServerInstanceCommand, options: &Options) -> Result<(), anyhow::Error> {
     use InstanceCommand::*;
 
     match &cmd.subcommand {
         Create(c) => create::create(c),
         Destroy(c) => destroy::destroy(c),
         ResetPassword(c) => reset_password::reset_password(c),
+        Link(c) => {
+            task::block_on(async {
```
tailhook, line 41: let's move `task::block_on` into `link::link` for uniformity.

tailhook (commented):
Sorry, most of my feedback could be contributed to the original `authenticate` proposal, but lets fix it as we are here.

src/options.rs
fantix: Fixed, so that we check and fail early before doing anything else.

src/options.rs
fantix: I'm sorry but I'm not sure I get your idea here - in case `edgedb.toml` is not found, `edgedb instance link ...` should just ask about connection parameters, no?

src/options.rs
fantix: Also reusing the password input/detection logic below.

src/options.rs
tailhook: Sure but there are few ways how `conn_params` can fail (so `builder is Err(..)`). For example if you passed `edgedb --dsn bad-proto://something` user should see failure instead of running interactive mode.

src/options.rs
tailhook: > Also reusing the password input/detection logic below.

It looks like that can be comfortably factored out to a function. I'm still feel uncomfortable when some commands are executed in the middle of the argument parsing (sans well-known ones like version and help).

tailhook (approved):

PR #406, opened 2021-07-29
Include remote instances in `instance status`

fantix:
Depends on #400

### DIFF
--- a/src/server/control.rs
+++ b/src/server/control.rs
@@ -7,6 +7,7 @@ use fn_error_context::context;
 use crate::credentials;
 use crate::server::detect;
 use crate::server::options::InstanceCommand;
+use crate::server::options::Status;
 use crate::server::metadata::Metadata;
 use crate::server::methods::Methods;
 use crate::server::revert;
@@ -69,49 +70,11 @@ pub fn instance_command(cmd: &InstanceCommand) -> anyhow::Result<()> {
         Logs(c) => inst?.logs(c),
         Revert(c) => revert::revert(inst?, c),
         Status(options) => match inst {
-            Ok(inst) => if options.service {
-                inst.service_status()
-            } else {
-                let status = inst.get_status();
-                if options.debug {
-                    println!("{:#?}", status);
-                    Ok(())
-                } else if options.extended {
-                    status.print_extended_and_exit();
-                } else if options.json {
-                    status.print_json_and_exit();
-                } else {
-                    status.print_and_exit();
-                }
-            },
+            Ok(inst) => local_status(options, inst),
             Err(e) => {
-                let path = credentials::path(name)?;
-                if !path.exists() {
-                    anyhow::bail!(e);
-                }
-                let status = task::block_on(
-                    status::RemoteStatus::new(name).probe(path)
-                );
-                if options.service {
-                    let path = credentials::path(name)?;
-                    println!("Remote instance: {}", path.display());
-                } else if options.debug {
-                    println!("{:#?}", status);
-                } else if options.extended {
-                    status.print_extended();
-                } else if options.json {
-                    println!(
-                        "{}",
-                        serde_json::to_string_pretty(&status.json())
-                            .expect("status is json-serializable"),
-                    );
-                } else if let Some(error) = status.status.get_error() {
-                    println!("{}: {}", status.status.display(), error);
-                } else {
-                    println!("{}", status.status.display());
-                }
-                status.exit()
-            }
+                remote_status(name, options)?;
+                Err(e)
+            },
         }
         | Create(_)
         | Destroy(_)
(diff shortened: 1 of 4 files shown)


src/server/status.rs
```diff
@@ -224,6 +242,101 @@ impl Status {
...
+                    ).await {
+                        version = Some(ver);
+                        status = "running";
+                    }
+                }
+                Ok(Err(e)) => {
+                    status = "error";
+                    error = Some(format!("{}", e));
+                }
+                Err(_) => {}  // timeout, ignore
```
tailhook, line 271: Do `log::info` for the error in the both cases. And why not `error = Some("timeout happened")` in this case?

src/server/status.rs
```diff
@@ -224,6 +242,101 @@ impl Status {
...
+            match future::timeout(
+                deadline - Instant::now() , builder.connect()
+            ).await {
+                Ok(Ok(mut con)) => {
+                    if let Ok(Ok(ver)) = future::timeout(
+                        deadline - Instant::now() , con.get_version()
+                    ).await {
+                        version = Some(ver);
+                        status = "running";
+                    }
```
tailhook, line 265: Also `log::info` for the error and `error =` in case of failure.

src/server/control.rs
```diff
@@ -54,7 +56,38 @@ pub fn instance_command(cmd: &InstanceCommand) -> anyhow::Result<()> {
     };
     let os = detect::current_os()?;
     let methods = os.get_available_methods()?.instantiate_all(&*os, true)?;
-    let inst = get_instance(&methods, name)?;
+    let inst = match get_instance(&methods, name) {
```
tailhook, line 59: This original `get_instance` was used for all commands not just status. This `RemoteStatus` code is supposed to be only for statuses so have be moved to `Status` case.

src/server/status.rs
```diff
@@ -278,15 +391,55 @@ pub fn print_status_all(extended: bool, debug: bool, json: bool)
...
+                            RemoteStatus::new(name.into())
+                        ))
+                    ));
+                }
+            }
+        }
+    }
+    let remote_statuses = task::block_on(async {
+        let mut rv: Vec<RemoteStatus> = Vec::new();
+        for (name, task) in tasks {
```
tailhook, line 426: This can be [paralellized](https://docs.rs/futures-util/0.3.16/futures_util/stream/trait.StreamExt.html#method.buffer_unordered) but we can keep for future improvements.

src/server/status.rs
```diff
@@ -278,15 +391,55 @@ pub fn print_status_all(extended: bool, debug: bool, json: bool)
...
+            match task.await {
+                Ok(Ok(status)) => rv.push(status),
+                Ok(Err(e)) => {
+                    eprintln!(
```
tailhook, line 430: It's better to use logging subsystem for this. So one can ignore some issues selectively.

src/server/status.rs
```diff
@@ -278,15 +391,55 @@ pub fn print_status_all(extended: bool, debug: bool, json: bool)
...
+                        "Cannot parse remote instance \"{}\": {:?}\
+                        \n    Is the credential file corrupted?",
+                        name, e
+                    );
+                }
+                Err(_) => {}  // timeout, ignore
```
tailhook, line 436: This is bad to ignore. Just insert a record with erroneous status.

tailhook (commented):
Generally `RemoteStatus::new` should be restructured:
1. `connect_and_get_version` should be combined to an async function, so you only use `timeout` once and don't track `deadline` manually
2. All errors (including timeouts) should rethrown right here with `?`
3. All errors are just gathered into `error`, along the lines of:

```rust
fn new(name: &str) -> RemoteStatus {
    ....
    let (status, error) = match get_status_real(builder) {
        Ok(status) => (status, None),
        Err(err) => (status, Some(err.to_string()),
    };
    return RemoteStatus {
       ...
       status, error,
    }
}
```

src/server/status.rs
```diff
@@ -224,6 +266,127 @@ impl Status {
...
+            status: RemoteStatusService::Unknown("uninitialized".into())
+        }
+    }
+
+    pub async fn probe(mut self, path: PathBuf) -> Self {
+        let deadline = Instant::now() + Duration::from_secs(1);
+        let builder = match async {
+            Ok::<Builder, anyhow::Error>(future::timeout(
+                deadline - Instant::now(),
+                Builder::read_credentials(&path)
```
tailhook, line 288: Generally we don't timeout disk functions. If your disk is super-slow you will not be able to use CLI or EdgeDB in any sensible way, so there is no need to put timeout here. And even less sense to subtract this time from the network time deadline because this will show network errors for perfectly working and responsive remote services.

src/server/status.rs
```diff
@@ -454,26 +451,27 @@ pub fn print_status_all(extended: bool, debug: bool, json: bool)
...
                 }
             }
         }
     }
+    let mut buffered_futs = stream::iter(futures).buffer_unordered(32);
     let remote_statuses = task::block_on(async {
         let mut rv: Vec<RemoteStatus> = Vec::new();
-        for task in tasks {
-            rv.push(task.await);
+        while let Some(status) = buffered_futs.next().await {
```
tailhook, line 473: I think you can do `buffer_unordered(32).collect::<Vec<_>>()`, or `buffer_unordered(32).collect::<Result<Vec<_>>>()`

tailhook (approved):
Take a look at `buffer_unordered` API again (see my comment). If you my hint doesn't help, feel free to merge it, we'll fix this nit later.

PR #408, opened 2021-07-29
Pass instance name to the server via env var

1st1:
This way the server instances can render a nice HTML page about
themselves.

Note: I'm not sure that what I'm doing here is correct. Few thoughts:

1. I initially wanted to add a command-line argument to `edgedb-server`, but that would make it harder for us to actually run it from the CLI: we'd have to somehow detect if the `edgedb-server` version we are about to run supports this new argument or not. With an environment variable we don't care; it's up to the server if it wants to use it -- but it's always there.

2. I've updated launchd and systemd entries to include the variable, as well as other places when we run the process, as well as when we create the container with docker.

### DIFF
--- a/src/server/docker.rs
+++ b/src/server/docker.rs
@@ -704,6 +704,8 @@ impl<'os, O: CurrentOs + ?Sized> DockerMethod<'os, O> {
                         options.port));
         cmd.arg(format!("--label=com.edgedb.metadata.start-conf={}",
                         options.start_conf));
+        cmd.env("EDGEDB_SERVER_INSTANCE_NAME", options.name);
+        cmd.arg("--env").arg("EDGEDB_SERVER_INSTANCE_NAME");
         cmd.arg(options.image.tag.as_image_name());
         cmd.arg("edgedb-server");
         cmd.arg("--runstate-dir").arg("/var/lib/edgedb/data/run");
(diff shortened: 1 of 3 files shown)


elprans (commented):
> I initially wanted to add a command-line argument

Add both.  We'll transition to using the argument after a while.  Command-line argument is nicer because it appears in process lists so you can easily do `ps | grep edgedb | grep instance-name`

1st1:
> Add both. 

OK: https://github.com/edgedb/edgedb/pull/2752

1st1:
I've updated the PR to also enable unsecure HTTP by default for servers managed with our CLI.

tailhook:
> enable unsecure HTTP by default for servers managed with our CLI.

Just to confirm: Is it just for deno?

tailhook (approved):

1st1:
> Just to confirm: Is it just for deno?

No, it's for all things. Basically, if you want to install EdgeDB on your devbox and shoot some GraphQL queries, currently you can't. Your browser will not let you because we have a self-signed TLS cert.

The idea is that in development, for HTTP specifically, we can allow non-encrypted HTTP. In prod you'd have to have a proper cert, issued by LE or some other provider. And of course, EdgeDB Cloud would automate that for you.

PR #409, opened 2021-07-29
Revise error messages

fmoor:
Skimmed source code checking for error message gramar and clarity. It is
dificult to do this accuratly without seeing the error messages in
context.

### DIFF
--- a/src/interactive.rs
+++ b/src/interactive.rs
@@ -157,14 +157,14 @@ fn _check_json_limit(json: &serde_json::Value, path: &mut String, limit: usize)
                 return false;
             }
             for (idx, item) in items.iter().enumerate() {
-                write!(path, "[{}]", idx).expect("formatting succeeds");
+                write!(path, "[{}]", idx).expect("formatting failed");
                 _check_json_limit(item, path, limit);
                 path.truncate(level);
             }
         }
         Object(pairs) => {
             for (key, value) in pairs {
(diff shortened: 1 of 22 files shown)


src/interactive.rs
```diff
@@ -157,14 +157,14 @@ fn _check_json_limit(json: &serde_json::Value, path: &mut String, limit: usize)
                 return false;
             }
             for (idx, item) in items.iter().enumerate() {
-                write!(path, "[{}]", idx).expect("formatting succeeds");
+                write!(path, "[{}]", idx).expect("formatting failed");
```
fmoor, line 160: I'm not sure I understood what `expect()` does. Many of the messages in `expect()` calls seemed to be the opposite of what it seamed they should be :confused:

tailhook: It writes something like:
```
expectation failed: formatting failed
expectation failed: formatting succeeds
```
So generally both forms read good enough for me, and I see both forms in other's Rust code. Note: this is printed along with more cryptic stuff and (hopefully) a backtrace and is treated as "assertion failure". So these errors are always bugs, and end users aren't supposed to understand what's going on. This is only for us to grep for the issue more quickly (especially if the is no stack trace).

tailhook (approved):

1st1 (approved):
Good stuff, Frederick.

PR #417, opened 2021-07-30
Refactor color highlighting of REPL output

1st1:
Refactor output highlighting; implement Tomorrow Night color scheme

* Refactor output highlighting to use the the same abstraction
    (`print::style::Styler`) that we use to highlight EdgeQL in input;
  This makes the input and output styling consistent.

* Cleanup and and new enum members to `print::style::Style`;

* Switch Styler to use xterm256 palette (non-RGB) for better
    compatibility (apparently macOS' default Terminal.app still doesn't
    support RGB);

* Implement the Tomorrow Night color scheme. It's not too bright and not
    too dark, should cater well to the majority of the users. It's also
    one of the most popular themes programmers use;

* Drop some dead code and unused `Style` enum members;

* Hide object name when it is "std::VirtualObject" -- that's
    anonumous shapes and showing a name there is confusing.

## Before

Screenshot from 2021-08-01 18-13-31

## After

Screenshot from 2021-08-01 18-14-10

### DIFF
--- a/src/highlight.rs
+++ b/src/highlight.rs
@@ -30,7 +30,7 @@ pub fn edgeql(outbuf: &mut String, text: &str, styler: &Styler) {
         }
         if let Some(st) = token_style(tok.token.kind, tok.token.value)
         {
-            styler.apply(st, tok.token.value, outbuf);
+            styler.write(st, tok.token.value, outbuf);
         } else {
             outbuf.push_str(tok.token.value);
         }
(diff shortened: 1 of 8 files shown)


src/highlight.rs
```diff
@@ -88,57 +88,68 @@ fn token_style(kind: Kind, value: &str) -> Option<Style> {
     use crate::print::style::Style as S;
 
     match kind {
-        T::Keyword => Some(S::Keyword),
+        T::Keyword => {
+            let lc = value.to_lowercase();
+            if lc == "true" || lc == "false" {
```
tailhook, line 93:
```suggestion
            if value.eq_ignore_ascii_case("true") || value.eq_ignore_ascii_case("false") {
```

`to_lowercase` is allocating, we try to ignore allocations if possible in hot paths.

tailhook (approved):

src/highlight.rs
1st1: got it, thanks!

PR #419, opened 2021-08-01
Rename `edgedb self` to `edgedb cli`

### LINKED ISSUE #416: Rename `edgedb self` to `edgedb cli`
(see https://github.com/edgedb/rfcs/blob/master/text/1006-simplified-cli.rst)

fantix:
Also reorganized `edgedb cli` related source files into `crate::cli` package.

The hidden command `edgedb _self_install` is kept unchanged to be compatible with the installation script at https://sh.edgedb.com.

Fixes #416


src/cli/cli_install.rs
```diff
@@ -94,7 +94,7 @@ pub struct Settings {
 fn print_long_description(settings: &Settings) {
     println!();
     print_markdown!("\
-        # Welcome to EdgeDB!\n\
```
tailhook, line 97: This is interesting. Does it have a better style? We should probably change the style instead?

src/cli/mod.rs
```diff
@@ -0,0 +1,10 @@
+pub mod cli_install;
```
tailhook, line 1: I would remove `cli_` from module names here. Looks like a duplication.

src/options.rs
```diff
@@ -362,6 +343,23 @@ fn get_deprecated_matches(mismatch_cmd: &str) -> Option<clap::ArgMatches> {
             }
             _ => return None,
         }
+        Some(first_cmd) if first_cmd == "self" => match args.next() {
```
tailhook, line 346: Shouldn't we skip `self` fallback, because there is no release which had these working?

tailhook (approved):

src/options.rs
fantix: Ah yes, fixed!

src/cli/cli_install.rs
fantix: Found a way to fix the alignment.

PR #421, opened 2021-08-02
Add 'edgedb info' command

1st1:
Displays the local paths EdgeDB installation is using.

macOS example: 

```
EdgeDB uses the following local paths:
┌────────────┬─────────────────────────────────────────────────────────────┐
│ Cache      │ /Users/fantix/Library/Caches/edgedb/                        │
│ Config     │ /Users/fantix/Library/Application Support/edgedb/           │
│ CLI Binary │ /Users/fantix/Library/Application Support/edgedb/bin/edgedb │
│ Data       │ /Users/fantix/Library/Application Support/edgedb/data/      │
│ Service    │ /Users/fantix/Library/LaunchAgents/                         │
└────────────┴─────────────────────────────────────────────────────────────┘
```


src/commands/info.rs
```diff
@@ -0,0 +1,45 @@
...
+{
+    let mut table = Table::new();
+
+    table.add_row(Row::new(vec![
+        Cell::new("Cache"),
+        Cell::new(dir_to_str(&platform::cache_dir()).as_ref()),
+    ]));
+    table.add_row(Row::new(vec![
+        Cell::new("Config"),
+        Cell::new(dir_to_str(&platform::config_dir()).as_ref()),
```
tailhook, line 36:
```suggestion
        Cell::new(platform::config_dir()?.display().to_string()),
```
This should be enough. Failing to find `config_dir` is so rare that it doesn't make too much sense to be to defensive here.

tailhook (commented):
Apart from displaying paths it's fine.

src/commands/info.rs
1st1: What about cache dir?

elprans (commented):
What about `data_dir` and the path to `edgedb-cli` itself?

src/commands/info.rs
tailhook: Same thing. They all kinda look for env var then look for `$HOME` then look for system home.

fantix:
`data_dir` is not present on Windows, I'll amend this PR with:

* `Data` (for Linux and macOS only)
* `Binary` for the directory containing `edgedb` executable

Do we want to also include:

* `Service` for the folder containing system service files created by `edgedb`
* `Installation` for installed EdgeDB server packages

elprans:
`Data` should still be printed even if it's the same as some other directory in the list.

src/commands/info.rs
```diff
@@ -0,0 +1,61 @@
...
+
+        table.add_row(Row::new(vec![
+            Cell::new("Service"),
+            Cell::new(&plist_dir(false)?.display().to_string()),
+        ]));
+    }
+
+    table.set_format(*table::FORMAT);
+
+    println!("EdgeDB uses the following local directories or paths:");
```
elprans, line 57: Just say "paths" and terminate directories in output with a directory separator to clarify.

elprans (approved):

1st1:
lgtm

src/commands/info.rs
```diff
@@ -0,0 +1,68 @@
...
+        Cell::new(&dir_to_str(platform::config_dir()?)),
+    ]));
+    if let Ok(current_exe) = env::current_exe() {
+        table.add_row(Row::new(vec![
+            if current_exe == platform::binary_path()? {
+                Cell::new("CLI Binary")
+            } else {
+                Cell::new("Custom Binary")
+            },
+            Cell::new(&current_exe.display().to_string()),
```
tailhook, line 36: I think we should put installation target too. So you can get `./edgedb cli info` and know installation path?

src/commands/info.rs
```diff
@@ -0,0 +1,68 @@
...
+        ]));
+    }
+    let data_dir = platform::data_dir()?;
+    if data_dir.exists() {
```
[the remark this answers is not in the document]
fantix: For Docker-only users or Windows users before we have WSL-native instances, data_dir is never created or used, so I thought it's not necessary to show it.

src/commands/info.rs
fantix: I'm sorry I'm not getting it - "installation target" as in `platform::binary_path()` or the installed *versions* of EdgeDB servers?

src/commands/info.rs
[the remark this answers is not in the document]
tailhook: I think `if cfg!(windows)` is better. For docker only users we might want to name directory better, but being able to tell the data path before creating the first instance is valuable too.

src/commands/info.rs
fantix: `./edgedb cli install` is already saying:

```
The edgedb binary will be placed in the user bin directory located at:
/Users/fantix/Library/Application Support/edgedb/bin
```

It's a bit redundant to have `./edgedb cli info` I think?

tailhook: In that sense displaying `current_exe()` makes even smaller sense because you can run `which edgedb`.

fantix: OK I think you have two proposals here:

1. Add `edgedb cli info` to show `platform::binary_path()` regardless of its existance - I don't think we need this because `./edgedb cli install` will show that path anyways.
2. Drop `Custom Binary` in `edgedb info` and always show `platform::binary_path()` as `Binary` regardless of its existance.

If my understanding is correct, I'll fix (2) in a following PR.

PR #424, opened 2021-08-03
Remove `allow-non-empty` option

tailhook:


fantix (approved):

src/commands/restore.rs
```diff
@@ -113,26 +113,18 @@ pub async fn restore<'x>(cli: &mut Connection, options: &Options,
...
-            if options.command_line {
-                return Err(anyhow::anyhow!("\
-                    cannot restore: the database is not empty; \
-                    consider using the --allow-non-empty option"));
-            } else {
-                return Err(anyhow::anyhow!(
-                    "cannot restore: the database is not empty"));
-            }
-        }
+    if is_empty_db(cli).await.context("Error checking if the DB is empty")? {
```
1st1, line 125: if !is_empty?

src/commands/restore.rs
```diff
@@ -113,26 +113,18 @@ pub async fn restore<'x>(cli: &mut Connection, options: &Options,
...
+        return Err(anyhow::anyhow!("\
+            cannot restore: the database is not empty"));
     }
 
     // TODO(tailhook) check that DB is empty
```
elprans, line 130: This TODO entry seems to be made done with this diff.

src/commands/restore.rs
tailhook: Well, yes, misleading the name of the function. Fixed.

1st1 (approved):

PR #427, opened 2021-08-03
instance link --non-interactive default to untrust

fantix:
`--non-interactive` will no longer trust the peer certificate by
default. Instead, new option `--trust` option should be used together
with `--non-interactive`. `--trust` can also be used in interactive mode
to avoid the question.

Also added `--overwrite` to manage the existing credential file overwriting.


tailhook (approved):
Looks good. But I'm not sure about `--trust` name. Probably better to bikeshed in slack.

PR #428, opened 2021-08-03
Fix typo, update instance messages

colinhacks:
Fixes "does not exists"

Also removes several usages of "server instance" which I think is less clear than "EdgeDB instance" or just "instance"

### DIFF
--- a/src/server/options.rs
+++ b/src/server/options.rs
@@ -55,13 +55,13 @@ pub enum InstanceCommand {
 
 #[derive(EdbClap, Clone, Debug)]
 pub enum Command {
-    /// Show locally installed EdgeDB servers
+    /// Show locally installed EdgeDB versions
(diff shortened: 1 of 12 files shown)


elprans (approved):

src/server/options.rs
```diff
@@ -332,7 +332,7 @@ pub struct Revert {
 
 #[derive(EdbClap, Debug, Clone)]
 pub struct ResetPassword {
-    /// Database server instance name
+    /// Name of instance
```
1st1, line 335: Are we missing "the" here?

1st1 (approved):
@colinhacks I've left one nit comment/question, other than that looks great

src/server/options.rs
colinhacks: Updated

PR #431, opened 2021-08-04
EdbClap: Don't add from_str check if has validator

fantix:
This fixes the issue that invalid instance names can be used in `instance create` etc.

Also fixed _localdev check logic with proper error message.

Refs #356

### DIFF
--- a/edgedb-cli-derive/src/into_app.rs
+++ b/edgedb-cli-derive/src/into_app.rs
@@ -237,11 +237,6 @@ fn mk_arg(field: &types::Field, case: &Case) -> TokenStream {
             #arg = #arg.value_name(#name);
         });
     }
-    for (name, value) in &field.attrs.options {
-        modifiers.extend(quote! {
-            #arg = #arg.#name(#value);
-        });
-    }
     match field.parse.kind {
         ParserKind::TryFromStr => {
             let ty = &field.ty;
@@ -270,6 +265,13 @@ fn mk_arg(field: &types::Field, case: &Case) -> TokenStream {
             });
         }
         _ => {}
+    }
+    // The arbitrary options must be added in the end so that e.g. explicit
+    // validator() could overwrite the validators added by default previously
+    for (name, value) in &field.attrs.options {
+        modifiers.extend(quote! {
+            #arg = #arg.#name(#value);
+        });
     }
 
     return quote! {
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -52,5 +52,5 @@ pub fn is_valid_name(name: &str) -> bool {
             return false;
         }
     }
-    return name != "_localdev"
+    return true;
 }
--- a/src/server/options.rs
+++ b/src/server/options.rs
@@ -188,6 +188,8 @@ assign an instance name to simplify future connections.")]
 pub struct Link {
     /// Specify a new instance name for the remote server. If not
     /// present, the name will be interactively asked.
+    #[clap(validator(instance_name_opt))]
+    #[clap(value_hint=ValueHint::Other)]
     pub name: Option<String>,
 
     /// Run in non-interactive mode (accepting all defaults)
@@ -211,6 +213,8 @@ pub struct Link {
 #[clap(long_about = "Unlink from a remote EdgeDB instance.")]
 pub struct Unlink {
     /// Specify the name of the remote instance.
+    #[clap(validator(instance_name_opt))]
+    #[clap(value_hint=ValueHint::Other)]  // TODO complete instance name
     pub name: String,
 }
 


src/project/init.rs
```diff
@@ -188,6 +188,9 @@ fn ask_name(methods: &Methods, dir: &Path, options: &Init)
             if confirm.ask()? {
                 return Ok((target_name, true));
             }
+        } else if is_local_dev(&target_name) {
+            eprintln!("no remote instance _localdev found, \
```
tailhook, line 192: I think it's more error of a kind `_localdev instance name is reserved. You can't initialize instance with this name, but you can link existing instance`. Because otherwise it's not clear why you have `to link first`. And we don't want users to link `_localdev` we want them to run `edb server`.

edgedb-cli-derive/src/into_app.rs
```diff
@@ -242,34 +242,37 @@ fn mk_arg(field: &types::Field, case: &Case) -> TokenStream {
...
-            let func = if let Some(func) = &field.parse.parser {
-                quote! { #func }
-            } else {
-                quote! { std::str::FromStr::from_str }
-            };
-            modifiers.extend(quote! {
-                #arg = #arg.validator(|v| {
-                    #func(v).map(|_: #ty| ())
+    let validator = syn::Ident::new("validator", Span::call_site());
+    if !field.attrs.options.contains_key(&validator) {
```
tailhook, line 246: We must obey `clap` semantics here. Is it what clap does? I'm also not sure why this is needed.

tailhook: Also if we do this check, we should put a flag in `field.attrs` instead of using `contains_key`, as everything else is done.

src/server/mod.rs
```diff
@@ -52,5 +52,9 @@ pub fn is_valid_name(name: &str) -> bool {
             return false;
         }
     }
-    return name != "_localdev"
+    return true;
+}
+
+pub fn is_local_dev(name: &str) -> bool {
```
tailhook, line 58: Not sure if we need this function.  `if name == "_localdev"` reads even cleaner than `if is_local_dev(name)`

src/server/unix.rs
```diff
@@ -250,16 +251,23 @@ pub fn instances_from_data_dir(dir: &Path, system: bool,
         if !item.file_type()?.is_dir() {
             continue;
         }
-        if let Some(name) = item.file_name().to_str() {
-            if !is_valid_name(name) {
-                continue;
-            }
-            instances.insert((name.to_owned(), system));
+
+        if let Some(name) = instance_name_from_os_str(&item.file_name()) {
```
tailhook, line 255: Why we are skipping `_localdev` here?

src/server/macos.rs
```diff
@@ -315,7 +315,22 @@ impl<'os> Method for PackageMethod<'os, Macos> {
         -> anyhow::Result<InstanceRef<'x>>
     {
         let dir = unix::storage_dir(name)?;
-        if dir.exists() {
+        let name_os = dir.file_name().ok_or_else(|| {
```
tailhook, line 318: This checks if storage dir is not root? This doesn't sound like we need it (`storage_dir` function explicitly attaches the name).

src/server/linux.rs
```diff
@@ -426,7 +426,22 @@ pub fn get_instance<'x>(method: &'x dyn Method, name: &str)
...
+    let name_os = dir.file_name().ok_or_else(|| {
+        InstanceNotFound(
+            anyhow::anyhow!("Invalid path: '{}'", dir.display())
+        )
+    })?;
+    if !dir.exists() {
+        Err(InstanceNotFound(
+            anyhow::anyhow!("Directory '{}' does not exist", dir.display())
+        ).into())
+    } else if unix::instance_name_from_os_str(name_os).is_none() {
```
tailhook, line 438: I'm not sure why this is needed? `name` == `name_os` here. And it's a contract that this should be a valid name here. We check for that either in CLI parser, or in directory listing.

If you think there is a bug somewhere, you can probably put an assertion. And assertion has to be before `dir.join()` otherwise this can be very interesting security issue.

src/server/linux.rs
```diff
@@ -426,7 +426,22 @@ pub fn get_instance<'x>(method: &'x dyn Method, name: &str)
     -> anyhow::Result<InstanceRef<'x>>
 {
     let dir = data_dir()?.join(name);
-    if dir.exists() {
+    let name_os = dir.file_name().ok_or_else(|| {
```
tailhook, line 429: Same as in MacOS. What we are checking here? That join works?

src/server/macos.rs
```diff
@@ -315,7 +315,22 @@ impl<'os> Method for PackageMethod<'os, Macos> {
...
+            InstanceNotFound(
+                anyhow::anyhow!("Invalid path: '{}'", dir.display())
+            )
+        })?;
+        if !dir.exists() {
+            Err(InstanceNotFound(
+                anyhow::anyhow!("Directory '{}' does not exist", dir.display())
+            ).into())
+        } else if unix::instance_name_from_os_str(name_os).is_none() {
```
tailhook, line 327: Same comment as in linux version.

tailhook (commented):
I see almost all of it exccessive. Just add few `name == '_localdev`` in needed places. Or am I missing some context?

edgedb-cli-derive/src/into_app.rs
fantix: Because we [put a "ParserKind::TryFromStr" for every field that is bool or didn't define a `parse()`](https://github.com/edgedb/edgedb-cli/pull/356/files#diff-ba3f1dea2ab02cd6ab3fa59af7fa5d5e3b4eb9de9c6a8a3756c4f6f7b657d3fbR77-R87), so we'll always append a `validator()` even when `validator()` is explicitly specified. Looks like the second `validator()` will overwrite the first, and that's why `validator(instance_name_opt)` stopped working. I'm open to fixing this in any proper way.

src/server/unix.rs
fantix: Because `edb server` will use that directory, and it's not managed by the CLI. We skip `_localdev` here so that it doesn't show up in `edgedb instance status` as a local instance; instead `edb cli` will create a remote instance which will show up in the status list properly.

src/server/macos.rs
fantix: Right, I'll `unwrap()` directly then.

src/server/linux.rs
fantix: Shoot, forgot we have name here ... this check is to make sure _localdev is not treated as a local instance, therefore `instance unlink _localdev` could run properly.

edgedb-cli-derive/src/into_app.rs
tailhook: Ah, yes. I think we should just move the block that applies arbitrary options (lines 240-244) after applying the parser. So replacing validator just works as expected.

src/server/linux.rs
tailhook: If there already is an instance named `_localdev` then users should be warned on unlink and should be able to destroy it. There is no reason to filter out it here.

edgedb-cli-derive/src/into_app.rs
fantix: Yeah sure, that works too.

src/server/unix.rs
fantix: Oh the creation is not a bug - `edb server` without `--data-dir` will just use that directory.

src/server/linux.rs
fantix: Per discussion, we won't reserve `_localdev` in this PR.

tailhook (approved):

PR #432, opened 2021-08-04
Rename --no-version-check to --no-cli-update-check (per the RFC)

1st1:
While there edit a few error/info messages.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -159,9 +159,14 @@ pub struct RawOptions {
     #[clap(short='V', long="version")]
     pub print_version: bool,
 
-    /// Disable version check
+    // (deprecated in favor of "no_cli_update_check")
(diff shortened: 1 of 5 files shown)


tailhook (approved):

src/options.rs
```diff
@@ -408,7 +424,7 @@ impl Options {
             } else {
                 OutputMode::JsonElements
             },
-            no_version_check: tmp.no_version_check,
+            no_version_check: no_version_check,
```
fantix, line 427: nit:

```suggestion
            no_version_check,
```

would be even better if we rename the variable and field name too for consistency.

fantix (approved):

src/options.rs
1st1: Updated!

PR #443, opened 2021-08-05
Error out upgrading linked instance

fantix:
A few sample errors:

```
/tmp $ edgedb instance upgrade ccc --to-version 1-beta2
Instance "ccc" is used by the following project:
  /Users/fantix/workspace/test
To continue with the upgrade, run:
  edgedb project upgrade --to-version 1-beta2 --project-dir '/Users/fantix/workspace/test'
edgedb error: Upgrade aborted.
```

```
~/workspace/test $ edgedb instance upgrade ccc --to-version 1-beta2
Instance "ccc" is used by the following project:
  /Users/fantix/workspace/test
To continue with the upgrade, run:
  edgedb project upgrade --to-version 1-beta2
edgedb error: Upgrade aborted.
```

```
~/workspace/test $ edgedb instance upgrade ccc --to-version 1-beta2
Instance "ccc" is used by the following projects:
  /Users/fantix/workspace/test
  /Users/fantix/workspace/test2
To continue with the upgrade, run:
  edgedb project upgrade --to-version 1-beta2
  edgedb project upgrade --to-version 1-beta2 --project-dir '/Users/fantix/workspace/test2'
edgedb error: Upgrade aborted.
```

I know @vpetrovykh convinced the team that `--force` should not override the behavior, but `--force` is essentially the same thing when you have multiple projects linking to the same instance, and the user should be warned to upgrade other projects too when upgrading from one of the projects.

### DIFF
--- a/src/server/upgrade.rs
+++ b/src/server/upgrade.rs
@@ -95,7 +95,14 @@ pub fn upgrade(options: &Upgrade) -> anyhow::Result<()> {
         if !project_dirs.is_empty() {
             destroy::print_instance_in_use_warning(name, &project_dirs);
             let current_project = project::project_dir_opt(None)?;
-            eprintln!("To continue with the upgrade, run:");
+            if options.force {
+                eprintln!(
+                    "To update the project{} after the instance upgrade, run:",
+                    if project_dirs.len() > 1 { "s" } else { "" }
+                );
+            } else {
+                eprintln!("To continue with the upgrade, run:");
+            }
             for pd in project_dirs {
                 let pd = destroy::read_project_real_path(&pd)?;
                 eprintln!(
--- a/src/server/upgrade.rs
+++ b/src/server/upgrade.rs
@@ -12,6 +12,7 @@ use crate::commands;
 use crate::connect::Connector;
 use crate::hint::HintExt;
 use crate::process::ProcessGuard;
+use crate::project;
 use crate::server::destroy;
 use crate::server::detect::{self, VersionQuery};
 use crate::server::errors::InstanceNotFound;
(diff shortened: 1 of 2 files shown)


elprans:
Hmm. I didn't realize that linking a single instance to multiple projects is allowed.  What happens when I `project upgrade` one of the linked projects and not the other.  Do they become decoupled into two instances?

fantix:
Nope, the same instance gets upgraded and the toml file for the other project gets outdated.

src/server/upgrade.rs
```diff
@@ -89,6 +89,24 @@ fn interpret_options(options: &Upgrade) -> anyhow::Result<ToDo> {
 
 pub fn upgrade(options: &Upgrade) -> anyhow::Result<()> {
     let todo = interpret_options(&options)?;
+    if let ToDo::InstanceUpgrade(name, version) = &todo {
+        let project_dirs = destroy::find_project_dirs(name)?;
+        if !project_dirs.is_empty() {
+            destroy::print_instance_in_use_warning(name, &project_dirs);
+            eprintln!("Go to each project directory and run:");
```
elprans, line 96: Multiple projects per instance is an exception, not the rule, so let's specialize the message to the common use case.  Also, for better DX, let's use `--project-dir` instead of asking the user to `cd` manually (probably only if the current directory isn't already that of the project).

src/server/upgrade.rs
```diff
@@ -89,6 +90,36 @@ fn interpret_options(options: &Upgrade) -> anyhow::Result<ToDo> {
 
 pub fn upgrade(options: &Upgrade) -> anyhow::Result<()> {
     let todo = interpret_options(&options)?;
+    if let ToDo::InstanceUpgrade(name, version) = &todo {
+        let project_dirs = destroy::find_project_dirs(name)?;
+        if !project_dirs.is_empty() {
+            destroy::print_instance_in_use_warning(name, &project_dirs);
+            let current_project = project::project_dir_opt(None)?;
+            eprintln!("To continue with the upgrade, run:");
```
tailhook, line 98: I think this message should be changed when force is active. Because writing `To continue with update` and then actually doing the upgrade is confusing.

tailhook (approved):
Other than a little confusing message looks good.

PR #448, opened 2021-08-06
Add an about section to `--help`

1st1:
With this patch:

```
EdgeDB CLI 

The `edgedb` command allows to open a REPL to the database,
manage projects, spin up local instances, create migrations,
and many other things.

Running `edgedb` without a subcommand opens a REPL.

USAGE:
    edgedb [FLAGS] [SUBCOMMAND]

FLAGS:
    -h, --help                   Prints help information
        --help-connect           Print all connection options for REPL and subcommands
    -V, --version                Show command-line tool version
        --no-cli-update-check    Disable checking if a new version of CLI is available

```

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -30,12 +30,18 @@ static CONNECTION_ARG_HINT: &str = "\
     Run `edgedb project init` or use any of `-H`, `-P`, `-I` arguments \
     to specify connection parameters. See `--help` for details";
 
+const CONN_OPTIONS_GROUP: &str =
+    "CONNECTION OPTIONS (`edgedb --help-connect` to see the full list)";
+
+const EDGEDB_ABOUT: &str = "\
+    Use the `edgedb` command-line tool to spin up local instances,\n\
+    manage EdgeDB projects, create and apply migrations, and more.\n\
+    \n\
+    Running `edgedb` without a subcommand opens an interactive shell.";
+
 pub trait PropagateArgs {
     fn propagate_args(&self, dest: &mut AnyMap, matches: &clap::ArgMatches);
 }
-
-const CONN_OPTIONS_GROUP: &str =
-    "CONNECTION OPTIONS (`edgedb --help-connect` to see the full list)";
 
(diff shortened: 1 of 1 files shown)


src/options.rs
```diff
@@ -30,13 +30,20 @@ static CONNECTION_ARG_HINT: &str = "\
     Run `edgedb project init` or use any of `-H`, `-P`, `-I` arguments \
     to specify connection parameters. See `--help` for details";
 
+const CONN_OPTIONS_GROUP: &str =
+    "CONNECTION OPTIONS (`edgedb --help-connect` to see the full list)";
+
+const EDGEDB_ABOUT: &str = "\
+    The `edgedb` command allows to open a REPL to the database,\n\
```
elprans, line 37:
```suggestion
    The `edgedb` command provides an interactive front-end to EdgeDB instances,\n\
```

src/options.rs
```diff
@@ -30,13 +30,20 @@ static CONNECTION_ARG_HINT: &str = "\
...
+    manage projects, spin up local instances, create migrations,\n\
```
elprans, line 38:
```suggestion
    allows managing projects, local instances, migrations,\n\
```

src/options.rs
```diff
@@ -30,13 +30,20 @@ static CONNECTION_ARG_HINT: &str = "\
...
+    and many other things.\n\
+    \n\
+    Running `edgedb` without a subcommand opens a REPL.";
```
elprans, line 41: How widespread is the understanding of REPL?  Python and Postgres call theirs "interactive shell" or "interactive terminal".

1st1: You have a point. "Interactive shell" sounds way better than "interactive front-end" though.

1st1:
@elprans @colinhacks I've updated the wording, take another look.

1st1:
> Run `edgedb` inside a project directory to open an interactive shell.

I don't want to imply that shell would only open when you're inside a project directory, so the new wording is

"Running `edgedb` without a subcommand opens an interactive shell."

PR #452, opened 2021-08-09
Stop using markdown to render clap help.

### LINKED ISSUE #450: Disable markdown styling in `--help`
It looks like markdown styling causes some artifacts in terminals. Most likely because text wrap happens after the escape codes for formatting are added, but wrapping and styling properly at the same time sounds like a complicated enough problem. So I propose simply disabling styling of markdown for now in terminal.

Screenshot from 2021-08-07 11-21-12

1st1:
The reason is unwanted rendering artifacts in some terminals causes by
clap wrapping the text after markdown parsing/styling is done. Perhaps
we can figure out how to actually fix this rendering issue in a better
way, but until then it's safer to just disable the feature.

Closes #450.


tailhook:
https://github.com/edgedb/edgedb-cli/issues/450#issuecomment-895365173

PR #453, opened 2021-08-09
Restyle markdown help formatting to not use background colors

### LINKED ISSUE #450: Disable markdown styling in `--help`

1st1:
While there, colorize the "about" section of `--help`.
The foreground color is also removed in favor of "bold" style.

Fixes #450; superseedes #452.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -951,6 +951,7 @@ dependencies = [
  "dirs 3.0.2",
  "downcast-rs",
  "edgedb-cli-derive",
+ "edgedb-cli-md",
  "edgedb-client",
  "edgedb-derive",
  "edgedb-protocol",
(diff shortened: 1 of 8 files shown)


1st1:
Output:

Screenshot from 2021-08-09 14-27-50

tailhook (approved):

elprans (approved):

1st1:
@tailhook I looked into generalizing this work and making it apply markdown styling after the derive phase, but I don't think it's possible; the problem is that `clap::App::about()` expects a static lifetime for the about string which I can't figure out how make... I'll merge this as is, but a follow up PR is necessary.

1st1:
See also https://github.com/clap-rs/clap/issues/2150

PR #456, opened 2021-08-10
Make `migration create` use the question crate

### LINKED ISSUE #429: Interactive prompts are not consistent
`edgedb project init` uses `>` as a user input prompt while `edgedb migration create` doesn't print any prompt at all. Perhaps the `>` prompt should be used every time user input is solicited interactively. Or at least the prompt should be consistent. 

- EdgeDB CLI Version: EdgeDB CLI 1.0.0-beta.2+d20210803.g62dbcf9e8
- OS Version: Linux 5.11.0-25-generic #27-Ubuntu SMP Fri Jul 9 23:06:29 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

Steps to Reproduce:

1. `edgedb project init`
2. `edgedb migration create`

1st1:
Fixes #429

### DIFF
--- a/src/migrations/create.rs
+++ b/src/migrations/create.rs
@@ -30,6 +30,7 @@ use crate::migrations::print_error::print_migration_error;
 use crate::migrations::prompt;
 use crate::migrations::source_map::{Builder, SourceMap};
 use crate::platform::tmp_file_name;
+use crate::question;
 
 const SAFE_CONFIDENCE: f64 = 0.99999;
 
@@ -40,6 +41,7 @@ pub enum SourceName {
     File(PathBuf),
 }
 
+#[derive(Clone, Debug)]
 pub enum Choice {
     Yes,
     No,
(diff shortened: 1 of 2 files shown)


1st1:
The consequence of this change is that by default we only show the short prompt now; a user needs to type `?` to show the help, which I think is a good thing (less noise).

Screenshot from 2021-08-10 12-46-24

1st1:
I've also added colors & updated rendering of `h`

Screenshot from 2021-08-10 13-58-37

elprans (approved):
LGTM

1st1:
Fixed "confirm" to "confirmed" with my recent force push

fantix (approved):
Nice!

1st1:
@tailhook please take a look tomorrow at this. The changes are fairly trivial, but the UX improvements are nice and I decided to merge it now to be part of the release build.

tailhook:
Using blocking in async can cause troubles in future, but currently there is nothing that depends on that, so it's fine. Other than that it's good.

PR #460, opened 2021-08-11
migration create: Rename "split" command back to "stop"

1st1:
The "split" alias is actually confusing so drop it.


elprans (approved):

tailhook:
For me stop == quit. Difference is very non-obvious.

PR #468, opened 2021-08-11
Actually create the env file when doing `edgedb cli migrate`

elprans:
Since we suggest to `source .../env` in the migrate message, make sure
the file actually exists.


1st1 (approved):
So we attempt to write `source .../env` to `.profile` etc?

fantix (approved):

elprans:
> So we attempt to write `source .../env` to `.profile` etc?

No, we write/modify `PATH` declaration there directly.  `env` is essentially an excerpt from that, because sourcing entire `.profile` might cause breakage.

PR #469, opened 2021-08-11
Highlight command results

fantix:
Also switched to use the question crate.

### DIFF
--- a/src/print/mod.rs
+++ b/src/print/mod.rs
@@ -380,3 +380,10 @@ pub fn success_msg(title: &str, msg: &str) {
         msg.bold().white(),
     )
 }
+
+pub fn warn(line: &str) {
+    eprintln!(
+        "{}",
+        line.bold().yellow(),
+    );
+}
--- a/src/print/mod.rs
+++ b/src/print/mod.rs
@@ -4,7 +4,7 @@ use std::io;
 use std::convert::Infallible;
 
 use async_std::stream::{Stream, StreamExt};
-use colorful::Colorful;
+use colorful::{Color, Colorful};
 use snafu::{Snafu, ResultExt, AsErrorSource};
 
 mod native;
@@ -338,4 +338,45 @@ pub fn json_item_to_string<I: FormatExt>(item: &I, config: &Config)
     }
     prn.end().unwrap_exc()?;
     Ok(out)
+}
+
+pub fn prompt(line: &str) {
+    println!(
+        "{}",
+        line.bold().color(Color::Orange3),
+    );
+}
+
+pub fn error(line: &str) {
+    eprintln!(
+        "{}",
+        line.bold().light_red(),
+    );
+}
+
+pub fn error_msg(title: &str, msg: &str) {
+    eprintln!(
+        "{}: {}",
+        title.bold().light_red(),
+        msg.bold().white(),
+    );
+}
+
+pub fn error_obj(e: &dyn fmt::Display) {
+    error_msg("edgedb error", &format!("{:#}", e));
+}
+
+pub fn success(line: &str) {
+    println!(
+        "{}",
+        line.bold().light_green(),
+    )
+}
+
+pub fn success_msg(title: &str, msg: &str) {
+    println!(
+        "{}: {}",
+        title.bold().light_green(),
+        msg.bold().white(),
+    )
 }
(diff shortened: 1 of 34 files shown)


src/server/upgrade.rs
```diff
@@ -138,17 +138,20 @@ pub fn upgrade(options: &Upgrade) -> anyhow::Result<()> {
...
-                    print_markdown!(
-                        "**EdgeDB instance '${name}' was successfully \
-                        upgraded to version ${version}!**",
-                        name=name,
-                        version=new_inst.get_current_version()?.unwrap(),
+                    let version = new_inst.get_current_version()?.unwrap();
+                    print::success_msg(
+                        &format!("Successfully upgraded EdgeDB instance \
+                                      '{}' to version", name),
+                        &format!("{}", version)
```
1st1, line 154: I like this way better than abusing markdown for simple status messages. I think there were a few more places with newly added `print_markdown()` calls for that in `master`. Can you rebase and take a look?

1st1 (approved):

src/server/upgrade.rs
tailhook: I'm not sure I like it more. Markdown gives more flexibility, and doesn't require tons of similar functions like `success` vs `success_msg`. The problem with markdown is that it has limited set of styles, unfortunately, so it's not great either :(

src/print/mod.rs
```diff
@@ -340,3 +340,44 @@ pub fn json_item_to_string<I: FormatExt>(item: &I, config: &Config)
...
+    eprintln!(
+        "{}: {}",
+        title.bold().light_red(),
+        msg.bold().white(),
+    );
+}
+
+pub fn error_obj(e: &dyn fmt::Display) {
+    error_msg("edgedb error", &format!("{:#}", e));
+}
```
tailhook, line 368: I think we wanted to prefix all error messages with `edgedb error` specifically, so that it's clear where error comes from in scripts. So I'm not sure what `error_msg` can be used for later. And why `error` doesn't have this prefix.

tailhook (commented):
I think these functions should accept `impl AsRef<str>` or `impl Display` instead of `&str`. So you don't have to prefix `&format!()` with ampersand in arguments. Probably latter is better, the compromise here is: `AsRef` is cheaper, while `Display` will require `.to_string()`, before applying color styles (which will clones `&'str` received), but also allows passing any other displayable objects. I'm pretty sure performance is not important here and can be fixed by using better colorizing API later.

src/server/upgrade.rs
1st1: > The problem with markdown is that it has limited set of styles

This, exactly.

I'm not against using markdown, I just think we should use it wisely. IMO having a function to print successful messages in green color is better than using `print_markdown!("# success")`, at the very least for better readability of the code. But up to you, I'm not trying to impose my opinion here too much.

tailhook (approved):

fantix:
@tailhook @1st1 thanks for approving! I just pushed another [quick fix](https://github.com/edgedb/edgedb-cli/pull/469/commits/44d512f0996ca21eb769bdcab0bce031497549ff) adding an `EDGEDB_USE_COLOR` env var and fixed the tests. If that looks okay, I'll click the merge button?

1st1:
> I just pushed another quick fix adding an EDGEDB_USE_COLOR

I'm not sure we need an env var for this. I'd prefer adding `edgedb --no-color` global argument.

tailhook:
I think we should use `CLICOLOR` convention https://github.com/mitsuhiko/clicolors-control/blob/master/src/common.rs#L13-L16 instead of adding our own edgedb-prefixed one.

1st1:
> I think we should use CLICOLOR convention https://github.com/mitsuhiko/clicolors-control/blob/master/src/common.rs#L13-L16 instead of adding our own edgedb-prefixed one.

Sure, let's use it, +1. @fantix can you update the PR?

fantix:
@1st1 @tailhook rebased and fixed at https://github.com/edgedb/edgedb-cli/pull/469/commits/16b93bac8213f8e388a723c03bbbd0230bd1ec0e

src/print/mod.rs
fantix: Added `edgedb error` for all errors - and removed the `error_msg()` function.

tailhook (approved):

PR #470, opened 2021-08-11
Output stdout/stderr of spawned subprocesses in grey

1st1:
Decided to take a stab at this just to see if I can implement something close to what we want.

Few notes:

* In this PR `proecss::run()` proxies output for all processes run with it.
* Instead it should probably get another explicit argument to do that, and then it should selectively be enabled only for some commands.

Unfortunately I'm not familiar enough with the code base to know what subprocess calls need to be instrumented and which ones just better be left as is.

@tailhook feel free to use this as a foundation for your PR or just feel free to close.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -1,6 +1,7 @@
 use std::io;
-use std::process::{Command, Child, exit};
+use std::process::{Command, Child, exit, Stdio};
 use std::time::Duration;
+use std::thread;
 
 use anyhow::Context;
 use serde::de::DeserializeOwned;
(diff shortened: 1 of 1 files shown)


1st1:
Screenshot from 2021-08-11 15-43-37

1st1:
> I don't think this is viable to apply to process::run in this form:

Yeah, I agree. It should be something like `process::run_colorized(cmd, tag: String)`, where tag can be, for example `edgedb`, to make logs appear in form of

```
[edgedb] Log line
[edgedb] Log line
```

and then we apply it selectively where it makes sense.

I'd also consider colorizing Docker output as it's confusing already (see slack)

tailhook:
Yes, tags are nice. But I'm more worried of buffering issues than tags. That case of start/stop can be fully capturing, I think.

1st1:
> But I'm more worried of buffering issues than tags.

Yeah, I get that. Let's apply to Postgres/Docker/EdegDB subprocesses where necessary. I think they all only output by line.

tailhook:
Superseded by #520

PR #475, opened 2021-08-16
project init: Allow linking to exising instances in non-interactive mode

### LINKED ISSUE #471: unable to link existing EdgeDB instance and existing project in non-interactive mode
- EdgeDB Version: 1-beta3
- EdgeDB CLI Version: 1-beta3
- OS Version: Ubuntu 18.04

Steps to Reproduce:

1. Start new EdgeDB instance as docker container:
```
docker run --rm -e EDGEDB_PASSWORD=edgedb -e EDGEDB_GENERATE_SELF_SIGNED_CERT=1 -p 5656:5656 edgedb/edgedb:1-beta3 
```
2. Link running container as EdgeDB instance:
```
echo "edgedb" | edgedb -u edgedb --port=5656 instance link --trust-tls-cert --non-interactive --password-from-stdin eq_test
```
3. Try to init EdgeDB project in non-interactive mode using new instance:
```
edgedb project init --non-interactive --server-instance eq_test
Found `edgedb.toml` in `/home/nik/CodeSources/edgeql-queries`
Initializing project...
[2021-08-12T23:16:40Z WARN  edgedb::project::init] Instance "eq_test" already exists
Checking EdgeDB versions...
┌────────────────────────┬──────────────────────────────────────────────────┐
│ Project directory      │ /home/nik/CodeSources/edgeql-queries             │
│ Project config         │ /home/nik/CodeSources/edgeql-queries/edgedb.toml │
│ Schema dir (non-empty) │ /home/nik/CodeSources/edgeql-queries/dbschema    │
│ Installation method    │ Docker Container                                 │
│ Version                │ 1-beta3-451baf4                                  │
│ Instance name          │ eq_test                                          │
└────────────────────────┴──────────────────────────────────────────────────┘
Initializing EdgeDB instance...
docker: Error response from daemon: Conflict. The container name "/edgedb_eq_test" is already in use by container "753db76205d65bc3957b77c0fc8bab0f8a7cb92133d56aeb4ddda0382549c55b". You have to remove (or rename) that container to be able to reuse that name.
See 'docker run --help'.
[2021-08-12T23:16:43Z ERROR edgedb::server::create] Bootstrap error, cleaning up...
Error response from daemon: remove edgedb_eq_test: volume is in use - [753db76205d65bc3957b77c0fc8bab0f8a7cb92133d56aeb4ddda0382549c55b]
[2021-08-12T23:16:43Z ERROR edgedb::server::create] Cannot clean up storage <docker volume "edgedb_eq_test">: process "/bin/docker" "volume" "rm" "edgedb_eq_test" failed: exit status: 1
edgedb error: cannot bootstrap <docker volume "edgedb_eq_test">: error starting server "/bin/docker" "container" "run" "--detach" "--restart=always" "--name" "edgedb_eq_test" "--user=999:999" "--publish=5656:5656" "--mount" "source=edgedb_eq_test,target=/var/lib/edgedb/data" "--label=com.edgedb.metadata.user=nik" "--label=com.edgedb.metadata.version=1-beta3" "--label=com.edgedb.metadata.current-version=1-beta3-451baf4" "--label=com.edgedb.metadata.port=5656" "--label=com.edgedb.metadata.start-conf=auto" "--env" "EDGEDB_SERVER_INSTANCE_NAME" "--env" "EDGEDB_SERVER_ALLOW_INSECURE_HTTP_CLIENTS=1" "edgedb/edgedb:1-beta3" "edgedb-server" "--runstate-dir" "/var/lib/edgedb/data/run" "--data-dir" "/var/lib/edgedb/data/eq_test" "--port" "5656" "--bind-address=0.0.0.0": process "/bin/docker" "container" "run" "--detach" "--restart=always" "--name" "edgedb_eq_test" "--user=999:999" "--publish=5656:5656" "--mount" "source=edgedb_eq_test,target=/var/lib/edgedb/data" "--label=com.edgedb.metadata.user=nik" "--label=com.edgedb.metadata.version=1-beta3" "--label=com.edgedb.metadata.current-version=1-beta3-451baf4" "--label=com.edgedb.metadata.port=5656" "--label=com.edgedb.metadata.start-conf=auto" "--env" "EDGEDB_SERVER_INSTANCE_NAME" "--env" "EDGEDB_SERVER_ALLOW_INSECURE_HTTP_CLIENTS=1" "edgedb/edgedb:1-beta3" "edgedb-server" "--runstate-dir" "/var/lib/edgedb/data/run" "--data-dir" "/var/lib/edgedb/data/eq_test" "--port" "5656" "--bind-address=0.0.0.0" failed: exit status: 125
```

I don't actually start EdgeDB container manually, as this is done in GitHub Actions pipeline, where EdgeDB is started as a `service`. And then I just want to link this `service` with the `edgedb project` so that I can get rid of instance names and DSNs in my tests.

It seems to me that the error is in this code: https://github.com/edgedb/edgedb-cli/blob/e2111ea2db12863276b2cbea49db95841008904d/src/project/init.rs#L164-L169

Shouldn't it be something like this, given that `default_name` is taken from `--server-instance` value?
```rust
if options.non_interactive {
    let exists = if instances.contains(&default_name) {
        log::warn!("Instance {:?} already exists", default_name);
        true
    } else {
        false
    };
    return Ok((default_name, exists));
}
```

tailhook:
Fixes #471

Although it's a simple fix, I'm not sure this is what we want. May be we want an explicit `project link` command? I.e. for interactive mode we do as much as possible by asking questions. But for non-interactive, `project link` would be a simpler and straightforward command with less useless (ignored) arguments.

### DIFF
--- a/src/project/init.rs
+++ b/src/project/init.rs
@@ -164,7 +164,7 @@ fn ask_name(methods: &Methods, dir: &Path, options: &Init)
     if options.non_interactive {
         let exists = instances.contains(&default_name);
         if exists {
-            log::warn!("Instance {:?} already exists", default_name);
+            log::warn!("Linking project to the existing tnstance {:?}", default_name);
         }
         return Ok((default_name, exists))
     }
--- a/src/project/init.rs
+++ b/src/project/init.rs
@@ -162,10 +162,11 @@ fn ask_name(methods: &Methods, dir: &Path, options: &Init)
         name
     };
     if options.non_interactive {
-        if instances.contains(&default_name) {
+        let exists = instances.contains(&default_name);
+        if exists {
             log::warn!("Instance {:?} already exists", default_name);
         }
-        return Ok((default_name, false))
+        return Ok((default_name, exists))
     }
     let mut q = question::String::new(
         "Specify the name of EdgeDB instance to use with this project"


nsidnev:
Thanks for this PR! Just tested it and it works great (at least on my machine :) )!
But I also agree with you that the new command will be more suitable for this situation. The `edgedb` CLI already has `instance link` command, so it would be idiomatic to have a similar command for `project`.

src/project/init.rs
```diff
@@ -162,10 +162,11 @@ fn ask_name(methods: &Methods, dir: &Path, options: &Init)
         name
     };
     if options.non_interactive {
-        if instances.contains(&default_name) {
+        let exists = instances.contains(&default_name);
+        if exists {
             log::warn!("Instance {:?} already exists", default_name);
```
elprans, line 167:
```suggestion
            log::warn!("Linking project to the existing tnstance {:?}", default_name);
```

fantix (approved):
I'm +1 for the `project link` command too - if my understanding is correct, `project init` will be a superset of `project link`, where `project link` only create necessary link files from an EdgeDB project to an existing instance? That'll also solve the issue that `project init` doesn't work with remote instances for now.

tailhook:
This is a good use case where you want link and never create:
https://github.com/nsidnev/edgedb-elixir/blob/25bcc0d1b7af8090a214086470bd9544f0f97566/.github/workflows/tests.yml#L51-L61
Although, I'm not sure whether this has anything to do with non-interactive `project init` making a link too.

tailhook:
On the team meeting today we decided that we want this work in non-interactive mode only when `--link` argument is specified on `project init` command-line.

PR #480, opened 2021-08-19
Fix compile-time detection of aarch64

tailhook:


1st1 (approved):

src/server/detect.rs
```diff
@@ -14,10 +14,13 @@ use anyhow::Context;
 
 #[cfg(target_arch="x86_64")]
 pub const ARCH: &str = "x86_64";
+#[cfg(target_arch="aarm64")]
+pub const ARCH: &str = "aarm64";
 #[cfg(not(any(
     target_arch="x86_64",
+    target_arch="aarm64",
```
elprans, line 21: Is it really `aarm64`?  I think it should be either `aarch64` or `arm64`.

tailhook: Yes. This is somehow weird that I've written aarch in commit and branch name, but not in code :man_shrugging:

PR #481, opened 2021-08-20
Initial implementation of the configuration

tailhook:


src/main.rs
```diff
@@ -95,13 +97,19 @@ fn _main() -> anyhow::Result<()> {
...
+    let cfg = config::get_config();
 
     let mut builder = env_logger::Builder::from_env(
         env_logger::Env::default().default_filter_or("warn")
     );
     log_levels::init(&mut builder, &opt);
     builder.init();
 
+    let cfg = cfg.unwrap_or_else(|e| {
+        log::warn!("Config error: {:#}", e);
```
1st1, line 109: Can you rephrase it to `Error in config file '/path/to/config/cli.toml': error text here`?

1st1 (approved):
lgtm. Can we also print (in grey) something like "applied /path/to/edgedb/cli.toml configuration file" when a user starts the repl?

1st1:
Also let's rename `limit` to `implicit_limit` as @elprans suggested, but that can be done in a separate PR if that's more convenient.

src/main.rs
tailhook: Config file path is added here: https://github.com/edgedb/edgedb-cli/pull/481/files/9a3f395c2c0958204d24f064676302fe24e3a4ea#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR46
(`#[context(..)]` decorator around `read_config`)

tailhook:
> Also let's rename `limit` to `implicit_limit` as @elprans suggested, but that can be done in a separate PR if that's more convenient.

Then we should rename `\set limit` ->  `\set implicit-limit`, right?

PR #484, opened 2021-08-24
repl: Add print for used configuration

tailhook:
This also introduces a new way for color printing:
1. Styles are semantic, no colors thorough code
2. Don't need intermediate strings (`x.to_string().light_blue()`)
3. Disabling colors works even when formatting to a string
4. Themes can be used later

Also introduces new print macro `echo` which:
1. Doesn't require format pattern (patterns have little use with colors)
2. Buffers the whole line before printing (friendlier to other
   processes, faster)

This unfortunately introduces another way for color printing thorought
code, but we eventually have to convert colors to semantic names for
theme support anyway.

Other than that, this color priting is mostly in proof of concept state
as more styles should be added to make it useful.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,5 +1,5 @@
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use fn_error_context::context;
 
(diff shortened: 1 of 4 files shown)


fantix (approved):
LGTM!

Q: will there be an `eecho!`-ish macro that prints into the stderr?

tailhook:
> Q: will there be an `eecho!`-ish macro that prints into the stderr?

Yes. Although, I don't like the name. Maybe there are better alternatives to `echo`?

PR #494, opened 2021-09-04
add `project link` command

nsidnev:
In general, this is a more limited version of the  `edgedb project init` command that only asks for the name of an existing EdgeDB instance and creates project directory for it.

Also I'm not sure where to put tests for this PR. Should I create a separate test for this command, or would it be enough to modify existing tests in [tests/github-nightly/project.rs](https://github.com/edgedb/edgedb-cli/blob/fc67d94e5f12d0f37096640a1b167b82ec04b38c/tests/github-nightly/project.rs)?

Refs #471
Refs https://github.com/edgedb/setup-edgedb/pull/21#issuecomment-909422888
Supersedes (? not sure about that) #475

### DIFF
--- a/src/project/link.rs
+++ b/src/project/link.rs
@@ -1,0 +1,104 @@
+use std::{collections::BTreeSet};
+
+use anyhow::Context;
+
+use crate::print;
+use crate::project::config;
+use crate::project::options::Link;
+use crate::project::{project_dir, stash_path, write_stash_dir};
+use crate::question;
+use crate::server::control::get_instance;
+use crate::server::detect::{self, VersionQuery};
+use crate::server::is_valid_name;
+use crate::server::methods::Methods;
+
+pub fn link(options: &Link) -> anyhow::Result<()> {
+    let project_dir = project_dir(options.project_dir.as_ref().map(|x| x.as_path()))?;
+    let stash_dir = stash_path(&project_dir)?;
+
+    if stash_dir.exists() {
+        anyhow::bail!("Project is already linked");
+    }
+
+    let config_path = project_dir.join("edgedb.toml");
+    let config = config::read(&config_path)?;
+    let ver_query = match config.edgedb.server_version {
+        None => VersionQuery::Stable(None),
+        Some(ver) => ver.to_query(),
+    };
+
+    let os = detect::current_os()?;
+    let avail_methods = os.get_available_methods()?;
+    let methods = avail_methods.instantiate_all(&*os, true)?;
+
+    let name = ask_instance_name(&methods, options)?;
+
+    let instance = get_instance(&methods, &name)?;
+    let version = instance.get_version()?;
+    if !ver_query.matches(version) {
+        print::warn(format!(
+            "WARNING: existing instance has version {}, \
+                but {} is required by `edgedb.toml`",
+            version.title(),
+            ver_query
+        ));
+    }
+
+    write_stash_dir(&stash_dir, &project_dir, &name)?;
+
+    print::success("Project linked");
+    if let Some(dir) = &options.project_dir {
+        println!(
+            "To connect to {}, navigate to {} and run `edgedb`",
+            name,
+            dir.display()
+        );
+    } else {
+        println!("To connect to {}, run `edgedb`", name);
+    }
+
+    Ok(())
+}
+
+
+fn ask_instance_name(methods: &Methods, options: &Link) -> anyhow::Result<String> {
+    let instances = methods
+        .values()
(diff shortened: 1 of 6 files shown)


src/project/link.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    if options.non_interactive {
+        anyhow::bail!("Existing instance name should be specified")
+    }
+
+    let mut q =
+        question::String::new("Specify the name of EdgeDB instance to link with this project");
+    loop {
+        let target_name = q.ask()?;
+        if !is_valid_name(&target_name) {
```
tailhook, line 91: We probably don't need validation. As we are checking `contains` anyway.

tailhook (commented):
Thanks for the contribution! Code looks good. But I'm not sure we've settled on the semantics with a team:

1. I think project link could work in two ways: `project link -Iname` to link to an instance and `project link --host localhost --port 8080 --trust-tls-cert` to combine `instance link` and `project link` into a single step (as a bonus unique name is autogenerated). To make your code forward compatible to that we should make `name` an optional `-I` argument instead of positional.

2. I'm not sure interactive mode is very useful here. You can already link to a name in interactive mode using `project init`. And we have interactive `instance link` to create an instance. Although, being non-interactive by default is probably not forward compatible to adding it later, so should be considered carefully.

@elprans, @1st1, @fantix

nsidnev:
Thanks for the review. I can fix both parts if you think this will be better, but I'd like to share my opinion.

1. I don't really like the idea that `project link` command is union of `instance link` + `project link` commands. I'd expect it to perform only one action (creating files in this case). If I need to link a project with an instance that is not yet defined and this command will do it by automatically, then it seems to me that this is not very intuitive.
2. I added an interactive mode for this command to make it look similar to other existing commands, and especially to `instance link`, and also to unify it with the behavior of the `edgedb` CLI when user entered invalid data.  In most cases, `edgedb` CLI makes it possible to correct the input, doesn't it?

src/project/link.rs
nsidnev: Yeah, you are right. That part is really meaningless here.  I'll remove it

tailhook:
> Thanks for the review. I can fix both parts if you think this will be better, but I'd like to share my opinion.
> 
>     1. I don't really like the idea that `project link` command is union of `instance link` + `project link` commands. I'd expect it to perform only one action (creating files in this case). If I need to link a project with an instance that is not yet defined and this command will do it automatically, then it seems to me that this is not very intuitive.
> 

It's automatic but not implicit. I.e. you have to specify at least one connection argument to make that work. So I don't see a problem here.

>     2. I added an interactive mode for this command to make it look similar to other existing commands, and especially to `instance link`, and also to unify it with the behavior of the `edgedb` CLI when user entered invalid data.  In most cases, `edgedb` CLI makes it possible to correct the input, doesn't it?

Generally edgedb doesn't allow to correct the input, in the sense that everything taken from the command-line is applied as is. We only allow additional questions for the data that was not provided in command-line. I.e. if you run `edgedb project init --server-version=1-beta3` we may fail to install that version for some reason, but will never install any other version. This is done so that if you follow some tutorial and that tutorial assumes specific things, you don't get that things wrong because you are confused by the prompt.

So this doesn't answer if we want this command to be interactive. Do you have a use case for interactive `edgedb project link`? I think it will always be used in a script (most likely in CI). Otherwise you will just follow tutorial and run `edgedb project init` interactively.

tailhook:
So on the team meating today we decided that we want it to be named `edgedb project init --link`.

tailhook:
Let me know if you want to work on that.

nsidnev:
Yeah, I'd like to continue working on that, thanks! But I think that I'll make a new PR and this one will be closed. And before that happens, could you answer one question about new semantics of the `project init` command?

Do I understand correctly that new usage will be something like this (according to your [comment](https://github.com/edgedb/edgedb-cli/pull/475#issuecomment-914504249)):
```bash
$ edgedb project init --non-interactive --link --server-instance eq_test
```
or will it be `edgedb project init --link eq_test` and non-interactive mode will be implicitly enabled?

tailhook:
I don't think that `--non-interactive` mode should be implied. This also means that `project init --link` should only ask for a name (of an existing instance) in interactive mode, which should be a simple change. And yes `--link` is a flag and name is specified as `--server-instance=name`.

Also we've agreed that `project init --non-interactive` (without `--link`) should fail if existing instance is specified.

nsidnev:
Got it! Thanks for the comment, I'll implement this in a new PR.

PR #509, opened 2021-09-28
Add a configuration setting to disable typename injection

msullivan:
This will be useful for me doing development. If there is a way to
make the setting undocumented, it might be worth doing that, since it
seems fairly marginal for normal usage.

### DIFF
--- a/src/commands/parser.rs
+++ b/src/commands/parser.rs
@@ -156,7 +156,7 @@ pub enum Setting {
     Limit(Limit),
     /// Set output format.
     OutputFormat(OutputFormat),
-    /// Whether to inject typenames in default output mode
+    /// Display typenames in default output mode
     DisplayTypenames(SettingBool),
     /// Stop escaping newlines in quoted strings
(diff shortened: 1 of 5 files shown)


src/commands/parser.rs
```diff
@@ -156,6 +156,8 @@ pub enum Setting {
     Limit(Limit),
     /// Set output format.
     OutputFormat(OutputFormat),
+    /// Whether to inject typenames in default output mode
```
elprans, line 159:
```suggestion
    /// Resolve object type names in default output mode
```

elprans:
> If there is a way to make the setting undocumented

We could add a "devmode" feature to the CLI crate and enable it when building the CLI from `setup.py`.

elprans (approved):

src/commands/parser.rs
msullivan: How about "Display object type names"?

tailhook:
Technically we have `dev_mode` feature and we build with it from `setup.py`.

PR #522, opened 2021-10-12
project init: add --server-start-conf

tailhook:
Also changed error status on `instance create --start-conf=manual` to
return 0 (success) when service cannot be created.

Simplifies a github action a bit: https://github.com/edgedb/setup-edgedb/pull/22


tailhook:
> `edgedb project init --server-start-conf manual` is failing at migration on at least macOS:

Fixed!

PR #526, opened 2021-10-15
Nightly test: issue fix and speed up

fantix:
This PR fixes most of the nightly tests, and reuses the built CLI and test executables accross sharded tests.

Remaining issues:

- [x] After `edgedb instance revert`, the SSL certificate is somehow incorrect.
- [ ] Some of the `project` test is unstable.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ on:
     - cron: "0 */6 * * *"
 
 env:
-  RUST_TEST_THREADS: 2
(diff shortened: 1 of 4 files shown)


fantix:
> for new updates we have to keep certificate instead of regenerating on upgrade, right?

Yes I think so!
 
> maybe we want start command to poll for status until process have been started?

Yeah that would be nice - I think I met a similar situation some time ago and decided to depend on the REPL/client reconnecting.

PR #533, opened 2021-10-25
Update env vars accroding to the new spec

tailhook:
See more: https://github.com/edgedb/edgedb/discussions/3049

### DIFF
--- a/src/credentials.rs
+++ b/src/credentials.rs
@@ -11,6 +11,7 @@ use edgedb_client::Builder;
 use edgedb_client::credentials::Credentials;
 
 use crate::platform::config_dir;
+use crate::question;
 use crate::server::reset_password::write_credentials;
 use crate::server::is_valid_name;
 
@@ -18,7 +19,7 @@ use crate::server::is_valid_name;
 
 pub fn get_connector(name: &str) -> anyhow::Result<Builder> {
     let mut builder = Builder::uninitialized();
-    task::block_on(builder.read_credentials(path(name)?))?;
+    task::block_on(builder.read_instance(name))?;
     Ok(builder)
 }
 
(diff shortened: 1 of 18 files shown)


src/server/debian_like.rs
```diff
@@ -175,7 +175,6 @@ impl Debian {
             .arg("-y")
             // TODO(tailhook) version
             .arg(format!("edgedb-server-{}", pkg.slot))
-            .env("_EDGEDB_INSTALL_SKIP_BOOTSTRAP", "1")
```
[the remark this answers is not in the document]
tailhook: Yes

tailhook:
Also take a look at: https://github.com/edgedb/edgedb-cli/issues/537

src/server/docker.rs
```diff
@@ -791,6 +795,11 @@ impl<'os, O: CurrentOs + ?Sized> DockerMethod<'os, O> {
                 DROP ROLE {};
             "###, role=tmp_role)).await?;
 
+            // TODO: skip if --non-interactive
+            credentials::ask_update_credentials(
```
tailhook, line 799: I don't think we need to ask here. We already rewrite credentials to add a certificate, just fix the credentials anyways. I think it's fine (to not ask) even if we keeping certificate the same, as the expectation is the same: your credentials might be updated on server upgrade.

src/credentials.rs
```diff
@@ -61,3 +62,21 @@ pub fn add_certificate(instance_name: &str, certificate: &str)
...
+pub async fn ask_update_credentials(
+    instance_name: &str, builder: &Builder
+) -> anyhow::Result<()> {
+    if builder.should_update_credential_file() {
+        let creds_path = path(instance_name)?;
+        if question::Confirm::new(format!(
+            "The format of the instance credential file at {} is outdated, \
+             update now?",
+            creds_path.display(),
+        )).ask()? {
```
tailhook, line 75: Using sync in async code, is always weird.

The problem here is that in `process.background_for` async block we handle signals differently. This is where async code signals are explicitly awaited and cancel other futures. So in this case `TERM` will not work during asking (can't cancel currently blocked future) and I'm not sure about `SIGINT` (it may fail input reader, but maybe not for all OSs).

And this is very bad since it can happen to wait for input in non-interactive environment, and then supervisor can't kill it normally by `TERM` and when it `KILL`s it, there is a hanging `process`.

src/server/unix.rs
```diff
@@ -482,9 +482,16 @@ fn _reinit_and_restore(
...
 
     cmd.background_for(async {
+        let builder = inst.get_connector(true)?;
         upgrade::restore_instance(
-            inst, &dump_path, inst.get_connector(true)?
+            inst, &dump_path, builder.clone()
         ).await?;
+
+        // TODO: skip if --non-interactive
+        credentials::ask_update_credentials(
```
tailhook, line 491: I think this also doesn't require asking user, as described earlier. Also this doesn't need to be done in `background_for` block.

src/project/init.rs
```diff
@@ -955,12 +955,16 @@ async fn migrate(inst: &InstInfo<'_>, ask_for_running: bool)
         };
     };
 
+    let builder = inst.get_builder().await?;
+    if ask_for_running {
```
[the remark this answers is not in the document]
fantix: It was a common place for `project init` - but nvm, I've created a method and calling from proper locations.

src/credentials.rs
fantix: Yeah, I changed it to a blocking function.

tailhook (commented):
Looks good to me

PR #540, opened 2021-11-04
Update for #533

fantix:
Depends on https://github.com/edgedb/edgedb-rust/pull/107

I don't think we should ask the user to update the creds file when executing commands in this file: https://github.com/edgedb/edgedb-cli/blob/master/src/commands/execute.rs, so I didn't add the update logic to them.

### DIFF
--- a/src/server/reset_password.rs
+++ b/src/server/reset_password.rs
@@ -10,7 +10,7 @@ use edgedb_client::credentials::Credentials;
 use fn_error_context::context;
 use rand::{Rng, SeedableRng};
 
-use crate::credentials;
+use crate::credentials as creds_crate;
 use crate::print;
 use crate::server::options::ResetPassword;
 use crate::server::detect;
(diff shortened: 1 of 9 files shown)


src/server/reset_password.rs
```diff
@@ -10,7 +10,7 @@ use edgedb_client::credentials::Credentials;
 use fn_error_context::context;
 use rand::{Rng, SeedableRng};
 
-use crate::credentials;
+use crate::credentials as creds_crate;
```
tailhook, line 13: Kinda weird name I would read it as "credentials crate", so it should be a separate crate rather than a module within the current one.

Also we `use crate::credentials;` everywhere I think and I don't see a name conflict here.

tailhook (approved):
Other than nitpicking on module alias looks good.

src/server/reset_password.rs
fantix: Yeah you're right - fixed!

PR #543, opened 2021-11-05
Idle timeout

tailhook:
It's in draft because requires to get edgedb-rust back to master when edgedb/edgedb-rust#108 is applied.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ dependencies = [
 [[package]]
 name = "edgedb-client"
 version = "0.2.0"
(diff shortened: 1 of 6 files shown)


src/interactive.rs
```diff
@@ -558,6 +572,9 @@ async fn _interactive_main(options: &Options, state: &mut repl::State)
         state.ensure_connection()
             .race(ctrlc.wait_result())
             .await?;
+        if let Some(conn) = state.connection.as_mut() {
+            update_idle_transaction_timeout(conn).await?;
+        }
```
tailhook, line 577: We need to adjust it after every reconnect. It should be somewhere in `state.ensure_connection` I think.

fantix (approved):

PR #547, opened 2021-11-11
Add CLI config idle-transaction-timeout

fantix:


src/repl.rs
```diff
@@ -105,10 +106,15 @@ impl State {
...
-                    "CONFIGURE SESSION SET \
-                 session_idle_transaction_timeout := <std::duration>'5m';"
-                ).await.context(
+                let secs = self.idle_transaction_timeout.unwrap_or(0);
+                log::info!(
+                    "Setting session_idle_transaction_timeout to {}s", secs
+                );
+                conn.execute(&format!(
+                    "CONFIGURE SESSION SET session_idle_transaction_timeout \
+                     := <std::duration>'{}s'",
```
1st1, line 115: You don't need 's' suffix; numbers without a unit are assumed to be seconds for std::duration.

```suggestion
                     := <std::duration>'{}'",
```

1st1 (approved):

src/commands/backslash.rs
```diff
@@ -588,6 +595,15 @@ pub async fn execute(cmd: &BackslashCmd, prompt: &mut repl::State)
                         prompt.print.max_items = Some(limit);
                     }
                 }
+                IdleTransactionTimeout(t) => {
+                    let secs = t.value.expect("only set here");
+                    prompt.idle_transaction_timeout = if secs == 0 {
```
tailhook, line 600: Not sure why we're doing that `0 => None` dance if `0` on the server also means no timeout. It's rather confusing that `idle_transactio_timeout = None` could mean "default" or could mean "no timeout".

tailhook (commented):
I would prefer adding a duration parser. This isn't something too hard to implement. And we need one for input variables anyway. And we already have one in `--wait-until-available` that has to be replaced with one more consistent with the server.

fantix:
> we don't have postgres-compatible formatting for duration yet.

I think we do?

> So we just dump microseconds. 

This is because EdgeQL `CONFIGURE ... SET ...` doesn't support parameters:

```
>>> import edgedb
>>> conn = edgedb.connect("_localdev")
>>> conn.query("CONFIGURE INSTANCE SET session_idle_transaction_timeout := <std::duration>$timeout", timeout='0')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python@3.10/3.10.0_2/envs/edgedb10/lib/python3.10/site-packages/edgedb/blocking_con.py", line 350, in query
    return self._execute(
  File "edgedb/blocking_con.py", line 329, in _execute
    raise e
  File "edgedb/blocking_con.py", line 316, in _execute
    return self._get_protocol().sync_execute_anonymous(
  File "edgedb/protocol/blocking_proto.pyx", line 105, in edgedb.protocol.blocking_proto.BlockingIOProtocol.sync_execute_anonymous
  File "edgedb/protocol/blocking_proto.pyx", line 91, in edgedb.protocol.blocking_proto.BlockingIOProtocol._iter_coroutine
  File "edgedb/protocol/protocol.pyx", line 628, in execute_anonymous
  File "edgedb/protocol/protocol.pyx", line 292, in _parse
edgedb.errors.QueryError: non-constant expression in CONFIGURE INSTANCE SET
```

> And we have limit number of microseconds for that to work, because postgres's parser doesn't accept more that u32 micros in the parser. Right?

Yes, Postgres limits this to non-negative i32 `[0, 2147483648)`.
 
> let's make a comment of why there is such a limit

Sure thing!

>  and todo on query formatter. Also let's make sure we have an open issue on duration formatting

I'm not so sure about this one - looks to me that `edgedb-rust` or the server works fine with duration formatting.

> (Also current `duration.to_string()` should work even if its formatting is uglier so what's wrong with that?)

`duration.to_string()` also works, I just want to have less convertings - but not a big deal anyways.

PR #549, opened 2021-11-12
Fix truncated output

fantix:
Looks like using `join` would simply fix the truncated output problem - I think it's safe to do so, please kindly advise if not.


tailhook (approved):
Yes, this makes total sense.

PR #552, opened 2021-11-17
project init: Fixing initialization with start-conf=manual

tailhook:
As reported in https://github.com/edgedb/setup-edgedb/pull/22#issuecomment-970451629


tailhook:
Superseded by #553

PR #556, opened 2021-11-19
Use C locale for all subprocesses

### LINKED ISSUE #523: Non-English locale breaks certain behavior
- EdgeDB Version: 1-rc1
- EdgeDB CLI Version: 1-rc2-dev
- OS Version: Debian Stretch

Steps to Reproduce:

1. Install `locales` and add a non-English locale to `/etc/locale.gen` and reconfigure `locales`.
2. Run some CLI commands that are depending on the output of a system command with the locale, like `LANG=zh_CN.UTF-8 edgedb server uninstall --all`.

tailhook:
Fixes #523

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -8,6 +8,8 @@ use std::path::{Path, PathBuf};
 use std::process::exit;
 use std::time::{SystemTime, UNIX_EPOCH};
 
+use once_cell::sync::Lazy;
+
 use anyhow::Context;
 use async_process::{Command, Stdio, ExitStatus, Output};
 use async_std::io::prelude::{BufReadExt};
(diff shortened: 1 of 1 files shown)


src/process.rs
```diff
@@ -77,10 +77,13 @@ impl Native {
         cmd: impl AsRef<Path>)
         -> Native
     {
+        let mut command = Command::new(cmd.as_ref());
+        command.env("LANG", "C");
+        command.env("LC_ALL", "C");
```
elprans, line 82: Can we detect if C.utf8 is present and use that?

elprans (approved):

fantix (approved):

src/process.rs
tailhook: Well, yes. Reading logs is a big ugly with locale `C`. But what to do if users current locale/terminal is not utf8? Should we check current locale first? (also there is no other `C` except `C` and `C.utf8` so we can't transpose any encoding on `C`)

PR #567, opened 2021-11-23
Format errors with hint/context in special cases

### LINKED ISSUE #559: Showing error information
Compare the following errors:
```
edgedb> select {'1', 2};
error: InvalidTypeError: set constructor has arguments of incompatible types 'std::str' and 'std::int64'
  ┌─ query:1:8
  │
1 │ select {'1', 2};
  │        ^^^^^^^^ Consider using an explicit type cast or a conversion function.

edgedb> select <datetime>'1234-11-11';
InvalidValueError: invalid input syntax for type std::datetime: '1234-11-11'
```
The rendering is a bit inconsistent, one has `error:` prefix the other does not. There's also a difference in colors. Very importantly, the second error does not render the "hint" at all, even though the field should contain the following message `Please use ISO8601 format. Example: 2010-12-27T23:59:59-07:00. Alternatively "to_datetime" function provides custom formatting options.`.

As far as I can tell, currently we only render the hint if and only if the source context is present (I've tried generating the first error, but omitting the context). Instead we should render the hint when the hint is present, regardless of whether there is source context as well.

fantix:
This PR fixes the missing hint in error output for query `select <datetime>'1234-11-11';` in interactive mode for all output formats.

I have also checked all `print::error` calls to make sure we're not passing server errors to that function. I have also tested both `interactive.rs` and `non_interactive.rs` for all output formats with the two queries mentioned in #559, with or without this server patch:

```patch
diff --git a/edb/edgeql/compiler/func.py b/edb/edgeql/compiler/func.py
index a789505e6..2a3cf73ba 100644
--- a/edb/edgeql/compiler/func.py
+++ b/edb/edgeql/compiler/func.py
@@ -498,8 +498,7 @@ def compile_operator(
                 )
             raise errors.InvalidTypeError(
                 msg,
-                hint=hint,
-                context=qlexpr.context)
+                hint=hint)
         elif len(matched) > 1:
             if in_abstract_constraint:
                 matched_call = matched[0]
```

Hint is always printed now.

Fixes #559

(@vpetrovykh however I cannot reproduce the hint-missing issue with this server patch for query `select {'1', 2}; ` with CLI 1.0-rc2, am I missing anything here?)

### DIFF
--- a/src/error_display.rs
+++ b/src/error_display.rs
@@ -7,10 +7,10 @@ use codespan_reporting::term::{emit};
 use termcolor::{StandardStream, ColorChoice};
 
 use edgedb_client::errors::{Error, InternalServerError};
-use edgedb_protocol::error_response::FIELD_POSITION_START;
-use edgedb_protocol::error_response::FIELD_POSITION_END;
-use edgedb_protocol::error_response::{FIELD_HINT, FIELD_DETAILS};
-use edgedb_protocol::error_response::FIELD_SERVER_TRACEBACK;
+use edgedb_client::errors::FIELD_POSITION_START;
+use edgedb_client::errors::FIELD_POSITION_END;
+use edgedb_client::errors::{FIELD_HINT, FIELD_DETAILS};
+use edgedb_client::errors::FIELD_SERVER_TRACEBACK;
 
 use crate::print;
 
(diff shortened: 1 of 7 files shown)


src/interactive.rs
```diff
@@ -456,7 +456,9 @@ async fn execute_query(options: &Options, mut state: &mut repl::State,
                             source: ref error,
                             ..
                         } => {
-                            eprintln!("{:#}", error);
+                            print_query_error(
+                                error, statement, state.verbose_errors
+                            )?;
                         }
                         _ => eprintln!("{:#?}", e),
```
tailhook, line 463: As we're here, I'm not sure why this is debug output, not `{:#}` like everything else.

tailhook (approved):
This looks good, and propagates `verbose_errors`.

But in addition to this, I think we should move `edgedb_client::error_response::display_error` to `edgedb_errors` crate and make `display_error(e, false)` same as `fmt::Display for Error`. This will ensure that hint/details is printed even if we forgot to match error and more importantly show them by default for Rust bindings.

src/interactive.rs
fantix: Looks like `PrintError` (e) didn't implement `fmt::Display`.

src/error_display.rs
```diff
@@ -37,8 +37,23 @@ pub fn print_query_error(err: &Error, query: &str, verbose: bool)
...
+        print::error(context_error.join(": "));
+    }
     let diag = Diagnostic::error()
-        .with_message(&format!("{:#}", err))
+        .with_message(&format!(
+            "{:#}{:#}",
+            err.kind_name(),
+            err
+                .initial_message()
+                .map(|s| format!(": {:#}", s))
```
tailhook, line 54: nit: since these are strings, you don't need alternate `{:#}` formatting anywhere. Just `{}` is fine. It's not a problem here, but confuse the reader thinking that something special is going on here.

tailhook (approved):

PR #595, opened 2021-12-02
edgedb-init: Fix Rosetta fallback on macOS

elprans:


edgedb-init.sh
```diff
@@ -173,7 +173,7 @@ get_architecture() {
     esac
 
     if [ "$_cputype" = "arm64" ]; then
-        if [ "$_ostype" = "macos" ]; then
+        if [ "$_ostype" = "apple-darwin" ]; then
             # Rely on Rosetta for now
```
1st1, line 177: Can we print here that rosetta will be used? Maybe even print that "softwareupdate --install-rosetta --agree-to-license" can be used to manually install rosetta 2?

elprans: What does macOS print when you try to run an x86 binary?  It seems to me that this should be automatic on macOS side of things, no?

tailhook (approved):

PR #599, opened 2021-12-08
migration create: Fixes timeout handling

### LINKED ISSUE #598: ClientConnectionEosError when creating migrations
- EdgeDB Version: RC3
- OS Version: macOS 11.6

Steps to Reproduce:
I can't for the life of me find reproducible steps but I keep getting this error very frequently when trying to create migrations. Some migrations trigger it and some don't and I can't really find the pattern of what triggers it:

```
Error applying statement: ClientConnectionEosError
Rolling back last operation...
edgedb error: ClientInconsistentError: Connection is inconsistent state. Please reconnect.
```

tailhook:
Should fix #598

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -468,9 +468,9 @@ dependencies = [
 
 [[package]]
 name = "blocking"
-version = "1.0.2"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c5e170dbede1f740736619b776d7251cb1b9095c435c34d8ca9f57fcd2f335e9"
+checksum = "046e47d4b2d391b1f6f8b407b1deb8dee56c1852ccd868becf2710f601b5f427"
 dependencies = [
  "async-channel",
  "async-task",
@@ -990,6 +990,7 @@ dependencies = [
  "bigdecimal",
  "blake2b_simd",
  "blake3",
+ "blocking",
  "bytes",
  "chrono",
  "clap",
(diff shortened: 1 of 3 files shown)


vpetrovykh:
I run into this kind of issue a lot when I'm using REPL for testing:
```
db> start transaction;
OK: START TRANSACTION
db[tx]> create type Bar;
edgedb error: ClientInconsistentError: Connection is inconsistent state. Please reconnect.
Reconnecting...
```
All it took was a brief pause between the `start transaction` and the next command. So any time I need to type/edit some query that's not completely trivial this comes up. Basically, unless I'm copy-pasting my queries from somewhere there's a good chance I run into this.

I think that the transaction timeout in REPL should be significantly higher than the default setting, maybe even completely ignoring the base config. The reason being that the times that I actually _need_ a transaction in REPL are the times when I might need extra time to double-check what I'm doing, look up something in the middle of it, compare some output to a bug report, etc. All of these things can reasonably take minutes and having my carefully set up transaction scrapped because I needed to validate some results or plainly got up to grab a coffee is not a great experience.

I could easily see REPL transactions be reasonable with hours' long timeouts.

tailhook:
@vpetrovykh, we have increased timeout to 5 min. I agree that it may be not enough, on the other hand, REPL is also for connecting to production databases and for that you could block your whole app's progress or prevent vacuuming by holding some transaction open. So I'm not sure what default would be the best.t
You can also set your own timeout with `\set idle-transaction-timeout 20min`. 

Feel free to open a separate issue to discuss the default.

PR #602, opened 2021-12-09
Better error reporting on connection timeouts

tailhook:


src/repl.rs
```diff
@@ -140,7 +140,13 @@ impl State {
         Ok(())
     }
     pub async fn soft_reconnect(&mut self) -> anyhow::Result<()> {
-        if !self.in_transaction() {
+        if self.in_transaction() {
+            let is_closed = self.connection.as_ref()
+                .map(|c| !c.is_consistent()).unwrap_or(false);
+            if is_closed {
+                anyhow::bail!("connection closed by server");
```
elprans, line 147: Perhaps the server should send a message before closing to indicate the reason, so we can make an even better message: "connection closed by server due to inactivity (session_idle_transaction_timeout=<foo>)"?

tailhook: The problem is that at this point `is_consistent` could mean anything: from timeout to network error to bugs in code. So yes, we can do that, but that's a different PR (and to the `edgedb-rust` in the first place).

PR #607, opened 2021-12-14
Fix `--nightly` option in `edgedb-init.sh`

elprans:
The nightly CLI builds in `dist/` no longer have the `_nightly` suffix.

While at it, sync helpers with the latest `rustup-init.sh` to fix the
bogus "Not forcing TLS v1.2, this is potentially less secure" warning.


tailhook (approved):

tailhook:
It's a bit unfortunate that secure HTTPS from the command-line is still so complex :(

PR #610, opened 2021-12-29
project init: add `--print-project-info`

tailhook:
Also includes:
1. Project commands don't print to stdout anymore
2. `echo` macro support semicolon char (no-space output)

Note: argument is hidden for now, as this is mostly needed internally for windows/WSL support.


tailhook:
This turned out not needed for now.

PR #645, opened 2022-01-27
add `edgedb project instance start` command

### LINKED ISSUE #644: add `edgedb project instance start` for easier use on WSL
At the moment, I mostly use WSL to work on my projects. The CLI support for WSL has made working with EdgeDB very convenient, thank you!

The only thing that is a bit annoying is the need to start the EdgeDB instance linked to the project with the `edgedb instance start` command. Currently, to start instance after rebooting the computer you need to remember the exact name of the instance. It seems to me that it would be very convenient if you could start a linked EdgeDB instance with a simple `edgedb project instance start` (or something similar) command, which would itself determine which instance to use.

What do you think about that?

nsidnev:
fixes #644

I added a subcommand only to start the instance for now. Are there any other subcommands needed here from the original `instance` command? It seems to me that some of them may be useful for users. For example the followings:
* `stop`
* `restart`
* `reset-password`
* `logs`
* `status`


1st1 (changes requested):
Nik, thanks so much for working on this. Unfortunately we'll have to wait with this a bit.

-----

Elvis and I thought about this and decided to wait until we have a clearer picture about how our cloud will be integrated into the CLI.

For now it's easy to learn the current project's instance name:

```
edgedb project info --instance-name
```

Instance names don't change and usually match the project name, so for local development needs working with instances should be straightforward.

A generic alias can be created for say starting a project instance, e.g.

```
edgedb instance start $(edgedb project info --instance-name)
```

All of these aren't the most optimal solutions, and having a proper command to control project's instance is still preferable, but this won't happen until 1.0 and at least our cloud RFC.

cc @elprans

nsidnev:
> Unfortunately we'll have to wait with this a bit.

Got it, adding integration for the CLI with the cloud version of EdgeDB sounds amazing! Of course, this PR can be postponed until an appropriate time.

But, if possible, I'd like to continue impementing project's instance control feature in the CLI after adding the necessary RFC.  BTW, should I close this PR for now or can I keep it open?

> edgedb instance start $(edgedb project info --instance-name)

Yep, I use the exact same alias right now :)

nsidnev:
I've added other commands for instance except for `revert`. I'm not sure if it should be placed under `edgedb project instance`  since it seems like it should also change `edgedb.toml` and it would probably fit better with `edgedb project upgrade`.

colinhacks:
Hey Nik, really excellent work here. This is terrible timing, but we hit upon another approach this morning that we think is cleaner and provides a more consistent CLI structure: https://github.com/edgedb/edgedb-cli/issues/739. Let us know if you have any thoughts (and sorry we didn't land on this sooner) 😑

tailhook:
While we did not update our instance commands to use project config, we've settled on the approach different than this PR in #739, and made a first step (deprecating positional instance name argument). We'll do project support in the next release (so that people take some time to adopt the new `-I` option).

PR #648, opened 2022-02-03
Disable session_idle_transaction_timeout in `migrate`

elprans:
When migrations are applied, the wrapping transaction is expected to
be idle for extended periods (due to migration script parsing,
preparation of data for migration etc).  Rather than set some arbitrary
interval, disable the timeout altogether.

Per complaint by Regan Laitila in edgedb/edgedb#3395

### DIFF
--- a/src/migrations/migrate.rs
+++ b/src/migrations/migrate.rs
@@ -4,6 +4,8 @@ use async_std::path::Path;
 use async_std::stream::StreamExt;
 use colorful::Colorful;
 use edgedb_client::client::Connection;
+use edgedb_client::model::Duration;
+use edgeql_parser::helpers::quote_string;
 use linked_hash_map::LinkedHashMap;
 
 use crate::commands::Options;
@@ -147,6 +149,13 @@ pub async fn migrate(cli: &mut Connection, _options: &Options,
         }
         return Ok(());
     }
+
+    // Disable idle transaction timeout while migrations get applied.
+    let old_timeout = cli.query_row::<Duration, _>(
+        "SELECT cfg::Config.session_idle_transaction_timeout", &()).await?;
+    cli.execute("CONFIGURE SESSION SET session_idle_transaction_timeout \
+                     := <std::duration>'0'").await?;
+
     // TODO(tailhook) use special transaction facility
     cli.execute("START TRANSACTION").await?;
     for (_, migration) in migrations {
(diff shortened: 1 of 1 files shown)


src/migrations/migrate.rs
```diff
@@ -147,6 +149,13 @@ pub async fn migrate(cli: &mut Connection, _options: &Options,
         }
         return Ok(());
     }
+
+    // Disable idle transaction timeout while migrations get applied.
+    let old_timeout = cli.query_row::<Duration, _>(
+        "SELECT cfg::Config.session_idle_transaction_timeout", &()).await?;
+    cli.execute("CONFIGURE SESSION SET session_idle_transaction_timeout \
```
1st1, line 156: Just set it inside the transaction? Then you don't need to save the old value.

1st1: This is what I do in our server-side dump implementation https://github.com/edgedb/edgedb/blob/3402eb6f2b4877c1273b9bdf7e98b99442b979d5/edb/server/protocol/binary.pyx#L2396-L2409

elprans: This only works if you `ROLLBACK` at the end.  We don't have `CONFIGURE TRANSACTION` yet, unfortunately.

1st1: ah, true

1st1 (approved):

src/migrations/migrate.rs
```diff
@@ -176,6 +185,13 @@ pub async fn migrate(cli: &mut Connection, _options: &Options,
         }
     }
     cli.execute("COMMIT").await?;
+
+    cli.execute(format!(
```
tailhook, line 189: This is not a correct way to undo timeout, since it will not restore the timeout when commit fails but connection is still alive.

elprans: `migration create` has similar code.  Can you suggest a better way?

elprans: In what scenario would connection be still alive here?  I made `COMMIT` fail and the whole thing crashed.

tailhook: Not exactly. It doesn't jump out of here (no question mark):
https://github.com/edgedb/edgedb-cli/blob/fd3307282eb1fafdb65fed954d0694665d8c3965/src/migrations/create.rs#L553

And then carefully ensures that all statements have been executed regardless of the results (and errors if any are propagated only after that):
https://github.com/edgedb/edgedb-cli/blob/fd3307282eb1fafdb65fed954d0694665d8c3965/src/migrations/create.rs#L557-L563

Although, these, have probably slipped into the code by a mistake:
https://github.com/edgedb/edgedb-cli/blob/fd3307282eb1fafdb65fed954d0694665d8c3965/src/migrations/create.rs#L530-L533

So yes, that has to be fixed too.

Anyway, I can take care of all of that.

tailhook:
Superseded by #653

PR #696, opened 2022-04-01
Add `edgedb ui` command

fantix:
* If run within a project it should use the project instance; if launched differently, it should require `-I`.
* When an instance is determined, it should open your local browser pointing to `http://$INSTANCE_HOST:$INSTANCE_PORT/admin`.
* Known issue: it's currently not checking if the target server is valid or not. Like `edgedb ui --host 127.0.0.1 --port 8080` will also open `http://127.0.0.1:8080/admin`.
* This command is not hidden in the `--help`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1106,6 +1106,7 @@ dependencies = [
  "nix 0.23.1",
  "num-bigint",
  "once_cell",
+ "open",
  "openssl",
  "os-release",
  "pem 1.0.2",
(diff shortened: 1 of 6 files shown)


1st1:
> This command is not hidden in the --help.

Can we hide it?

fantix:
Now hidden 🕵️

elprans:
Why hide? We are pushing this to the nightly CLI, aren't we?

1st1:
Some people install nightlies and use them with v1. Maybe we also come up with a better name. IOW I don't see any reason why it must be in `--help` *now*.

1st1 (approved):

PR #712, opened 2022-04-17
Add `edgedb instance credentials` command

fantix:
`credentials` subcommand is NOT hidden in the `--help`.

```
edgedb-instance-credentials
Echo credentials to connect to the instance

USAGE:
    edgedb instance credentials [OPTIONS]

OPTIONS:
        --json            Output in JSON format (password is included in cleartext)
        --insecure-dsn    Output a DSN with password in cleartext
    -h, --help            Print help information

CONNECTION OPTIONS (`edgedb --help-connect` to see the full list):
    -I, --instance <instance>
            Local instance name created with edgedb instance create to connect to
            (overrides host and port)

        --dsn <dsn>
            DSN for EdgeDB to connect to (overrides all other options except password)

        --credentials-file <credentials_file>
            Path to JSON file to read credentials from
```

### DIFF
--- a/src/portable/credentials.rs
+++ b/src/portable/credentials.rs
@@ -10,8 +10,8 @@ pub fn show_credentials(options: &Options, c: &ShowCredentials) -> anyhow::Resul
     let connector = options.create_connector()?;
     let builder = connector.get()?;
     let creds = builder.as_credentials()?;
-    let result = if c.json {
-        serde_json::to_string_pretty(&creds)?
+    if let Some(result) = if c.json {
+        Some(serde_json::to_string_pretty(&creds)?)
     } else if c.insecure_dsn {
         let mut url = Url::parse(&format!(
             "edgedb://{}@{}:{}",
@@ -35,23 +35,19 @@ pub fn show_credentials(options: &Options, c: &ShowCredentials) -> anyhow::Resul
             }
             _ => {}
         }
-        url.to_string()
+        Some(url.to_string())
     } else {
-        format!(
-            "Host: {}\n\
-            Port: {}\n\
-            User: {}\n\
-            Password: {}\n\
-            Database: {}\n\
-            TLS Security: {:?}",
-            creds.host.unwrap_or("localhost".into()),
-            creds.port,
-            creds.user,
-            creds.password.map(|_| "<hidden>").unwrap_or("<none>"),
-            creds.database.unwrap_or("<default>".into()),
-            creds.tls_security,
-        )
-    };
-    stdout().lock().write_all((result + "\n").as_bytes()).expect("stdout write succeeds");
+        crate::table::settings(&[
+            ("Host", creds.host.as_deref().unwrap_or("localhost")),
+            ("Port", creds.port.to_string().as_str()),
+            ("User", creds.user.as_str()),
+            ("Password", creds.password.map(|_| "<hidden>").unwrap_or("<none>")),
+            ("Database", creds.database.as_deref().unwrap_or("<default>")),
+            ("TLS Security", format!("{:?}", creds.tls_security).as_str()),
+        ]);
+        None
+    } {
+        stdout().lock().write_all((result + "\n").as_bytes()).expect("stdout write succeeds");
+    }
     Ok(())
 }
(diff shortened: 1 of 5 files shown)


src/portable/options.rs
```diff
@@ -431,6 +434,31 @@ pub struct Info {
...
+    pub _type: Option<ShowCredentialsType>
+}
+
+impl FromStr for ShowCredentialsType {
+    type Err = anyhow::Error;
+    fn from_str(v: &str) -> anyhow::Result<ShowCredentialsType> {
+        use ShowCredentialsType::*;
+        match v {
+            "json" => Ok(JSON),
+            "insecure_dsn" => Ok(InsecureDSN),
```
tailhook, line 457:
```suggestion
            "insecure-dsn" => Ok(InsecureDSN),
```
I think we use kebab-case everywhere else

src/portable/options.rs
```diff
@@ -431,6 +434,31 @@ pub struct Info {
...
+#[derive(Debug, Clone, Copy)]
+pub enum ShowCredentialsType {
+    JSON,
+    InsecureDSN,
+}
+
+#[derive(EdbClap, Clone, Debug)]
+pub struct ShowCredentials {
+    #[clap(long, possible_values=&[
+    "json", "insecure_dsn",
```
tailhook, line 446:
```suggestion
    "json", "insecure-dsn",
```

tailhook (commented):
I find it weird that we don't expose `tls_ca` here. Why? You can't connect without it and with normal tls_security. Is it so that JSON looks prettier, but unusable, or am I missing something?

fantix:
> I find it weird that we don't expose `tls_ca` here. Why? You can't connect without it and with normal tls_security. Is it so that JSON looks prettier, but unusable, or am I missing something?

Per Zoom discussion, recap here: we should always expose `tls_ca` while there won't be a `tls_ca` to expose if we're sure to trust the signing cert and link the instnace correctly.

src/portable/credentials.rs
```diff
@@ -0,0 +1,44 @@
...
+                }
+                TlsSecurity::Insecure => {
+                    url.set_query(Some(&format!("tls_security=insecure")));
+                }
+                TlsSecurity::NoHostVerification => {
+                    url.set_query(Some(&format!("tls_security=no_host_verification")));
+                }
+                _ => {}
+            }
+            print::echo!(url);
```
tailhook, line 37: `echo` prints to stderr, and should be used for messages. This is data. Use the regular `println!` (or `stdout().lock().write_all()`) instead.

Actually I suggest something like this:
```
stdout().lock().write_all((url.to_string() + "\n").as_bytes()).expect("stdout write succeeds");
```
This is better than `println` as it will print the line with `\n` atomically (if possible). So easier to read in multi-process environments where other processes can print concurrently (i.e. tests).

src/portable/credentials.rs
```diff
@@ -0,0 +1,44 @@
...
+        }
+        _ => {
+            print::echo!(serde_json::to_string_pretty(&creds)?);
```
tailhook, line 40: Same here.

src/portable/options.rs
```diff
@@ -431,6 +434,31 @@ pub struct Info {
...
+    JSON,
+    InsecureDSN,
+}
+
+#[derive(EdbClap, Clone, Debug)]
+pub struct ShowCredentials {
+    #[clap(long, possible_values=&[
+    "json", "insecure-dsn",
+    ])]
+    pub _type: Option<ShowCredentialsType>
```
tailhook, line 448: We have `--json` everywhere. So I would either use `--json/--insecure-dsn` or use `--format=json/--format=insecure-dsn` and add a `--json` shortcut for the former.

Also we don't use `--json` by default anywhere (Do we?). Maybe we want some human-readable text/table with hidden password by default?

It's also interesting why we have `insecure-dns` but plain `json`, while `json` also has the same password.

tailhook (commented):
Sorry for the long delay. This is mostly fine. Except fixing prints. And lets bikeshed on option names just a little bit.

PR #721, opened 2022-05-19
Update instance routing to use org/instance

fmoor:
depends on https://github.com/edgedb/nebula/pull/14

### DIFF
--- a/src/cloud/ops.rs
+++ b/src/cloud/ops.rs
@@ -1,20 +1,17 @@
 use std::collections::HashMap;
 use std::fs;
 use std::io;
-use std::path::PathBuf;
 use std::time::{Duration, Instant};
 
 use async_std::task;
-use colorful::Colorful;
 use edgedb_client::credentials::Credentials;
 use edgedb_client::Builder;
 
-use crate::cloud::auth;
 use crate::cloud::client::CloudClient;
 use crate::commands::ExitCode;
 use crate::credentials;
 use crate::options::CloudOptions;
-use crate::portable::local::is_valid_name;
+use crate::portable::local::is_valid_instance_name;
 use crate::print::{self, echo, err_marker, Highlight};
 use crate::question;
 use crate::table::{self, Cell, Row, Table};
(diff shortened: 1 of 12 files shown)


src/cloud/ops.rs
```diff
@@ -141,43 +148,43 @@ pub async fn create_cloud_instance(
...
-    let instance = wait_instance_create(instance, client, false).await?;
-    write_credentials(&cred_path, instance).await?;
+    wait_instance_create(instance, client, false).await?;
     Ok(())
 }
 
+pub fn split_cloud_instance_name(name: &str) -> anyhow::Result<(String, String)> {
+    let mut splitter = name.splitn(2, '/');
+    match splitter.next() {
+        None => anyhow::bail!("unreachable"),
```
tailhook, line 162:
```suggestion
        None => unreachable!(),
```

src/portable/options.rs
```diff
@@ -466,10 +454,35 @@ impl fmt::Display for StartConf {
...
-                (regex: ^[a-zA-Z_][a-zA-Z_0-9]*$)".into())
+        regex: ^[a-zA-Z_][a-zA-Z_0-9]*$".into())
+}
+
+pub fn instance_or_cloud_name_opt(name: &str) -> Result<(), String> {
+    if is_valid_instance_name(&name) {
+        return Ok(())
+    }
+    if !name.contains("/") {
+        // todo: should this also mention the format for cloud instance names?
```
tailhook, line 469: Sure.

tailhook (approved):

src/portable/options.rs
fmoor: I'm not sure how releases work in this project. Are we worried about hinting at cloud instance names in a pre 2.0 release of the CLI?

fmoor: or at least before the cloud is actually available?

elprans: The upcoming cloud is not a secret, so it's fine to mention.

src/cloud/ops.rs
```diff
@@ -141,43 +148,43 @@ pub async fn create_cloud_instance(
...
-    if cred_path.exists() {
-        anyhow::bail!("File {} exists; abort.", cred_path.display());
-    }
+    let url = format!("orgs/{}/instances", instance.org_slug);
     let instance: CloudInstance = client
-        .post("instances/", serde_json::to_value(instance)?)
+        .post(url, serde_json::to_value(instance)?)
         .await?;
-    let instance = wait_instance_create(instance, client, false).await?;
-    write_credentials(&cred_path, instance).await?;
```
nsidnev, line 152: I think removing this line is a regression, no? I see one for linking the cloud instance, but nothing for creating it.

src/options.rs
```diff
@@ -629,7 +631,15 @@ pub fn conn_params(tmp: &ConnectionOptions) -> anyhow::Result<Builder> {
         task::block_on(bld.read_dsn(dsn))?;
         bld.read_extra_env_vars()?;
     } else if let Some(instance) = &tmp.instance {
-        task::block_on(bld.read_instance(instance))?;
+        if instance.contains("/") {
```
nsidnev, line 634: Shouldn't this PR also change the instance name parsing part of the Rust client before merging? Right now, for the name with format like `org_name/instance_name` CLI will complain about invalid characters for the name.

I think the problem is here: https://github.com/edgedb/edgedb-rust/blob/6b385b3c9f8c548fddc61cd53c1182c13c92ded7/edgedb-client/src/builder.rs#L296

elprans: Yes, the client validation should be updated too. `<org>/<instance>` should be automatically treated as a Cloud instance.

src/cloud/ops.rs
fmoor: The other one should be removed too. Thanks for catching that.

elprans (approved):

PR #727, opened 2022-05-30
respect `--project-dir` when applying migrations on new project

nsidnev:


src/portable/project.rs
```diff
@@ -192,6 +192,7 @@ pub struct Upgrade {
 pub struct Handle {
     name: String,
     instance: InstanceKind,
+    project_dir: Option<PathBuf>,
```
tailhook, line 195: I think `None` here inconsistently means either "no set/unknown" or "current dir". We probably should add `project_dir` argument to `Handle::probe`. And then probably do not make the field optional (if possible).

tailhook (commented):
In addition to line comments:
1. We probably can get rid of `project_dir` argument in `do_link` as it's now in Handle
2. I think current PR would ignore `--project-dir` in case of linking to existing instance, but consistently setting `handle.project_dir` should solve the issue (see line comments)

tailhook (approved):

tailhook:
Code looks good. But it's better to also add tests.

Since this touches a bunch of code branches we probably have to copy the whole `tests/portable_project.rs` and change all `current_dir` to `--project-dir=`. What do you think?

nsidnev:
> But it's better to also add tests. What do you think?

I'm in :) Thanks for pointing out how to add them, I'll try to do that soon!

nsidnev:
@tailhook I found that your commit with increasing threads for tests worked for this PR. If this is an acceptable solution, maybe I can `cherry-pick` it onto this branch?

tailhook:
Apparently, it works when tests are run in parallel (https://github.com/edgedb/edgedb-cli/pull/731/commits/90a3fa0ec2f3ef1ba786d4a11f0134a11a8fe049) and not if they don't :( I'm not sure what the reason of that is. But I think we can proceed with `RUST_TEST_THREADS: 3`.

nsidnev:
I think this could be merged? I'd like to continue working on #736 on top of that

tailhook:
Yes. Thanks!

PR #730, opened 2022-06-02
Use `execve` instead of spawning process for starting instance

tailhook:
@elprans take a look at env var name (and format?)


elprans (approved):
LGTM

PR #736, opened 2022-06-05
add support for custom schema directory

### LINKED ISSUE #732: Add support for custom `schema-directory` in `edgedb.toml`
I'd like to be able to move the `dbschema` directory with my schema and migrations from the root of my projects to the `priv` directory, which is used to store some additional files needed for the application in the Elixir world.

This is also one of the features described in the [projects RFC](https://github.com/edgedb/rfcs/blob/master/text/1005-edgedb-project.rst#project-file-structure), but which is not currently implemented in the CLI.

nsidnev:
I marked this one as a draft because it overlaps in some parts with #727, and IMO it is better to wait until the other one is merged.

Tested the new flow locally, but I can also try to add real tests for it.

BTW. 
~Maybe with this change we should also add new commands to handle project migrations? Something like `edgedb project migrate` + `edgedb project migration <action>` which will automatically use `schema-directory` from the project config?~

I suppose this PR should also add support for new config setting for `edgedb migrate`/`edgedb migration <action>` commands, although I'm not quite sure it's necessary in this particular PR.

Closes #732

### DIFF
--- a/src/portable/config.rs
+++ b/src/portable/config.rs
@@ -10,13 +10,11 @@ use crate::portable::repository::{Channel, Query};
 use crate::platform::tmp_file_path;
 use crate::print::{self, echo, Highlight};
 
-static DEFAULT_SCHEMA_DIR: &str = "dbschema";
-
 #[derive(serde::Deserialize)]
 #[serde(rename_all="kebab-case")]
 pub struct SrcConfig {
     pub edgedb: SrcEdgedb,
-    pub project: Option<toml::Spanned<SrcProject>>,
+    pub project: Option<SrcProject>,
     #[serde(flatten)]
     pub extra: BTreeMap<String, toml::Value>,
 }
@@ -34,7 +32,7 @@ pub struct SrcEdgedb {
 #[serde(rename_all="kebab-case")]
 pub struct SrcProject {
     #[serde(default)]
-    pub schema_dir: Option<toml::Spanned<String>>,
+    pub schema_dir: Option<String>,
     #[serde(flatten)]
     pub extra: BTreeMap<String, toml::Value>,
 }
@@ -53,7 +51,7 @@ pub struct Edgedb {
 
 #[derive(Debug)]
 pub struct Project {
-    pub schema_dir: Option<PathBuf>
+    pub schema_dir: PathBuf
 }
 
 fn warn_extra(extra: &BTreeMap<String, toml::Value>, prefix: &str) {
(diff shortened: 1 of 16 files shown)


tailhook:
> I suppose this PR should also add support for new config setting for `edgedb migrate`/`edgedb migration <action>` commands, although I'm not quite sure it's necessary in this particular PR.

I'm not sure what do you mean by that. But we decided on the team call, that we want migration commands to always read `edgedb.toml` if that exists (regardless of the command connects to). I.e. algorithm is the following:
1. If there is `--schema-dir` use that
2. Search for `edgedb.toml` normally (i.e. current dir, parent dirs), if it's present, check for the setting. If there is no setting use `dbschema` in the same dir as `edgedb.toml`
3. Use `dbschema` from the current dir

The high level idea here is that we already read local files (`./dbschema`) so reading project config is also fine.

> Tested the new flow locally, but I can also try to add real tests for it.

Yes. I think we can repurpose one of the existing migration tests to work with a different schema directory.

tailhook:
Quick note: I'm not sure we want to put `schema-directory = "dbschema"` in the generated config by default. I have to think about that more.

src/portable/config.rs
```diff
@@ -24,6 +24,8 @@ pub struct SrcConfig {
 pub struct SrcEdgedb {
     #[serde(default)]
     pub server_version: Option<toml::Spanned<Query>>,
+    #[serde(default)]
+    pub schema_directory: Option<toml::Spanned<String>>,
```
tailhook, line 28: I think we want `schema_dir`. Since we already consistently use that contraction in command-line (`--schema-dir` and `--project-dir`). I don't think remembering when to abbreviate and when not is a good idea.

nsidnev:
> The high level idea here is that we already read local files (./dbschema) so reading project config is also fine.

Do we need to add `--project-dir` for `edgedb migration` commands, so that the user can run migrations from custom directories or do we expect the user to specify the path explicitly?

Having the option for command will be useful in the `setup-edgedb` action, where `--project-dir` will be supported.

tailhook:
On config format:
1. We want to put `schema-dir` into `[project]` section.
2. We don't want to have that section+setting to be present in configuration file if it's the default value

> Do we need to add `--project-dir` for `edgedb migration` commands so that the user can run project migrations from any directories, or do we expect the user to explicitly specify the path to the migrations directory in `-schema-dir`?

I think we want `--project-dir` to be connection optino (i.e. present in all commands that have connection options). But this can be a separate PR.

> Having the `--project-dir` option for the command will be useful in the `edgedb/setup-edgedb` action, where `--project-dir` will be supported soon, but I am not sure how to properly handle the use of `--project-dir` and `--schema-dir` at the same time. Maybe prohibit passing both at once and show an error to the user?

There is no reason to prohibit that. `--schema-dir` as more specific one takes precedence.

nsidnev:
> We want to put schema-dir into [project] section.

I tried to implement this by adding a new `project: Option<toml::Spanned<SrcProject>>>` field to `SrcConfig`, so that the current changes tracking logic for config can be kept. Unfortunately, I encountered into a [problem](https://github.com/alexcrichton/toml-rs/issues/451) with the toml library for such field in struct. Right now I am unsure what is the best way to solve that. Could you give some advice on that, please?

I pushed a commit with the new changes to this branch, so you can look at it if you want, even though it's not working correctly at the moment.

tailhook:
I don't think we need to modify `schema-dir` programmatically. I.e. if you do `edgedb instance upgrade --schema-dir=xxx` where schema dir is something different from value in config, we just take that and ignore the config.

Then we don't need to keep track of the location of the schema dir.

Does this solve the issue?

nsidnev:
> Does this solve the issue?

Yes, that makes things a lot easier :)

PR #740, opened 2022-06-06
Use user session domain instead of GUI domain on macOS for instances

elprans:
The `gui/<uid>` domain requires an active GUI login session, which
might not be the case if one logs in to a machine via SSH.
The `user/<uid>` works in both scenarios.  An caveat is that, unlike
`gui/`, the `user/` domain does not terminate when a user logs out.

### DIFF
--- a/src/portable/control.rs
+++ b/src/portable/control.rs
@@ -112,7 +112,7 @@ pub fn detect_supervisor(name: &str) -> bool {
     if cfg!(windows) {
         false
     } else if cfg!(target_os="macos") {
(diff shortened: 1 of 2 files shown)


tailhook (approved):
I think I couldn't get it running on login with `user/`. But maybe the `LimitLoadToSessionType: Background` was the issue.

elprans:
> But maybe the `LimitLoadToSessionType: Background` was the issue.

Yeah, this key is required for `user/` services.

PR #745, opened 2022-06-10
Socket activation

tailhook:


src/portable/create.rs
```diff
@@ -30,6 +30,11 @@ pub fn create(options: &Create) -> anyhow::Result<()> {
         );
         return Err(ExitCode::new(exit_codes::DOCKER_CONTAINER))?;
     }
+    if options.start_conf.is_some() {
+        print::warn("The option `--start-conf` is deprecated. \
+                     Use `edgedb instance start/stop` to control \
+                     the instance.");
```
nsidnev, line 36: I think after merging this a new issue should be created in https://github.com/edgedb/setup-edgedb that uses this option right now

src/portable/control.rs
```diff
@@ -61,13 +61,15 @@ pub fn do_start(inst: &InstanceInfo) -> anyhow::Result<()> {
     }
 }
 
-pub fn get_server_cmd(inst: &InstanceInfo) -> anyhow::Result<process::Native> {
+pub fn get_server_cmd(inst: &InstanceInfo, is_shutdown_supported: bool)
```
elprans, line 64:
```suggestion
pub fn get_server_cmd(inst: &InstanceInfo, auto_shutdown: bool)
```

tailhook:
@jaclarke this also adds `server-status: "ready"` to `instance list/status --json` that means database is not running, but can be activated by connecting to the socket.

src/portable/control.rs
elprans: All supported server versions support auto-shutdown.  What's the concern here?

elprans:
@tailhook, do service unit files get re-written on instance upgrade?

src/portable/control.rs
tailhook: I think we support socket activation only for 2.0. So even if 1.x supports shutdown it's bad idea to enable shutdown there.

tailhook:
> @tailhook, do service unit files get re-written on instance upgrade?

Yes. It's kept from the times where we had to change the path to the binary.

tailhook:
I've just fixed "ready" status on macos.
Should be ready to go. But still didn't have a chance to test Windows

src/portable/create.rs
colinhacks: Created this pre-emptively so we don't forget: https://github.com/edgedb/setup-edgedb/issues/32

tailhook:
Updated windows. Should be ready for merge if tests pass.

elprans (approved):

PR #749, opened 2022-06-24
Fix interactive mode with TLS

### LINKED ISSUE #738: instance link --trust-tls-cert only seems to work in non-interactive mode
see also https://github.com/edgedb/edgedb-cli/issues/734

The edgedb instance link command below seems to surface a bug in the nightly version of edgedb CLI. I tested using EdgeDB CLI 1.2.0-dev.764+610eaa8 and it fails with the invalid peer certificate: CertNotValidForName error. However the same command succeeds for me using EdgeDB CLI 1.1.2+58eb29e.

```
edgedb instance link \
   --trust-tls-cert \
   --host <ip-or-dns> \
   --port 5656 \
   --user edgedb \
   --database edgedb \
   aws
```

Using the --non-interactive version listed below works for me on nightly.

```
echo -n "> " && read -s PASSWORD
printf $PASSWORD | edgedb instance link \
  --password-from-stdin \
  --trust-tls-cert \
  --non-interactive \
  --host <ip-or-dns> \
  aws
```

tailhook:
Fixes #738

To do:
1. [x] This works but I'm not sure this is right solution
2. [x] Look at the To Do's in `InteractiveCertVerifier`

### DIFF
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -1,9 +1,7 @@
-use std::sync::Arc;
 use std::time::Duration;
 
 use async_std::future::{timeout, pending};
 use async_std::prelude::FutureExt;
-use rustls::client::ServerCertVerifier;
 
 use edgedb_client::{Builder, Config};
 use edgedb_client::errors::Error;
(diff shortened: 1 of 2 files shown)


fantix (approved):
I think this is good.

fmoor (approved):

PR #755, opened 2022-07-02
Support range types

msullivan:


src/print/native.rs
```diff
@@ -220,6 +220,36 @@ impl FormatExt for Value {
                 })
             }
             V::Enum(v) => prn.const_enum(&**v),
+            V::Range { lower, upper, inc_lower, inc_upper, empty } => {
+                prn.call("range", |prn| {
+                    lower.format(prn)?;
+                    prn.comma()?;
+                    upper.format(prn)?;
```
elprans, line 227: I think we should omit one of the bounds for empty ranges for shorter representation.

msullivan: That isn't valid on the input side though, right?

elprans (approved):

src/print/native.rs
elprans: It is perfectly valid.  Furthermore, the only reason we require you to specify at least one bound is for type inference.

PR #757, opened 2022-07-05
Generate JWT auth token for `edgedb ui` command

fantix:
See also: https://github.com/lawliet89/biscuit/blob/master/doc/supported.md#jwe-algorithms-for-key-management


src/portable/options.rs
```diff
@@ -57,6 +57,9 @@ pub enum InstanceCommand {
     Revert(Revert),
     /// Reset password for a user in the instance
     ResetPassword(ResetPassword),
+    #[clap(name="_ui_token")]
+    #[edb(hide=true)]
+    _UIToken(UIToken),
```
tailhook, line 62: Since we use `clap(name=..)` let's remove underscore from the variant name.

src/portable/windows.rs
```diff
@@ -866,6 +866,21 @@ pub fn revert(options: &options::Revert) -> anyhow::Result<()> {
...
+    let token = wsl.edgedb()
+        .arg("instance")
+        .arg("_ui_token")
+        .arg(name)
+        .get_stdout_text()?;
+    Ok(token)
+}
+
+pub fn ui_token(name: &str) -> anyhow::Result<()> {
+    println!("{}", get_ui_token(name)?);
```
tailhook, line 880: nit: this looks like will print double linefeed? (one from the captured text and one from println here).

src/commands/ui.rs
```diff
@@ -14,3 +37,99 @@ pub fn show_ui(options: &Options) -> anyhow::Result<()> {
...
+    let jwe = signature::EcdsaKeyPair::from_pkcs8(
+        &signature::ECDSA_P256_SHA256_FIXED_SIGNING,
+        pem.contents.as_slice(),
+    )?;
+
+    let priv_key = agreement::EphemeralPrivateKey::generate(&agreement::ECDH_P256, &rng)?;
+    let pub_key =
+        agreement::UnparsedPublicKey::new(&agreement::ECDH_P256, jwe.public_key().as_ref());
+    let epk = priv_key.compute_public_key()?.as_ref().to_vec();
+    let cek = agreement::agree_ephemeral(priv_key, &pub_key, ExitCode::new(1), |key_material| {
```
tailhook, line 96: I don't think `ExitCode::new(1)` is a good idea here. We use `ExitCode` only in places where we print error description before returning `ExitCode`. So you could either use `anyhow::anyhow!(..)` here. But this perhaps has some cost on the success path. So I would use bare `()` and `.map_err(` with proper error formatting.

tailhook (approved):
Mostly looks good. Better error reporting is the most important fix.

src/commands/ui.rs
```diff
@@ -1,11 +1,34 @@
...
+        match if cfg!(windows) {
+            crate::portable::windows::get_ui_token(instance)
+        } else {
+            generate_jwt(instance)
+        } {
+            Ok(token) => {
+                url = format!("{}?authToken={}", url, token);
+            }
+            Err(e) => {
+                print::warn(format!("Cannot generate authToken: {:#}", e));
```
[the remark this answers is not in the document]
fantix: Oh because here it still opens the UI in the browser, only without `authToken` set. The UI somehow may already have the token set in the browser cookie - like maybe it's just a linked remote instance I guess?

PR #764, opened 2022-07-07
Drop the hidden _ui_token command

fantix:
Looks like there is a way to read WSL files directly, so we don't need
the _ui_command command anymore.

### DIFF
--- a/src/commands/ui.rs
+++ b/src/commands/ui.rs
@@ -9,7 +9,7 @@ use ring::{aead, agreement, digest, rand, signature};
 use crate::commands::ExitCode;
 use crate::options::Options;
 use crate::platform::data_dir;
-use crate::portable::local::instance_data_dir;
+use crate::portable::local::{instance_data_dir, InstanceInfo};
 use crate::print;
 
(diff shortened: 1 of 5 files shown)


elprans (commented):
We use it in the recently added `edb ui`. Do you have a plan for a replacement there?

src/commands/ui.rs
```diff
@@ -9,19 +9,22 @@ use ring::{aead, agreement, digest, rand, signature};
...
     if let Some(instance) = &options.conn_options.instance {
-        match if cfg!(windows) {
-            crate::portable::windows::get_ui_token(instance)
-        } else {
-            generate_jwt(instance)
-        } {
+        if instance != "_localdev" {
+            let inst = InstanceInfo::read(instance)?;
+            if inst.get_version()?.specific().major < 2 {
+                print::error("Instance doesn't have UI to open; consider upgrade it.");
```
elprans, line 23:
```suggestion
                print::error("the specified instance runs EdgeDB v<version>, which does not include the Web UI; consider upgrading the instance to EdgeDB 2.x or later");
```

PR #766, opened 2022-07-08
Fix rendering of ranges and implicit limit

### LINKED ISSUE #765: Float and decimal ranges are rendered inconsistently.
The float and decimal ranges don't show inclusivity of the bounds.

Steps to Reproduce:
```
EdgeDB 2.0-dev.6716+local (repl 1.2.0)
Type \help for help, \quit to quit.
edgedb> select range(1.0, 7.0, inc_lower := false);
{range(1, 7)}
edgedb> select range(1, 7, inc_lower := false);
{range(2, 7, inc_lower := true, inc_upper := false)}
edgedb> select range(1.0n, 7.0n, inc_lower := false);
{range(1.0n, 7.0n)}
```

Contrast that with:
```
edgedb> select range(1.0);
{range(1, {}, inc_lower := true, inc_upper := false)}
edgedb> select range(<cal::local_date>'2022-07-07', <cal::local_date>'2022-07-17',
);
{
  range(
    <cal::local_date>'2022-07-07',
    <cal::local_date>'2022-07-17',
    inc_lower :=     true,
    inc_upper :=     false,
  ),
}
edgedb> select range(datetime_current(), datetime_current()+<duration>'10:00:00');

{
  range(
    <datetime>'2022-07-08T02:54:26.751466Z',
    <datetime>'2022-07-08T12:54:26.751469Z',
    inc_lower :=     true,
    inc_upper :=     false,
  ),
}
```

The bug seems to be specific to `float32`, `float64` and `decimal` and only when both bounds are present. Unbounded ranges for these types are rendered with all the relevant details.

tailhook:
Fixes #765

### DIFF
--- a/src/interactive.rs
+++ b/src/interactive.rs
@@ -285,7 +285,7 @@ async fn execute_query1(options: &Options, mut state: &mut repl::State,
             compilation_flags:
                 CompilationFlags::INJECT_OUTPUT_OBJECT_IDS |
                 CompilationFlags::INJECT_OUTPUT_TYPE_NAMES,
-            implicit_limit: state.implicit_limit.map(|x| x as u64),
+            implicit_limit: state.implicit_limit.map(|x| (x+1) as u64),
             output_format: match state.output_format {
                 Default | TabSeparated => IoFormat::Binary,
                 JsonLines | JsonPretty => IoFormat::JsonElements,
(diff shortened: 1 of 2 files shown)


src/interactive.rs
```diff
@@ -285,7 +285,7 @@ async fn execute_query1(options: &Options, mut state: &mut repl::State,
             compilation_flags:
                 CompilationFlags::INJECT_OUTPUT_OBJECT_IDS |
                 CompilationFlags::INJECT_OUTPUT_TYPE_NAMES,
-            implicit_limit: state.implicit_limit.map(|x| x as u64),
+            implicit_limit: state.implicit_limit.map(|x| (x+1) as u64),
```
elprans, line 288: Why is this change? It'll break limit set to zero (no limit)

tailhook: "No limit" is represented as `None` both in `state.implicit_limit` and `Execute1::implicit_limit`. And `map` is only applied to `Some(x)`. We avoid using special values like zero use enums instead in Rust.

We need `+1` to display `...` that was broken in protocol 1.0 until this change.

PR #770, opened 2022-07-14
Minimal processing of concurrent state/global changes

tailhook:


fantix (approved):
Tested, works great for me!

PR #773, opened 2022-07-15
Add --print-url for `edgedb ui` to only show URL

fantix:
Also unhide the `ui` command in `--help`.

### DIFF
--- a/src/commands/ui.rs
+++ b/src/commands/ui.rs
@@ -38,16 +38,16 @@ pub fn show_ui(options: &Options, args: &UI) -> anyhow::Result<()> {
             }
         }
     }
-    if args.url {
-        print::echo!(url);
-        Ok(())
-    }
-    else if open::that(&url).is_ok() {
-        Ok(())
-    } else {
+    if !args.url && !open::that(&url).is_ok() {
         print::error("Cannot launch browser, please visit URL:");
-        print::echo!("  ", url);
+        println!("{}", url);
         Err(ExitCode::new(1).into())
+    } else {
+        if !args.url {
+            print::success("Opening URL in browser:");
+        }
+        println!("{}", url);
+        Ok(())
     }
 }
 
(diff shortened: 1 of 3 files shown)


1st1 (approved):
I also want @elprans to look at this before merge in case he wants to consider naming the option differently

elprans:
This is fine, but maybe we should just always print the URL? This is what URL-opening CLI commands normally do in case the browser doesn't open.

src/commands/ui.rs
```diff
@@ -38,11 +38,15 @@ pub fn show_ui(options: &Options) -> anyhow::Result<()> {
             }
         }
     }
-    if open::that(&url).is_ok() {
+    if args.url {
+        print::echo!(url);
```
tailhook, line 42: It should go to stdout. At least in `--url` case. So you can do:
```
url=$(edgedb ui --url)
```

tailhook (changes requested):

elprans:
> That's the thing -- i don't want the browser to open, i just want to see the url and open it myself in the browser i want

Sure, it's a separate issue.  Always printing the URL is for cases where opening the default browser didn't work for some reason so at least you can click on the link and don't have to go looking for the `--url` option.

1st1:
Ok, let's make it always print the URL.

src/commands/ui.rs
```diff
@@ -38,12 +38,16 @@ pub fn show_ui(options: &Options) -> anyhow::Result<()> {
             }
         }
     }
-    if open::that(&url).is_ok() {
-        Ok(())
-    } else {
+    if !args.url && !open::that(&url).is_ok() {
```
tailhook, line 41: Why we don't print error returned by `open::that` ? Sometimes users may want to see why we couldn't open a browser.

tailhook (approved):

src/commands/ui.rs
fantix: Fixed!

PR #780, opened 2022-07-21
Build for aarch64-apple-darwin

elprans:


fantix (approved):
LGTM but it's probably better if @tailhook takes a look too

tailhook (approved):

PR #784, opened 2022-07-25
Better error reporting in `edgedb ui` command

tailhook:
Old error:
```
Cannot generate authToken: No such file or directory (os error 2)
```

New error:
```
Cannot generate authToken: failed to open file `/work/tmp/.local/share/edgedb/_localdev/edbjwskeys.pem`: No such file or directory (os error 2)
```


fantix (approved):
Thanks!

PR #789, opened 2022-07-27
Fix aarch64 linux support

Syzuna:
Currently even with aarch64 build out there that seem to start and run fine, the CLI did not support them because they were not added as possible targets in platform checks.
This PR aims to fix that issue so the CLI can be used to install and manage EdgeDB on linux on the aarch64 architecture.

### DIFF
--- a/edgedb-init.sh
+++ b/edgedb-init.sh
@@ -177,7 +177,7 @@ get_architecture() {
     esac
 
     if [ "$_cputype" = "aarch64" ]; then
-        if [ "$_ostype" != "apple-darwin" && "$_ostype" != "unknown-linux-musl" ]; then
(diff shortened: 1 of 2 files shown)


elprans (approved):

edgedb-init.sh
```diff
@@ -177,7 +177,7 @@ get_architecture() {
     esac
 
     if [ "$_cputype" = "aarch64" ]; then
-        if [ "$_ostype" != "apple-darwin" ]; then
+        if [ "$_ostype" != "apple-darwin" && "$_ostype" != "unknown-linux-musl" ]; then
```
elprans, line 180:
```suggestion
        if [ "$_ostype" != "apple-darwin" ] && [ "$_ostype" != "unknown-linux-musl" ]; then
```

elprans:
Thanks!

PR #796, opened 2022-08-01
Add --yolo option

quinchs:
## Summary
as discussed in [discord](https://discord.com/channels/841451783728529451/841451783728529454/1003731814143250522) there should be a `--yolo` flag thats short hand for `--non-interactive`, `allow_empty`, and `allow_unsafe`, this PR adds exactly that.


colinhacks:
I like this idea unironically.

tailhook:
Option is added, but it doesn't do anything, except blocking other options in this PR :)

What's the use case for `--allow-empty` here? If you want "just apply the schema" then empty migration is not something useful. We might better return with success exit code but without creating a migration in `--yolo` mode.

*(I think `--allow-empty` was added for completeness, because there is such an option in git, but might be a little bit useful for writing (possibly data) migrations manually; although, I think we need a better tool for them)*

quinchs:
> Option is added, but it doesn't do anything, except blocking other options in this PR :)

Testing locally it has the same effect as specifying all flags

> What's the use case for `--allow-empty` here? 

For completeness, it can defiantly be removed

1st1 (changes requested):
Sorry for being party pooper here, but i'm -1 on calling it "yolo" no matter how funny it seems to be now.

tailhook:
> > Option is added, but it doesn't do anything, except blocking other options in this PR :)
> 
> Testing locally it has the same effect as specifying all flags
> 

This shouldn't be the case. We automatically apply that for the first migration, maybe this is what you've tested it with? (otherwise we have a bug somewhere in our clap generator, I'll take a look)

> Sorry for being party pooper here, but i'm -1 on calling it "yolo" no matter how funny it seems to be now.

I think we can just assume `--non-interactive` in `--allow-unsafe`? (because latter has no action in interactive mode anyways).

elprans:
IMO we should just add a short option for `--non-interactive`, e.g. `-y` and `-f` for `--allow-unsafe`, so  instead of `create --yolo` you'd type `create -yf`

1st1:
> 100% of people would instantly know the implications of --yolo

I see the point, but I don't agree that 100% of people would instantly understand everything that it does just by looking at its name. E.g. I would still have to check with the docs. Moreover, overarching names like this seem intuitive at first, but imagine that in the future we want to include another option/behavior to `--yolo` because...it would make sense given the name. We might not able to do it, because that would be technically an incompatible change. So you'd end up teaching people about `--yolo --some-other-annoying-flag` etc.

In general, it's a bad idea to use non-technical jargon in API design. It rarely ages well, and I don't want our core tool chain to evolve into something barely comprehensible like PHP standard library. And it's not a popular approach to naming in other tools either, otherwise instead of `git push --force` we'd have `git push --yolo`.

/2c

i0bs:
There are plenty of words with aberration to technical meaning used in today's devtools. I believe `--yolo` perfectly captures the scope of what's being requested and adds an unique part to EdgeDB that makes it stand out a little more than others.

You can't tell me "yeah I use MySQL, but have you ever yolo'd an EdgeDB?" does not sound poggers

i0bs:
There are plenty of words with aberration to technical meaning used in today's devtools. I believe `--yolo` perfectly captures the scope of what's being requested and adds an unique part to EdgeDB that makes it stand out a little more than others.

You can't tell me "yeah I use MySQL, but have you ever yolo'd an EdgeDB?" does not sound poggers

1st1:
Alright, alright, I admit "but have you ever yolo'd an EdgeDB?" has something in it. @elprans seems to also be OK with this. Let's add it. Although, we should also support `-y` as an alias for `--yolo` and it shouldn't include `--allow-empty` as @tailhook pointed out.

quinchs:
> Let's add it. 

I think the current changes I've made does in fact do nothing except add it as a valid flag, for implementation properly imma leave that up to tailhook as I suck at rust 👀

i0bs:
To further clarify, `--allow-empty` is a pre-emptive flag so it makes sense for it to not be required in the preset. I would like to ask though if `-y` is intentionally meant to be lowercase, as most single-lettered flags are uppercased like `-Y`.

tailhook:
In nightly, we have `edgedb migrate --dev-mode` and ([edgedb watch](https://github.com/edgedb/edgedb/discussions/4682)). They are not the same thing as `edgedb migration create --yolo`. But I think this is what was `--yolo`'s primary use case.

Please try new watch/dev-mode workflow. And feel free to reopen describing the use case that doesn't fit that workflow.

PR #798, opened 2022-08-02
Listen on IPv6 by default too

### LINKED ISSUE #797: Local instance listen on IPv6 ::1 too
@colinhacks:

> Node 18 now resolves localhost to ::1 by default which causes connection issues because the instance doesn’t listen on ::1  by default.

Only listening on `127.0.0.1` was added in #745, and edgedb/edgedb#3899 is overwriting instance config with socket activation config, so this issue cannot be worked around with a `edgedb configure set listen_addresses 127.0.0.1 ::1`.

Ideally, the local instance could listen on `::1` too.

tailhook:
Fixes #797

### DIFF
--- a/src/portable/linux.rs
+++ b/src/portable/linux.rs
@@ -5,7 +5,7 @@ use std::path::{PathBuf};
 use anyhow::Context;
 use fn_error_context::context;
 
-use crate::platform::{home_dir, current_exe};
+use crate::platform::{home_dir, current_exe, detect_ipv6};
 use crate::portable::destroy::InstanceNotFound;
 use crate::portable::local::{InstanceInfo, runstate_dir, log_file};
 use crate::portable::options::{Logs, instance_arg};
@@ -102,14 +102,20 @@ Description=EdgeDB Database Service socket, instance {instance_name:?}
 Documentation=https://edgedb.com/
 
 [Socket]
-ListenStream=127.0.0.1:{port}
 FileDescriptorName=edgedb-server
+ListenStream=127.0.0.1:{port}
+{ipv6_listen}
 
 [Install]
 WantedBy=default.target
     "###,
         instance_name=name,
         port=info.port,
+        ipv6_listen=if detect_ipv6() {
+            format!("ListenStream=[::1]:{port}", port=info.port)
+        } else {
+            String::new()
+        },
     ))
 }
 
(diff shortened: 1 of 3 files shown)


src/portable/linux.rs
```diff
@@ -103,6 +103,7 @@ Documentation=https://edgedb.com/
 
 [Socket]
 ListenStream=127.0.0.1:{port}
+ListenStream=[::1]:{port}
```
elprans, line 106: What would happen if the system has no support for IPv6?  Should we listen on `localhost:{port}` instead?

fantix: I've been testing this, looks like the service will fail to start in that case. `localhost` is not always good because:

1. Some distribution don't have `::1 localhost` in `/etc/hosts` - my Ubuntu Jammy has only `::1     ip6-localhost ip6-loopback`
2. GitHub Action has `::1 localhost` but it doesn't support IPv6 if I remember correctly

1st1: Can't we have a second Socket unit for "ListenStream=[::1]:{port}" not `WantedBy` by the service? Which would presumably make it an optional socket to start with?

tailhook: I think we can try listening on `[::1]:0` in the CLI and then create an appropriate unit?

elprans: Oh, there's a better way.

```
ListenStream={port}
```

This way systemd will automatically listen on all available IP versions.

elprans: Nevermind, this will also listen on all interfaces, which is not what we want.

tailhook: > Can't we have a second Socket unit for "ListenStream=[::1]:{port}" not `WantedBy` by the service? Which would presumably make it an optional socket to start with?

Even if that trick will work on linux, I think there is no such feature on macos.

elprans: AFAIK non-working IPv6 on macOS is probably a lot less likely, so that's fine.

elprans: > I think we can try listening on `[::1]:0` in the CLI and then create an appropriate unit?

This seems like the most reliable approach to me.

tailhook:
This should not be merged until edgedb/edgedb#4189, right?

fantix:
Well it's okay to merge this I think, it's just that the server will "randomly" pick one of `127.0.0.1` or `::1` to listen on before https://github.com/edgedb/edgedb/pull/4189, no failure as I tested.

elprans:
Please merge.

fantix:
I've manually tested this on macOS - the CI would need a new nightly server to pass I think.

elprans (approved):

PR #802, opened 2022-08-03
M1 upgrade

tailhook:


src/portable/platform.rs
```diff
@@ -96,9 +96,14 @@ pub fn optional_docker_check() -> anyhow::Result<bool> {
 }
 
 pub fn is_arm64_hardware() -> bool {
-    process::Native::new("uname", "uname", "uname")
+    process::Native::new("uname", "uname", "arch")
```
elprans, line 99: This prints "arch: Unknown architecture: arm64e" to stderr on x86 macs.  Can we inhibit stderr?  Or maybe we can use `uname -m` and then detect Rosetta explicitly (`$(sysctl -in sysctl.proc_translated) = 1`).

tailhook:
`sysctl.proc_translated` doesn't give us information which arch is native for this platform. But should be good enough for next 10 years.

elprans (approved):

PR #808, opened 2022-08-06
Support opening UI for remote instances

### LINKED ISSUE #807: Add support for opening the UI on non-local instances
For remote instances `edgedb ui` should simply open the UI URL without the auth token and allow the user to log in manually.  Right now the CLi fails with an unhelpful error like

```
edgedb error: error reading instance info: "<data-dir>/<instance>/instance_info.json": No such file or directory (os error 2)
```

fantix:
Added `edgedb ui --skip-version-check` because the CLI will now try to connect to the remote instance for version check before opening the browser.

Fixed #807

### DIFF
--- a/src/commands/ui.rs
+++ b/src/commands/ui.rs
@@ -16,7 +16,7 @@ use ring::{aead, agreement, digest, rand, signature};
 use crate::commands::ExitCode;
 use crate::options::{Options, UI};
 use crate::platform::data_dir;
-use crate::portable::local::{instance_data_dir, InstanceInfo};
+use crate::portable::local::{instance_data_dir, InstanceInfo, NonLocalInstance};
 use crate::portable::ver::Specific;
 use crate::print;
 
@@ -25,7 +25,7 @@ pub fn show_ui(options: &Options, args: &UI) -> anyhow::Result<()> {
     let builder = connector.get()?;
     let mut url = format!("http://{}:{}/ui", builder.get_host(), builder.get_port());
 
-    if !args.skip_version_check {
+    if !args.no_server_check {
         let version = if let Some(meta) = builder
             .get_instance_name()
             .map(InstanceInfo::try_read)
@@ -47,11 +47,7 @@ pub fn show_ui(options: &Options, args: &UI) -> anyhow::Result<()> {
     }
 
     if let Some(instance) = builder.get_instance_name() {
-        match if cfg!(windows) {
-            crate::portable::windows::read_jose_keys(instance)
-        } else {
-            read_jose_keys(instance)
-        } {
+        match read_jose_keys(instance) {
             Ok(keys) => match generate_jwt(keys) {
                 Ok(token) => {
                     url = format!("{}?authToken={}", url, token);
@@ -60,12 +56,12 @@ pub fn show_ui(options: &Options, args: &UI) -> anyhow::Result<()> {
                     print::warn(format!("Cannot generate authToken: {:#}", e));
                 }
             },
+            Err(e) if e.is::<NonLocalInstance>() => {
+                // Continue without token for remote instances
+                log::debug!("Assuming local instance because: {:#}", e);
+            }
             Err(e) => {
-                // Ignore the error assuming it's a remote instance
-                log::debug!(
-                    "Cannot find JOSE key files ({:#}), assuming remote instance.",
-                    e,
-                )
+                print::warn(format!("Cannot read JOSE key files: {:#})", e));
             }
         }
     }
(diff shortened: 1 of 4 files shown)


1st1:
Hm, can we detect that the instance is remote and imply this flag? Also, the option naming is confusing -- version check of what? Of the CLI? Of the server? Of the UI version?

fantix:
> Hm, can we detect that the instance is remote and imply this flag? Also, the option naming is confusing -- version check of what? Of the CLI? Of the server? Of the UI version?

Right ... remote instances could also be EdgeDB 1.0 I guess? If that's too marginal, I guess we don't need to check remote instance versions, and always check local instance version for the better error message if <2.0.

src/commands/ui.rs
```diff
@@ -11,36 +17,63 @@ use crate::commands::ExitCode;
...
+                Ok(token) => {
+                    url = format!("{}?authToken={}", url, token);
+                }
+                Err(e) => {
+                    print::warn(format!("Cannot generate authToken: {:#}", e));
+                }
+            },
             Err(e) => {
-                print::warn(format!("Cannot generate authToken: {:#}", e));
+                // Ignore the error assuming it's a remote instance
```
tailhook, line 64: Please emit a special error and use `e.is::<NonLocalInstance>()`. Swallowing all the errors will make it hard to debug some non-obvious case.

tailhook:
> Hm, can we detect that the instance is remote and imply this flag? Also, the option naming is confusing -- version check of what? Of the CLI? Of the server? Of the UI version?

Well, why we have to imply it. It's better to always check and give a nice error. This is quite niche flag that maybe used to make edgedb print link quicker in scripts, or whatever?

On the other hand, we have `edgedb --no-version-check` so it can be easily confused with `edgedb ui --skip-version-check`. So probably `--no-server-check` is a better name? (I'm skipping `-version-` for conciseness, and also to note that we don't connect to server at all in this mode).

fantix:
Over the meeting, Elvis suggested we just try to retrieve the UI URL in CLI before opening, and provide a friendly error on 404 etc.

src/portable/windows.rs
```diff
@@ -886,8 +886,12 @@ pub fn read_jose_keys(name: &str) -> anyhow::Result<(String, String)> {
     };
 
     Ok((
-        wsl.read_text_file(data_dir.clone() + "edbjwskeys.pem")?,
-        wsl.read_text_file(data_dir + "edbjwekeys.pem")?,
+        wsl.read_text_file(data_dir.clone() + "edbjwskeys.pem")
+            .map_err(NonLocalInstance)?
```
tailhook, line 890: I think `if not data_dir.exists() { return Err(NonLocalInstance); }`. Other errors should be propagated.

src/commands/ui.rs
```diff
@@ -59,33 +88,39 @@ pub fn show_ui(options: &Options, args: &UI) -> anyhow::Result<()> {
     }
 }
 
+#[tokio::main]
+async fn open_url(url: &str) -> Result<reqwest::Response, reqwest::Error> {
+    reqwest::Client::new().get(url).send().await
```
tailhook, line 93: I would add `User-Agent: edgedb` like we use for downloads.
Also might use `HEAD` method instead, as you don't read the response body anyway.

src/commands/ui.rs
```diff
@@ -10,37 +10,66 @@ use ring::{aead, agreement, digest, rand, signature};
...
+            Ok(reqwest::StatusCode::OK) => (),
+            Ok(reqwest::StatusCode::NOT_FOUND) => {
+                print::error("the specified EdgeDB server is not serving Web UI.");
+                print::echo!(
+                    "  If you have EdgeDB 2.0 and above, try to run the \
+                    server with `--admin-ui=enabled`."
+                );
+                return Err(ExitCode::new(2).into());
+            }
+            Ok(_) => {
```
tailhook, line 32: Why are you hiding status code here? Print it at least in `log::info`.

tailhook: Also, we're accessing without token here. Does EdgeDB return 200 in such a case? I think it should return 403 or something.

src/commands/ui.rs
```diff
@@ -10,37 +10,66 @@ use ring::{aead, agreement, digest, rand, signature};
...
                 print::error(format!(
-                    "the specified instance runs EdgeDB v{}, which does not \
-                    include the Web UI; consider upgrading the instance to \
-                    EdgeDB 2.x or later",
-                    ver,
+                    "cannot connect to the specified EdgeDB server: {:#}",
+                    e,
                 ));
-                return Err(ExitCode::new(2).into())
+                return Err(ExitCode::new(4).into());
```
tailhook, line 44: Should this be warning instead? I think this check is merely informative. There still are reasons why you could connect via browser but not via CLI (i.e. proxy only set up in browser).

src/commands/ui.rs
```diff
@@ -10,37 +10,66 @@ use ring::{aead, agreement, digest, rand, signature};
...
 use crate::options::{Options, UI};
 use crate::platform::data_dir;
-use crate::portable::local::{instance_data_dir, InstanceInfo};
+use crate::portable::local::{instance_data_dir, NonLocalInstance};
 use crate::print;
 
 pub fn show_ui(options: &Options, args: &UI) -> anyhow::Result<()> {
     let connector = options.create_connector()?;
     let builder = connector.get()?;
     let mut url = format!("http://{}:{}/ui", builder.get_host(), builder.get_port());
```
tailhook, line 19: Only checking `http` looks wrong.

I think we can try `https` first (also we're supposed to have apropriate certificate for that, right?). And only then try http. No? Still not sure whether to print `http` after that (i.e. browser will probably don't have a certificate, but someone may use let's encrypt or corporate certificate that is installed in the browser).

This matters for remote instances. For localhost, any approach is fine (and then always print `http://`)

elprans: For secure endpoints there will be an http -> https redirect.

src/commands/ui.rs
fantix: Right I thought it was returning 401 too, but this is just the `index.html` that is either 200 or 404.

src/commands/ui.rs
fantix: Sure thing! But unfortunately the server doesn't support `HEAD`.

src/commands/ui.rs
fantix: > proxy only set up in browser

In that case - quoting Elvis - the CLI should usually have the `HTTP_PROXY` in its env and `reqwest` is supposed to handle that correctly. I'm convinced that in a dev env, if the CLI cannot open UI, the browser is likely unable to.

src/commands/ui.rs
tailhook: 1. If there is a redirect and reqwest follows redirect it should have a certificate, right? Or will we use redirect only when using globally recognized certificates?
2. Accessing HTTP then HTTPS is suboptimal as it already exposes some info in HTTP mode. So it would be great if we can figure out workflow that tries HTTPS first and only falls back to HTTP when TLS doesn't work for some reason.

src/commands/ui.rs
tailhook: At least print URL besides the error, so if proxy is set up in the wrong way, there is something to debug.

1st1:
@fantix are you working on the updated version? We should fix this asap.

src/commands/ui.rs
fantix: I'm doing a HTTPS check on remote instances anyways, so that we can get this merged. Please feel free to continue with the discussion.

> Accessing HTTP then HTTPS is suboptimal as it already exposes some info in HTTP mode.
We don't feed auth token to remote instances. Anything else I'm missing here?

elprans: Yeah, this check is solely to see if the URL in question is accessible, we aren't fetching or sending any data, so I think the HTTP/HTTPS dilemma here is moot.  For maximum compatibility the HTTP endpoints will always support connecting without TLS and will redirect to HTTPS.  Checking for cert validity here also doesn't really achieve much aside from adding friction, because you can "trust" a cert in your browser and it's much harder to do system-wide.

elprans (approved):

PR #809, opened 2022-08-06
Store password in remote instances

fantix:
`edgedb instance link` was not storing password in the generated credential files, this PR fixes that.

Q: was it intentional that password is not stored for remote instances, so that the user must specify `--password` option and type in the password everytime? Or it was just lost in the refactors?


tailhook (approved):
Thanks! It wasn't intentional.

PR #818, opened 2022-08-15
Ignore invalid certs checking UI through HTTPS

fantix:
Do we also want the `.danger_accept_invalid_hostnames(true)`?


elprans:
> Do we also want the `.danger_accept_invalid_hostnames(true)`?

Sure, the point of the check isn't to validate the cert, but to basically ping the URL, as the former will be done by the browser anyway.

elprans (approved):

PR #825, opened 2022-08-22
Fix publishing on linuxmusl

fantix:


elprans (approved):
And nightly?

fantix:
Ahh right! One sec

PR #828, opened 2022-08-23
Fix incomplete help for `--unix-path`

### LINKED ISSUE #827: Incomplete docs string
https://github.com/edgedb/edgedb-cli/blob/ff5426f54805f410f6fb0be8f2d2a442106aff6f/src/options.rs#L83

In the help text for `edgedb --help-connect`:

```
        --unix-path <unix_path>
            Unix socket dir for the
```

tailhook:
Fixes #827

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -80,7 +80,10 @@ pub struct ConnectionOptions {
     #[clap(conflicts_with_all=&["dsn", "credentials_file", "instance"])]
     pub port: Option<u16>,
 
-    /// Unix socket dir for the
+    /// Unix socket dir or path to unix socket for the EdgeDB connection
+    ///
+    /// When dir is specified here, path to the unix socket also depends on
+    /// `--port` and `--admin` parameters.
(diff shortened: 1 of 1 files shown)


src/options.rs
```diff
@@ -80,7 +80,10 @@ pub struct ConnectionOptions {
     #[clap(conflicts_with_all=&["dsn", "credentials_file", "instance"])]
     pub port: Option<u16>,
 
-    /// Unix socket dir for the
+    /// Unix socket dir or path to unix socket for the EdgeDB connection
+    ///
+    /// When dir is specified here, path to the unix socket also depends on
+    /// `--port` and `--admin` parameters.
```
1st1, line 86:
```suggestion
    /// A path to a unix socket for the EdgeDB connection
    ///
    /// When the supplied path is a directory, the actual path will be
    /// computed using the `--port` and `--admin` parameters.
```

1st1 (approved):

PR #842, opened 2022-09-23
Migration edit command

tailhook:


tailhook:
@elprans, take a look at CLI interface at least, probably.

elprans (commented):
The interface looks OK to me.

Do we want to prohibit editing applied migrations?

tailhook:
> Do we want to prohibit editing applied migrations?

With current impl it will implicitly prohibit, because checking via database will fail. Unless you do `--no-check` or `force` in interactive mode. We can't "allow" it in checked mode without abstract databases (or temporary database dance).

PR #857, opened 2022-10-13
List cloud instances by default

fantix:
Drop `--cloud` option in `edgedb instance list`, and list cloud instances in the same table if logged in.

### DIFF
--- a/src/cloud/ops.rs
+++ b/src/cloud/ops.rs
@@ -1,5 +1,6 @@
 use std::time::{Duration, Instant};
 
+use async_std::future::timeout;
 use async_std::task;
 use edgedb_client::credentials::Credentials;
 use edgedb_client::Builder;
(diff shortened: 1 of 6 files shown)


tailhook:
I think we should probably enforce some timeout on cloud instance list in addition to "intermediate feedback". Some 30-60 sec probably. On regular remote instances we have a timeout for each attempt I think.

tailhook:
> Drop `--cloud` option in `edgedb instance list`, and list cloud instances in the same table if logged in.

I also thing that some `--cloud[-only]` filter might make sense for scripting. But can be a follow up PR.

tailhook (commented):
Generally looks good, two things to be done:
1. Add a timeout
2. Figure out statuses

elprans (approved):

tailhook:
@fantix, one thing is missing here: if cloud is timeouted or otherwise inaccessible, we should print local (and probably remote) instances, then print error and exit with erroneous exit code. This is a common case for offline work, so list of local instances should be there to work without too much hassle.

fantix:
Right, let me create another PR.

PR #862, opened 2022-10-21
Implement cloud upgrade

nsidnev:


src/cloud/ops.rs
```diff
@@ -16,8 +16,8 @@ use crate::print::{self, echo, err_marker, Highlight};
 use crate::question;
 use crate::table::{self, Cell, Row, Table};
 
-const INSTANCE_CREATION_WAIT_TIME: Duration = Duration::from_secs(5 * 60);
-const INSTANCE_CREATION_POLLING_INTERVAL : Duration = Duration::from_secs(1);
+const INSTANCE_OPERATION_WAIT_TIME: Duration = Duration::from_secs(5 * 60);
```
tailhook, line 19:
```suggestion
const OPERATION_WAIT_TIME: Duration = Duration::from_secs(5 * 60);
```
When you'll have to split operations into instance and non-instance, you can rename constant.

src/cloud/ops.rs
```diff
@@ -16,8 +16,8 @@ use crate::print::{self, echo, err_marker, Highlight};
...
+const INSTANCE_OPERATION_POLLING_INTERVAL : Duration = Duration::from_secs(1);
```
tailhook, line 20:
```suggestion
const POLLING_INTERVAL : Duration = Duration::from_secs(1);
```

src/cloud/ops.rs
```diff
@@ -99,30 +105,31 @@ pub async fn find_cloud_instance_by_name(
...
+    operation: &str,
 ) -> anyhow::Result<CloudInstance> {
-    if !quiet && instance.status == "creating" {
-        print::echo!("Waiting for EdgeDB Cloud instance creation...");
+    if !quiet && instance.status == operation {
+        print::echo!("Waiting for result of EdgeDB Cloud instance", operation, "...");
     }
     let url = format!("orgs/{}/instances/{}", instance.org_slug, instance.name);
-    let deadline = Instant::now() + INSTANCE_CREATION_WAIT_TIME;
+    let deadline = Instant::now() + INSTANCE_OPERATION_WAIT_TIME;
```
tailhook, line 118: Consider using [progressbar (spinner)](https://docs.rs/indicatif/latest/indicatif/struct.ProgressBar.html#method.new_spinner) because waiting for 5 minutes is not something that users would expect.

src/cloud/ops.rs
```diff
@@ -143,6 +166,18 @@ pub async fn create_cloud_instance(
...
 
+pub async fn upgrade_cloud_instance(
+    client: &CloudClient,
+    instance: &CloudInstanceUpgrade,
+) -> anyhow::Result<()> {
+    let url = format!("orgs/{}/instances/{}", instance.org, instance.name);
+    let instance: CloudInstance = client
+        .put(url, serde_json::to_value(instance)?)
+        .await?;
+    wait_instance_upgrade(instance, client, false).await?;
```
tailhook, line 177: `quiet` is never `true` as far as I can see. Is some instance propagation option missing, or why there is a flag?

tailhook (commented):
Minor tweaks. Generally okay.

src/cloud/ops.rs
nsidnev: I'm not sure why it was added originally. Neither `create` nor `upgrade` command supports that flag, so I removed it

tailhook (approved):

PR #863, opened 2022-10-24
Continue with instance list on cloud timeout

fantix:


src/portable/status.rs
```diff
@@ -388,13 +389,24 @@ async fn get_remote_async(instances: Vec<String>) -> anyhow::Result<Vec<RemoteSt
     Ok(result)
 }
 
+async fn get_cloud(cloud_client: &CloudClient) -> anyhow::Result<Vec<RemoteStatus>> {
+    match crate::cloud::ops::list(cloud_client).await {
+        Ok(rv) => Ok(rv),
+        Err(e) if e.is::<TimeoutError>() => {
```
tailhook, line 395: I think any error in cloud code could be processed the same way. I.e. connection refused, or 500 can be the same warning.

src/portable/status.rs
```diff
@@ -388,13 +389,24 @@ async fn get_remote_async(instances: Vec<String>) -> anyhow::Result<Vec<RemoteSt
...
+            print::warn("Timed out getting cloud instances.");
```
tailhook, line 396: I would also print the error after the table. Things before the table can be unnoticed, and then users will wonder why they don't see some instances.

So the stragegy as I see it: return normal `Result` here. And carry it until the table is printed then if there was the error, print it and exit with erroneous status code.

src/portable/status.rs
tailhook: Note #869

elprans (approved):

tailhook (commented):
Mostly minor improvements. But I think read_jose_keys is a bit overcomplicated. See inline comments.

**Edit:** also looks like #886 have been merged in here. Which is probably tolerable as here is already a review history for that, let's merge it together.

src/portable/status.rs
```diff
@@ -369,13 +370,33 @@ pub fn list_local<'x>(dir: &'x Path)
...
-                .map_err(|e| log::warn!("Cannot check remote instance {:?}: {:#}", name, e))
-                .ok()
+        .map(|name| {
+            let mut err_tx = err_tx.clone();
+            async move {
+                match _remote_status(&name, false).await {
+                    Ok(status) => {
+                        if let Some(ConnectionStatus::Error(e)) = &status.connection {
+                            err_tx
+                                .send(anyhow::anyhow!("probing {}: {:#}", name, e))
```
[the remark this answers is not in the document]
fantix: Because this error was borrowed and cannot be Clone-ed.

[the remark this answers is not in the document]
tailhook: Please add a comment in the code next time you encounter something like this.

PR #867, opened 2022-10-31
Enable `nonrecursive_access_policies` by default

### LINKED ISSUE #866: Insert `using future nonrecursive_access_policies` by default into new project schemas
https://github.com/edgedb/edgedb/pull/4574 is adding a 'future' system to opt into behavior changes, and adding a new nonrecursive_access_policies to fix the behavior of access policies.

We would like to make it a "soft default" in 2.6. This means that when `edgedb project init` creates a new project (for 2.6 or later), we ought to automatically insert `using future nonrecursive_access_policies` into the starting empty schema file (along with a link to documentation we'll writing, probably).
This also requires generating an initial migration file that does a `CREATE FUTURE nonrecursive_access_policies`

tailhook:
Fixes #866

### DIFF
--- a/src/portable/project.rs
+++ b/src/portable/project.rs
@@ -496,10 +496,14 @@ pub fn init_existing(options: &Init, project_dir: &Path, cloud_options: &crate::
     let schema_dir = config.project.schema_dir;
     let schema_dir_path = project_dir.join(&schema_dir);
     let schema_dir_path =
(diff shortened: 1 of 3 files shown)


tailhook:
Updated to write a separate file.

msullivan (approved):

elprans (approved):

PR #885, opened 2022-11-17
Drop JWE in JWT

fantix:
Refs https://github.com/edgedb/edgedb/pull/4587


elprans (changes requested):
Dropping support from CLI is premature: we still want to support the current workflow for 2.x servers.

PR #888, opened 2022-11-17
Connect to Cloud instances with computed hostname

fantix:
This will currently break the cloud nightly test because:
1. certificate issues.
2. `api.g.<dns-zone>` is not resolvable.


src/cloud/client.rs
```diff
@@ -5,12 +5,14 @@ use std::io;
...
 use std::time::Duration;
 
+use anyhow::Context;
 use surf::http::auth::{AuthenticationScheme, Authorization};
 
 use crate::options::CloudOptions;
 use crate::platform::config_dir;
 
 const EDGEDB_CLOUD_BASE_URL: &str = "https://free-tier0.ovh-us-west-2.edgedb.cloud";
+const EDGEDB_CLOUD_DEFAULT_DNS_ZONE: &str = "ovh-us-west-2.edgedb.cloud";
```
elprans, line 15: These are obsolete.  The default API URL should be `https://api.g.<dns-zone>` and there is no `dns-zone` default, it's always determined from the token.

fantix: So the user should provide a `dns-zone` in `edgedb cloud login`?

elprans: Ah, good point.  Let's use `aws.edgedb.cloud` as the default for logins for now.

src/cloud/client.rs
```diff
@@ -5,12 +5,13 @@ use std::io;
...
 use std::time::Duration;
 
+use anyhow::Context;
 use surf::http::auth::{AuthenticationScheme, Authorization};
 
 use crate::options::CloudOptions;
 use crate::platform::config_dir;
 
-const EDGEDB_CLOUD_BASE_URL: &str = "https://free-tier0.ovh-us-west-2.edgedb.cloud";
+const EDGEDB_CLOUD_DEFAULT_DNS_ZONE: &str = "aws.edgedb.cloud";
```
elprans, line 14: We still want the ability to override the complete API host address with `EDGEDB_CLOUD_BASE_URL`.

elprans (approved):

PR #898, opened 2022-11-24
Abstracted out error collections for status

tailhook:
The key thing is: use borrow checker to figure out that channel is closed instead of careful checking manually and relying on runtime issue.
 
I'm also thinking that `Collector` may internally be just `(Async)Mutex<Vec<T>>`. But I'm not sure what would be better for performance. And we don't care about those tiny preformance differences here anyways.

We can also use non-async mpsc channel here and avoid awaits, as we know that recv always happens after all sends. But somehow I hesitate to do that. @fantix, what do you think?


fantix (approved):
I think it’s okay to use the sync version, because there won’t be any `await` while the channel is locked?

tailhook:
Yes, `std::sync::mpsc` is marginally more concise on both implementation and usage.

PR #900, opened 2022-11-25
Support multiple cloud profiles

### LINKED ISSUE #893: New cloud credentials config layout
Per discussion, to enable future support for auth profiles, the cloud auth secret keys should be saved as
`$CONFIG/cloud-credentials/<profile>.json`, with "default" as the default profile name.

Also, the CLI should use the secret key by default when connecting to cloud instances.

fantix:
Fixes #893


src/cloud/auth.rs
```diff
@@ -70,12 +72,36 @@ pub async fn do_login(client: &CloudClient) -> anyhow::Result<()> {
...
+            Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok(()),
+            Err(e) => anyhow::bail!(e),
+        };
+        for item in dir_entries {
+            let item = item?;
+            if let Ok(filename) = item.file_name().into_string() {
+                write_json(
+                    &cloud_creds.join(filename),
+                    "cloud config",
+                    &CloudConfig { secret_key: None },
```
tailhook, line 89: Oh, sorry this was in original code too. But why don't you just drop a file? Why writing a JSON with an empty secret key?

tailhook (approved):

tailhook:
We should probably also warn if there are linked projects and the profile is being logged out?

tailhook (changes requested):
Oh, I don't see `project init` and `instance create` linking to a profile. Am I missing something?

elprans:
> Oh, I don't see `project init` and `instance create` linking to a profile. Am I missing something?

Can we do this in a subsequent PR?

src/cloud/auth.rs
fantix: Per f2f - this was probably because we had or tried to have more than just `secret_key` in the JSON file in the past. Changed to dropping the file now.

src/portable/project.rs
```diff
@@ -1377,29 +1398,51 @@ pub fn info(options: &Info) -> anyhow::Result<()> {
...
+        .map(|projects| projects.into_values().flatten().collect())
+}
+
 #[context("could not read project dir {:?}", stash_base())]
-pub fn find_project_dirs(name: &str) -> anyhow::Result<Vec<PathBuf>> {
-    let mut res = Vec::new();
+pub fn find_project_dirs(
+    get: &str,
+    f: impl Fn(&str) -> bool,
+    noisy: bool,
```
tailhook, line 1439:
```suggestion
    verbose: bool,
```
We usually use this.

src/cloud/auth.rs
```diff
@@ -70,12 +79,157 @@ pub async fn do_login(client: &CloudClient) -> anyhow::Result<()> {
...
-        &cloud_config_file()?,
-        "cloud config",
-        &CloudConfig { access_token: None },
-    )?;
-    print::success("You're now logged out from EdgeDB Cloud.");
-    Ok(())
+fn find_real_project_dirs(
+    f: impl Fn(&str) -> bool,
+) -> anyhow::Result<HashMap<String, Vec<PathBuf>>> {
+    find_project_dirs("cloud-profile", f, false).map(|projects| {
```
tailhook, line 85:
```suggestion
    let projects = find_project_dirs("cloud-profile", f, false)?;
```
No need for map and nesting here.

tailhook, line 85:
```suggestion
    let projects = find_project_dirs("cloud-profile", f, false)?;
```
No need for map and nesting here.

src/portable/project.rs
```diff
@@ -1377,29 +1398,51 @@ pub fn info(options: &Info) -> anyhow::Result<()> {
...
 
+pub fn find_project_dirs_by_instance(name: &str) -> anyhow::Result<Vec<PathBuf>> {
+    find_project_dirs("instance-name", |val| name == val, true)
+        .map(|projects| projects.into_values().flatten().collect())
+}
+
 #[context("could not read project dir {:?}", stash_base())]
-pub fn find_project_dirs(name: &str) -> anyhow::Result<Vec<PathBuf>> {
-    let mut res = Vec::new();
+pub fn find_project_dirs(
```
tailhook, line 1436:
```suggestion
pub fn find_project_stash_dirs(
```
We already use "stash dir" terminology everywhere. So it makes sense to have "stash dir" vs "project dir" (instead of "project dir" which can be two things vs "real project dir").

tailhook, line 1436:
```suggestion
pub fn find_project_stash_dirs(
```
We already use "stash dir" terminology everywhere. So it makes sense to have "stash dir" vs "project dir" (instead of "project dir" which can be two things vs "real project dir").

tailhook (approved):
Few nits, but generally good. The secret key hack looks a bit unfortunate, but I'm not sure how to fix it yet.

PR #902, opened 2022-12-01
Parse query in DSN

fantix:
- [ ] Add shared test cases

### DIFF
--- a/edgedb-client/src/builder.rs
+++ b/edgedb-client/src/builder.rs
@@ -44,6 +44,7 @@ use crate::errors::{ClientConnectionFailedError, AuthenticationError};
 use crate::errors::{ClientError, ClientConnectionFailedTemporarilyError};
 use crate::errors::{ClientNoCredentialsError, ProtocolEncodingError};
 use crate::errors::{Error, ErrorKind, PasswordRequired, ResultExt};
+use crate::errors::InterfaceError;
 use crate::server_params::{PostgresAddress, SystemConfig};
 use crate::tls;
 
@@ -112,6 +113,12 @@ pub(crate) enum Address {
 }
 
 struct DisplayAddr<'a>(Option<&'a Address>);
+
+struct DSN {
+    url: url::Url,
+    admin: bool,
+    query: HashMap<String, String>,
+}
 
 pub async fn timeout<F, T>(dur: Duration, f: F) -> Result<T, Error>
     where F: Future<Output = Result<T, Error>>,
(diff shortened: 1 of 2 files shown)


nsidnev (approved):

fantix:
I'll merge this one first, please feel free to leave further comments. I'll address them in a new PR with adding the shared test cases.

edgedb-client/src/builder.rs
```diff
@@ -322,6 +329,171 @@ impl fmt::Display for DisplayAddr<'_> {
     }
 }
 
+impl DSN {
+    fn new(dsn: &str) -> Result<Self, Error> {
```
tailhook, line 333: This could be `FromStr for Dsn`. But since the type is not public, could be just `from_str`.

edgedb-client/src/builder.rs
```diff
@@ -113,6 +114,12 @@ pub(crate) enum Address {
 
 struct DisplayAddr<'a>(Option<&'a Address>);
 
+struct DSN {
```
tailhook, line 117: We treat abbreviations as words in Rust. So:
```suggestion
struct Dsn {
```

This makes it easier to name variables and constants with that type.

edgedb-client/src/builder.rs
```diff
@@ -322,6 +329,171 @@ impl fmt::Display for DisplayAddr<'_> {
...
+    async fn get_value<T>(
+        &mut self,
+        key: &'static str,
+        mut v: Option<T>,
+        conv: impl FnOnce(String) -> Result<T, Error>,
+        default: impl FnOnce() -> T,
+    ) -> Result<T, Error> {
+        let v_query = self.query.remove(key);
+        let k_env = format!("{key}_env");
+        let v_env = self.query.remove(k_env.as_str());
```
tailhook, line 361:
```suggestion
        let v_env = self.query.remove(&k_env);
```
.. should be enough. `as_str` makes impression that you converting types here, which you actually don't do.

edgedb-client/src/builder.rs
```diff
@@ -322,6 +329,171 @@ impl fmt::Display for DisplayAddr<'_> {
...
+        }
+        if v.is_none() {
+            if let Some(rv) = rv {
+                let rv = conv(rv).with_context(|| {
+                    format!("failed to parse value of query parameter {}", key_used)
+                })?;
+                v = Some(rv);
+            }
+        }
+        Ok(v.unwrap_or_else(default))
```
tailhook, line 410: Honestly, this logic is hard to follow. And I'm not sure that this is inherent complexity.

What is the difference between `v` and `rv`? Why are you mutating both?

Logic here should be either:
```rust
let result = /* less priority value */;
if /* cond 1 */ {
   result = /* next priority */;
}
if /* cond 2 */ {
   result = /* next priority */;
}
Ok(result)
```
Or vice versa:
```result
let result = /* most priority value*/;
if result.is_none() && /* cond 1 */ {
    result = Some(...);
}
if result.is_none() && /* cond 2 */ {
    result = Some(...);
}
Ok(result.unwrap_or(/*default*/))
```
Usually the second style is used when some things are computationally intensive.

tailhook: You can also check for redundant query parameters at the start of the function and don't intermix it with the parsing logic, I think.

tailhook: Also can redundant parameters be checked in the loop at the start? I.e. any `xxx`, `xxx_env` and `xxx_file` are redundant, right? And then you can only check redundancy with the url's own value (like `host`) in this function.

But maybe moving checks to the start of `get_value` is also fine.

edgedb-client/src/builder.rs
```diff
@@ -322,6 +329,171 @@ impl fmt::Display for DisplayAddr<'_> {
...
+                "String {:?} is not a valid DSN",
+                dsn,
+            )));
+        };
+        let url = url::Url::parse(dsn).map_err(|e| {
+            ClientError::with_source(e).context(format!("cannot parse DSN {:?}", dsn))
+        })?;
+        let query = url
+            .query_pairs()
+            .into_owned()
```
tailhook, line 346: I think you could do this without making them owned. Just chosen a bit wrong set of abstractions here. Kinda this:
```
let url = Url::parse(..)?;
let query = url.query_pairs().collect::<HashMap<_, _>();
let host = get_value(url, query, "host", ...);
let port = get_value(url, query, "port", ...);
```
I.e. unpack values while URL is still on stack.

This can be further wrapped into a structure with a lifetime:
```
struct DsnHelper<'a> {
    url: &'a Url,
    query: &'a HashMap<..>,
}
let url = Url::parse(..)?;
let dsn = DsnHelper::from_url(url)?;
let host = dsn.get_host()?;
let port = dsn.get_port()?;
```

But it's unlikely that this will be a performance issue, so it's very minor thing here.

edgedb-client/src/builder.rs
```diff
@@ -322,6 +329,171 @@ impl fmt::Display for DisplayAddr<'_> {
...
+        Ok(Self { url, admin, query })
+    }
+
+    async fn get_value<T>(
+        &mut self,
+        key: &'static str,
+        mut v: Option<T>,
+        conv: impl FnOnce(String) -> Result<T, Error>,
+        default: impl FnOnce() -> T,
+    ) -> Result<T, Error> {
```
tailhook, line 358: Calls to this function are hard to read. Four args is not too many usually. But two lambdas and an Option and sometimes `T` is also an option, is hard to read. And also difference between `v` and `default` is a bit subtle.

I think that it would be easier if part of that is done in `retrieve_xx` (or rather `get_x`). Even if some logic is duplicated (which is `if Some(...)` and `.unwrap_or` I think).  I also think this is easier to do if conflict logic is separate from parsing logic.

tailhook (commented):
Not sure why you use `get_value`, but `retrieve_*`.

But generally I think parsing logic is a bit less clear than it should be so I have hard time to review it on correctness. Unless I'm misunderstanding some of the constraints I think it can be simplified. See my inline comments.

fantix:
Thanks @tailhook ! Addressed in the first 3 commits in #905

PR #905, opened 2022-12-02
Add shared test cases

fantix:
* CRF for #902 
* Store and use resolved `tls_security`
* Add error context for DSN parsing
* Added `SkipFields` to `parse_dsn()`
* DSN path `/` is None, not empty name
* Supported EDGEDB_CLIENT_SECURITY=strict
* EDGEDB_CLIENT_SECURITY=insecure_dev_mode is now only effective when EDGEDB_CLIENT_TLS_SECURITY=default
* Error on empty host, user and database
* Prevent DSN query parameters with the same name
* Strip leading `/` in ?database=/xxx
* Error out with multiple compound env vars
* Only take EDGEDB_WAIT_UNTIL_AVAILABLE on the env layer
* Reject multiple hosts
* Reject port=0
* Support connection parameters in DSN query
* [REPL will now try to find `edgedb.toml` in parent dirs](https://github.com/edgedb/edgedb-cli/pull/905/commits/227722781bc47c951ae658c6eee7252ce023820f)
* Fix credentials file legacy checking bug

### DIFF
--- a/edgedb-client/src/builder.rs
+++ b/edgedb-client/src/builder.rs
@@ -1093,15 +1093,31 @@ impl Builder {
                 Run `edgedb project init` or use environment variables \
                 to configure connection."));
         }
+        let tls_security;
         let verifier = match self.tls_security {
-            _ if self.insecure() => Arc::new(tls::NullVerifier) as Verifier,
-            Insecure => Arc::new(tls::NullVerifier) as Verifier,
+            _ if self.insecure() => {
+                match self.tls_security {
+                    Default => {
+                        tls_security = Insecure;
+                    }
+                    ts => {
+                        tls_security = ts;
+                    }
+                }
+                Arc::new(tls::NullVerifier) as Verifier
+            },
+            Insecure => {
+                tls_security = self.tls_security;
+                Arc::new(tls::NullVerifier) as Verifier
+            },
             NoHostVerification => {
+                tls_security = self.tls_security;
                 Arc::new(tls::NoHostnameVerifier::new(
                         self.trust_anchors()?
                 )) as Verifier
             }
             Strict => {
+                tls_security = self.tls_security;
                 Arc::new(rustls::client::WebPkiVerifier::new(
                     self._root_cert_store()?,
                     None,
@@ -1109,11 +1125,13 @@ impl Builder {
             }
             Default => match self.pem {
                 Some(_) => {
+                    tls_security = NoHostVerification;
                     Arc::new(tls::NoHostnameVerifier::new(
                             self.trust_anchors()?
                     )) as Verifier
                 }
                 None => {
+                    tls_security = Strict;
                     Arc::new(rustls::client::WebPkiVerifier::new(
                         self._root_cert_store()?,
                         None,
@@ -1133,7 +1151,7 @@ impl Builder {
             instance_name: self.instance_name.clone(),
             wait: self.wait,
             connect_timeout: self.connect_timeout,
-            tls_security: self.tls_security,
+            tls_security,
             insecure_dev_mode: self.insecure_dev_mode,
 
             // Pool configuration
@@ -1281,19 +1299,6 @@ impl Config {
         Ok(conn)
     }
 
-
-    fn do_verify_hostname(&self) -> Option<bool> {
-        use TlsSecurity::*;
-        if self.0.insecure_dev_mode {
-            return Some(false);
-        }
-        match self.0.tls_security {
-            Insecure => Some(false),
-            NoHostVerification => Some(false),
-            Strict => Some(true),
-            Default => None,
-        }
-    }
     /// Return a single connection.
     #[cfg(feature="unstable")]
     pub async fn connect(&self) -> Result<Connection, Error> {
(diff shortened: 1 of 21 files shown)


tailhook:
LGTM so far

edgedb-client/src/builder.rs
```diff
@@ -329,111 +350,96 @@ impl fmt::Display for DisplayAddr<'_> {
...
+        }
         Ok(Self { url, admin, query })
     }
 
-    async fn get_value<T>(
+    async fn retrieve_value<T>(
         &mut self,
         key: &'static str,
-        mut v: Option<T>,
+        v: Option<T>,
```
tailhook, line 372:
```suggestion
        v_from_url: Option<T>,
```
Since it's used twice there is no need to shorten it. And it's very unclear what is this value when reading function signature.

edgedb-client/src/builder.rs
```diff
@@ -1110,27 +1212,60 @@ impl Builder {
...
+        }
+        if self.get_user().is_empty() {
+            return Err(ClientError::with_message("invalid user: empty string"));
+        }
+        if self.get_database().is_empty() {
+            return Err(ClientError::with_message("invalid database: empty string"));
+        }
+        if self.get_port() == 0 {
+            return Err(ClientError::with_message("invalid port: 0"));
+        }
```
tailhook, line 1229: I think it's better to move these to appropriate setters? We don't have error result in `set_host`/`set_port`/`set_database`, but should probably have. DSN can already fail so these checks there will be just fine.

Also these are probably `ClientInterfaceError`. I.e. invalid usages by user, right?

edgedb-client/src/builder.rs
```diff
@@ -1110,27 +1212,60 @@ impl Builder {
...
+        let tls_security;
         let verifier = match self.tls_security {
-            _ if self.insecure() => Arc::new(tls::NullVerifier) as Verifier,
-            Insecure => Arc::new(tls::NullVerifier) as Verifier,
+            _ if self.insecure() => {
+                match self.tls_security {
```
tailhook, line 1233: why do you put match of the same thing inside the match? Please move these braches to the outer match.

tailhook (commented):
Minor tweaks are inline.

But generally I don't like `read_dsn(...SkipFields)` public API. What is the main reason? Is it to skip invalid parts of DSN if they are overridden by the env? I don't think accepting invalid DSNs is a great idea. Do other language bindings deal with this thing same way?

fantix:
> But generally I don't like `read_dsn(...SkipFields)` public API. What is the main reason? Is it to skip invalid parts of DSN if they are overridden by the env? I don't think accepting invalid DSNs is a great idea. Do other language bindings deal with this thing same way?

Yeah it is sort of "skip invalid parts of DSN", or more like "skip parsing those parts in DSN" if they are already set in the same layer/group. Other bindings are doing this in a reversed way as CLI does:

1. For each layer/group (args / envs / proj), set non-compound values first;
2. Set compound values and stop if exist;
3. If the compound value is a DSN, skip the non-compound values that are already set ([Python](https://github.com/edgedb/edgedb-python/blob/241c80d84d9a1b47833020cdff8ea7da97a64787/edgedb/con_utils.py#L718), [JS](https://github.com/edgedb/edgedb-js/blob/70e4d806472e5c6771ba26461c13258072f85868/packages/driver/src/conUtils.ts#L835) and [Go](https://github.com/edgedb/edgedb-go/blob/49aa10128cb08f9ca9d9ae48a3d866af1a94e123/internal/client/connutils.go#L1068-L1070)).

While the CLI was like:

1. Find the layer/group with compound values first by priority;
2. Set compound values in that layer; if DSN, all values will be parsed and set;
3. Loop back to the highest priority layer and overwrite with non-compound values from each layer.

Considering you're also working on porting the CLI to use `edgedb-rust`, I added this less-intrusive `Skip` hack in step 2 to pass the shared test suite, without having to rewrite the whole logic to follow the bindings. OR we can also fix the bindings and the test cases to always parse and verify all given values 🤷‍♂️

tailhook (approved):
Okay, let's keep `read_dsn` as is. I have to think on the API. I'll be able to do that as part of my work.

Please fix all small suggestions and tests, and let's merge it in.

edgedb-client/src/builder.rs
fantix: > I think it's better to move these to appropriate setters?

I'm not 100% sure if this is right for a "builder", but it's done now.

> Also these are probably `ClientInterfaceError`. I.e. invalid usages by user, right?

Sure, `InvalidArgumentError` I guess.

edgedb-client/src/builder.rs
fantix: After reviewing this, I think there was an inconsistancy issue in `insecure_dev_mode`. Fixed [here](https://github.com/edgedb/edgedb-cli/pull/905/commits/ec7c1af6fc339e40fdfb2f42940541b36779bd6f), please kindly double check.

tailhook: Looks good. Thanks!

tailhook:
Okay. Looks good. Do you have any ideas what's wrong with the test?

fantix:
There is definitely some panic in the `assert_cmd` piped stdout/err reading code that we don't have a traceback about. So I simply replaced it with a `std::process::Command` which seems to work fine.

PR #924, opened 2023-01-23
Use edgedb-tokio for talking to server

tailhook:


tailhook:
@fantix, might be quite a big change to review in one chunk. But, please, take a closer look to cloud stuff. And please try running it if you can, because currently its untested mostly.

src/portable/create.rs
```diff
@@ -24,10 +23,11 @@ use crate::print::{self, echo, err_marker, Highlight};
 use crate::process;
 use crate::question;
 
-use edgedb_client::credentials::Credentials;
+use edgedb_tokio::credentials::Credentials;
 
 
-fn ask_name(
+#[tokio::main]
+async fn ask_name(
```
fantix, line 30: I think this function doesn't have to be async now.

src/portable/project.rs
```diff
@@ -316,7 +315,8 @@ pub fn init(options: &Init, opts: &crate::options::Options) -> anyhow::Result<()
     Ok(())
 }
 
-fn ask_existing_instance_name(
+#[tokio::main]
+async fn ask_existing_instance_name(
```
fantix, line 319: Ditto, async not needed.

src/portable/project.rs
```diff
@@ -419,7 +419,8 @@ fn do_link(inst: &Handle, options: &Init, stash_dir: &Path)
     })
 }
 
-fn ask_name(
+#[tokio::main]
+async fn ask_name(
```
fantix, line 423: Ditto, async not required.

fantix (changes requested):
Some cloud operations are crashing because of creating runtime twice in some places, please see inline.

fantix (approved):

PR #926, opened 2023-01-24
Add support for `_EDGEDB_CLOUD_CERTS=local`

elprans:
Local virtual Nebula deploys use a self-signed root certificate.  Trust
it when `_EDGEDB_CLOUD_CERTS` is set to `local`.


fantix (approved):

fantix:
We can use a `match`, I'll copy and apply the changes over to https://github.com/edgedb/edgedb-rust/pull/194.

PR #929, opened 2023-01-25
client: Adjust cloud instance hostname derivation

elprans:


fantix (approved):
Fix is good, test failing on the shared test cases I believe, but I can follow up with a fix later if you don’t mind having a broken CLI master to unblock the cloud work.

PR #940, opened 2023-02-13
watch: Use `force_database_error`

tailhook:


fantix (approved):
* Should I restart `edgedb watch` after fixing issues in sdl files? Looks like the error is not cleared.
* Looks like the `error when trying to update the schema` is only triggered when compiling new queries (cache miss), is that expected?

tailhook:
> * Should I restart `edgedb watch` after fixing issues in sdl files? Looks like the error is not cleared.
> 

Fixed. Also renamed `with_ignore_erorr_state` -> `set_` method to make it more idiomatic.

>     * Looks like the `error when trying to update the schema` is only triggered when compiling new queries (cache miss), is that expected?

This is a bug of EdgeDB https://github.com/edgedb/edgedb/issues/5038

PR #949, opened 2023-02-23
interactive: Properly flush writes

### LINKED ISSUE #933: tokio CLI sometimes displays new prompt before output
```
edgedb> insert Thing { name := 'Alice2' };
edgedb> [{"id": "427806d0-a748-11ed-9f6f-ff8e23ce8888"}]
```

tailhook:
Fixes #933

### DIFF
--- a/src/interactive.rs
+++ b/src/interactive.rs
@@ -252,6 +252,13 @@ async fn execute_backslash(mut state: &mut repl::State, text: &str)
     Ok(())
 }
 
+async fn write_out(data: &str) -> anyhow::Result<()> {
+    let mut out = stdout();
+    out.write_all(data.as_bytes()).await?;
+    out.flush().await?;
+    Ok(())
+}
+
 async fn execute_query(options: &Options, state: &mut repl::State,
     statement: &str)
     -> anyhow::Result<()>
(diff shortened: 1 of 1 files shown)


src/interactive.rs
```diff
@@ -372,8 +379,7 @@ async fn execute_query(options: &Options, state: &mut repl::State,
                 };
                 // trying to make writes atomic if possible
                 text += "\n";
-                stdout().write_all(text.as_bytes()).await?;
-                stdout().flush().await?;
+                write_out(tex).await?;
```
msullivan, line 382: text

msullivan (approved):

msullivan:
I'm taking the liberty of merging this despite the CI failure, because this bug is driving me crazy and I really want the fix :P

I think that this is basically OK, because the test that is failing is test-bin-installable, which tests a `cargo install` of the CLI. Per the docs that I just looked up, cargo install ignores `Cargo.lock` and recomputes dependencies (unless you specify `--locked`). So the problem here is that a new token has been added to edgeql but the CLI doesn't support it yet, which doesn't have anything to do with this issue.

Maybe it ought to specify `--locked`, or maybe it should be a nightly job or specifically marked as non-critical?

msullivan:
Confirming that this does appear to have fixed it! Thanks!

PR #959, opened 2023-02-27
Add secret key in `instance credentials` result

### LINKED ISSUE #917: secret_key should be included in instance credentials output
Currently (EdgeDB CLI 2.3.0-dev.902+e29ccdb) `secret_key` doesn't appear in the output from `edgedb instance credentials` but it probably should.

fantix:
Fixes #917

### DIFF
--- a/src/portable/credentials.rs
+++ b/src/portable/credentials.rs
@@ -4,14 +4,27 @@ use url::Url;
 use crate::options::Options;
 use crate::portable::options::ShowCredentials;
 
+#[derive(serde::Serialize)]
+struct Credentials {
+    #[serde(skip_serializing_if="Option::is_none")]
+    secret_key: Option<String>,
+
+    #[serde(flatten)]
+    creds: edgedb_tokio::credentials::Credentials,
+}
+
 pub fn show_credentials(options: &Options, c: &ShowCredentials) -> anyhow::Result<()> {
     use edgedb_tokio::credentials::TlsSecurity;
 
     let connector = options.block_on_create_connector()?;
-    let builder = connector.get()?;
+    let builder: &edgedb_tokio::Builder = connector.get()?;
     let creds = builder.as_credentials()?;
+    let secret_key = builder.get_secret_key().map(String::from);
     if let Some(result) = if c.json {
-        Some(serde_json::to_string_pretty(&creds)?)
+        Some(serde_json::to_string_pretty(&Credentials {
+            secret_key,
+            creds,
+        })?)
     } else if c.insecure_dsn {
         let mut url = Url::parse(&format!(
             "edgedb://{}@{}:{}",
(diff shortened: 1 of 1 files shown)


fmoor (approved):

src/portable/credentials.rs
```diff
@@ -4,14 +4,27 @@ use url::Url;
 use crate::options::Options;
 use crate::portable::options::ShowCredentials;
 
+#[derive(serde::Serialize)]
+struct Credentials {
+    #[serde(skip_serializing_if="Option::is_none")]
+    secret_key: Option<String>,
```
tailhook, line 10: This is something weird. Either we support secret_key in credentials JSON. Or we don't output it here. I think JSON output here should be readable as a credentials file.

src/portable/credentials.rs
```diff
@@ -42,6 +62,7 @@ pub fn show_credentials(options: &Options, c: &ShowCredentials) -> anyhow::Resul
             ("Port", creds.port.to_string().as_str()),
             ("User", creds.user.as_str()),
             ("Password", creds.password.map(|_| "<hidden>").unwrap_or("<none>")),
+            ("Secret Key", secret_key.map(|_| "<hidden>").unwrap_or("<none>")),
```
tailhook, line 65: Can we show some basic parameters of the token (i.e. `<edgedb.cloud AWS token>`)? Or are all of them are encrypted?

src/portable/credentials.rs
```diff
@@ -23,17 +36,24 @@ pub fn show_credentials(options: &Options, c: &ShowCredentials) -> anyhow::Resul
...
+            ),
+            ("secret_key", secret_key),
+        ]
+        .iter()
+        .map(|(k, v)| v.as_deref().map(|v| format!("{}={}", k, v)))
+        .flatten()
+        .collect::<Vec<_>>();
+        if !query.is_empty() {
+            let query = query.join("&");
+            url.set_query(Some(&query));
```
tailhook, line 56: Please use `url.query_pairs_mut().append_pair()`. Technially your code is wrong if secret key contains anything that has to be urlencoded.

src/portable/credentials.rs
```diff
@@ -23,17 +36,24 @@ pub fn show_credentials(options: &Options, c: &ShowCredentials) -> anyhow::Resul
...
+                } else {
+                    Some(serde_json::to_value(&creds.tls_security).unwrap().as_str().unwrap().into())
+                },
+            ),
+            ("secret_key", secret_key),
+        ]
+        .iter()
+        .map(|(k, v)| v.as_deref().map(|v| format!("{}={}", k, v)))
+        .flatten()
+        .collect::<Vec<_>>();
```
tailhook, line 53: I use same indentation level for dot-something only if that is the only thing in that code block and still for short blocks. Otherwise it hinders readability.

src/portable/credentials.rs
```diff
@@ -23,17 +36,24 @@ pub fn show_credentials(options: &Options, c: &ShowCredentials) -> anyhow::Resul
...
+        let query = [
+            (
+                "tls_security",
+                if creds.tls_security == TlsSecurity::Default {
+                    None
+                } else {
+                    Some(serde_json::to_value(&creds.tls_security).unwrap().as_str().unwrap().into())
+                },
+            ),
+            ("secret_key", secret_key),
```
tailhook, line 48: Please use variables, they won't bite you. And this:
```
("tls_security", tls_security),
("secret_key", secret_key),
```
reads so much better.

tailhook (commented):
We have to decide whether we support `secret_key` in json, or don't output it here. This definitely has to be fixed one way or another. Reopening #917 for that.

Also some style issues.

src/portable/credentials.rs
fantix: Yeah we can extract the token issuer (domain) and show that

PR #969, opened 2023-03-08
Use new Builder API from edgedb-rust

tailhook:


src/portable/status.rs
```diff
@@ -250,9 +250,10 @@ fn normal_status(options: &Status) -> anyhow::Result<()> {
 }
 
 async fn try_get_version(creds: &Credentials) -> anyhow::Result<String> {
-    let mut builder = Builder::uninitialized();
-    builder.credentials(creds)?;
-    let mut conn = Connection::connect(&builder.build()?).await?;
+    let config = Builder::new()
+        .credentials(creds)?
+        .constrained_build()?;
```
fantix, line 255: Seems like `constrained_build()` would raise an error "unsupported constraint builder param" because `credentials` is set.

fantix (approved):
LGTM!

src/portable/status.rs
tailhook: Nice catch. Fixed.

PR #982, opened 2023-03-21
Initial implementation for `edgedb cloud secretkey` command group

elprans:
Here:

`secretkey list` -- lists existing secret keys
`secretkey create` -- creates a new secret key
`secretkey revoke` -- revokes a secret key

### DIFF
--- a/src/cloud/main.rs
+++ b/src/cloud/main.rs
@@ -1,7 +1,7 @@
 use crate::options::CloudOptions;
 use crate::cloud::options::CloudCommand;
 use crate::cloud::auth;
-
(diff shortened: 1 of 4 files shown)


fantix (approved):
Looks good and works fine for me.

(The issued token is not prefixed with `nbwt_`, but I think that's a cloud issue).

tailhook (approved):

elprans:
> would a user generally want the keys to live longer?

It depends on the use case.  In automation, the validity period can be even smaller.  For interactive sessions an hour is also often sufficient, e.g. when a developer needs to debug a production instance from their machine or something like that.

PR #985, opened 2023-03-27
Fix `edgedb ui` with prefixed tokens

fantix:
Backport the UI command fix 26b244cf883f5b07da78241d440671cbff32d356  for 3.x instances

### DIFF
--- a/src/commands/ui.rs
+++ b/src/commands/ui.rs
@@ -1,16 +1,8 @@
-use std::env;
 use std::io::{stdout, Write};
-use std::path::PathBuf;
-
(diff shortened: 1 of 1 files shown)


tailhook (commented):
I'm not against this. But maybe we should advise using `testing` channel for beta cloud users?

fantix:
Oh sure - this PR is more for the upcoming 3.0 server, in case people didn't upgrade their CLI.

fantix:
No longer needed as 3.x already hit the stable channel.

PR #1000, opened 2023-04-13
Squashing migrations

### LINKED ISSUE #976: Tool for squashing migration history for problematic upgrades
We're going to need a tool that uses migration rewrites to squash the migration history, in order to enable recovering from upgrades that make it impossible the apply the existing migrations to a fresh database.

The upgrade workflow I'm picturing is something like:
1. Try running the new version in your dev environment, applying all the migrations
2. If something breaks, fix it. Use rewrites to squash the migration history.
3. Once you can create a new dev database, test out dump/restore
4. Maybe do a backup dump to be safe?
5. Deploy any changes and migration squashes to prod
6. Prod dump
7. Upgrade prod version
8. Prod restore

tailhook:
Fixes #976 as discussed in edgedb/edgedb#5193


src/migrations/squash.rs
```diff
@@ -17,29 +38,92 @@ pub async fn main(cli: &mut Connection, options: &Options,
...
+    let key = MigrationKey::Fixup {
+        target_revision: squashed.id()?.to_owned(),
+    };
+    let fixup = if needs_fixup {
+        if create.non_interactive || cli.ping_while(want_fixup()).await? {
+            let parent = Some(&db_rev[..]);
+            Some(normal_migration(cli, &ctx, key, parent, create).await?)
+        } else {
+            None
+        }
```
fantix, line 69: If the user refused the fixup to include latest schema changes, we should at least create an empty fixup for the squash itself?

fantix (approved):
Cool!

PR #1004, opened 2023-04-25
analyze: add json output and input

tailhook:


src/analyze/mod.rs
```diff
@@ -60,28 +64,53 @@ pub fn render_explain(explain: &Analysis) -> anyhow::Result<()>
...
+        cli.query_required_single::<String, _>(&query, &()).await?
+    };
+    if let Some(out_path) = &options.debug_output_file {
+        if out_path == Path::new("-") {
+            let mut out = io::stdout();
+            out.write_all(data.as_bytes()).await?;
+            out.flush().await?;
+        } else if out_path.starts_with("/dev") {
+            async {
+                let mut out = fs::File::create(&out_path).await?;
```
fantix, line 92: This is for `/dev/stdout`?

fantix (approved):

src/analyze/mod.rs
fantix: Did you meant to write `} else if !out_path.starts_with("/dev") {` with the `not`?

PR #1008, opened 2023-05-01
cloud: Use cloud API to obtain the list of available server versions

elprans:
Also, fix display of cloud instance versions in `instance list`.

### DIFF
--- a/src/cloud/ops.rs
+++ b/src/cloud/ops.rs
@@ -47,14 +47,14 @@ impl RemoteStatus {
         secret_key: &str,
     ) -> anyhow::Result<Self> {
         let credentials = cloud_instance.as_credentials(secret_key).await?;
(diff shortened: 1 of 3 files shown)


src/cloud/ops.rs
```diff
@@ -121,6 +126,19 @@ pub async fn get_current_region(
...
+pub async fn get_versions(
+    client: &CloudClient,
+) -> anyhow::Result<Vec<Version>> {
+    let url = "versions";
+    client
+        .get(url)
+        .await
+        .or_else(|e| match e.downcast_ref::<ErrorResponse>() {
+            _ => Err(e),
+        })
```
tailhook, line 139: This does nothing?

tailhook (approved):

PR #1009, opened 2023-05-01
Normalize output of `instance list`

elprans:
The table produced by `instance list` now includes a `Location` column
instead of `port`, and cloud instances are shown as
`EdgeDB.Cloud @ <region>` instead of the underlying host name.

### DIFF
--- a/src/cloud/ops.rs
+++ b/src/cloud/ops.rs
@@ -24,6 +24,7 @@ pub struct CloudInstance {
     dsn: String,
     status: String,
     pub version: String,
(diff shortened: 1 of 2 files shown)


tailhook (approved):

tailhook:
My only concern is whether `EdgeDB.Cloud @` is too long (for users with small screens like laptops, for example). With modern trends I'd use `☁️ us-east-2` (I mean the cloud emoji, and our own cloud is certainly the first class citizen here).

PR #1010, opened 2023-05-02
Print credentials path if found

Dhghomon:
I've noticed that the UI in version 3.0 first goes to a login screen (well, after telling the browser that the address can be trusted) and took me a bit of time to find my credentials when I first encountered the login screen.


tailhook:
I think we should suggest a command. Currently it's:
```
edgedb instance credentials -I <name> --json
```
Which is quite bad (json format). We had a discussion on how to improve that, but nothing was implemented.

Dhghomon:
Hm, maybe the command could just be suggested on the UI login screen then.
That alone should do the trick.

2023년 5월 2일 (화) 오후 6:52, Paul Colomiets ***@***.***>님이 작성:

PR #1011, opened 2023-05-02
tests: ignore underscore properties in configure tests

tailhook:


tailhook:
Although, @fantix, maybe we want to the config param `cfg::internal = true` ? In addition to underscore or instead of making it underscored?

elprans (approved):

PR #1021, opened 2023-05-10
watch: Unhide the command to show in help

tailhook:


src/options.rs
```diff
@@ -319,7 +319,7 @@ pub enum Command {
     #[edb(inherit(CloudOptions), hide=true)]
     Cloud(CloudCommand),
     /// EdgeDB Cloud authentication
```
elprans, line 321: The doc and the inheritance of `CloudOptions` have probably been copy-pasted by mistake?

tailhook: Right. Fixed that.

elprans (approved):

Issue #973, opened 2023-03-13
Display something useful for `EXPLAIN` queries 


tailhook:
Depends a little on https://github.com/edgedb/edgedb/issues/5162

PR #1027, opened 2023-05-11
Nicer analyze output

tailhook:
This completes non-expanded explain output.

screenshot

The screenshot also uses fix of edgedb/edgedb#5467

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -854,6 +854,7 @@ dependencies = [
  "base32",
  "base64 0.13.1",
  "bigdecimal",
+ "bitvec",
  "blake2b_simd",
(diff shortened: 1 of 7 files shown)


tailhook:
@elprans please include a squashed version of this if you're going to release the CLI.

tailhook:
Fixes #973

fantix (approved):
I didn't look into the code super carefully, but it works pretty well for me (and looks cool!). I did notice one thing which you may be aware of already - when the query has parameters, it would raise an error like this:

```
edgedb error: DescriptorMismatch: query arguments expected
```

instead of asking for values like regular queries do.

PR #1037, opened 2023-05-16
restore: add usage example with the database name

### LINKED ISSUE #1025: Document `-d` option for `edgedb restore` in help

tailhook:
Also removes `--credentials-file` from the shortlist of connection options.

Fixes #1025

New usage is:
```
Restore a database backup from file

USAGE:
    edgedb restore [OPTIONS] <path>
    edgedb restore -d <database-name> <path>

ARGS:
    <path>    Path to file (or directory in case of --all) to read dump from. Use dash
              - to read from stdin

OPTIONS:
        --all        Restore all databases and the server configuration. path is a
                     directory in this case
    -v, --verbose    Verbose output
    -h, --help       Print help information

CONNECTION OPTIONS (`edgedb --help-connect` to see the full list):
    -I, --instance <instance>    Local instance name created with edgedb instance create
                                 to connect to (overrides host and port)
        --dsn <dsn>              DSN for EdgeDB to connect to (overrides all other options
                                 except password)
```

### DIFF
--- a/src/commands/parser.rs
+++ b/src/commands/parser.rs
@@ -349,6 +349,10 @@ pub struct Dump {
 }
 
 #[derive(EdbClap, Clone, Debug)]
+#[clap(override_usage(
+    "edgedb restore [OPTIONS] <path>\n    \
+     edgedb restore -d <database-name> <path>"
+))]
 pub struct Restore {
     /// Path to file (or directory in case of `--all`) to read dump from.
     /// Use dash `-` to read from stdin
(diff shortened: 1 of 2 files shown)


tailhook:
I'm not sure if this use case is visible enough. But bringing `-d` to `OPTIONS` is not possible within normal `clap` usage. And bringing it into `CONNECTION OPTIONS` for every command might be suboptimal (because -I and --dsn there are basically full-spec credentials, and someone might think that `-d` is the only thing that is needed for connection.

PR #1038, opened 2023-05-16
Fix Cloud command help

raddevon:
Updated these per @fantix's comments on https://github.com/edgedb/edgedb/pull/5487


fantix (approved):
Thanks!

PR #1039, opened 2023-05-16
cloud: Use version API for upgrade version filtering

elprans:
Also, bring back support for testing and nightly channels.

### DIFF
--- a/src/portable/options.rs
+++ b/src/portable/options.rs
@@ -177,7 +177,7 @@ pub struct Create {
     #[clap(long, default_value="edgedb")]
     pub default_user: String,
 
-    /// Do not ask questions, assume user wants to create instance
+    /// Do not ask questions, assume user wants to upgrade instance
     #[clap(long)]
     pub non_interactive: bool,
 }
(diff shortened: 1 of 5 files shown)


src/portable/options.rs
```diff
@@ -457,6 +457,10 @@ pub struct Upgrade {
     /// This is used by `project upgrade --force`
     #[clap(long, hide=true)]
     pub force_dump_restore: bool,
+
+    /// Do not ask questions, assume user wants to create instance
```
zackelan, line 461:
```suggestion
    /// Do not ask questions, assume user wants to upgrade instance
```

src/portable/upgrade.rs
```diff
@@ -147,41 +148,67 @@ fn upgrade_cloud(cmd: &Upgrade, org: &str, name: &str, opts: &crate::options::Op
...
+        version: target_ver.to_string(),
     };
     cloud::ops::upgrade_cloud_instance(&client, &request)?;
 
     print::echo!(
-        "EdgeDB Cloud instance",
-        format!("{}/{}", org, name).emphasize(),
-        "is successfully upgraded.",
+        "EdgeDB Cloud instance {inst_name} has been successfully \
+        upgraded to version {target_ver_str}.",
```
nsidnev, line 211: the `format!` call is missed here

elprans: I keep forgetting that our custom output macros don't support format strings.

PR #1040, opened 2023-05-16
cloud: Implement `instance status`

elprans:
Also, print a normal error instead of panicking on other unimplemented
ops.

### DIFF
--- a/src/portable/reset_password.rs
+++ b/src/portable/reset_password.rs
@@ -11,6 +11,7 @@ use edgedb_tokio::credentials::Credentials;
 use edgeql_parser::helpers::{quote_string, quote_name};
 
 use crate::credentials;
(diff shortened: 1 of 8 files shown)


src/portable/reset_password.rs
```diff
@@ -47,7 +48,10 @@ pub fn reset_password(options: &ResetPassword) -> anyhow::Result<()> {
                 name
             }
         },
-        InstanceName::Cloud { .. } => todo!(),
+        InstanceName::Cloud { .. } => {
+            print::error("This operation is not supported yet cloud instances.");
```
fantix, line 52:
```suggestion
            print::error("This operation is not supported on cloud instances yet.");
```

fantix (approved):

PR #1045, opened 2023-05-17
migration create: Default value for conversion expressions

### LINKED ISSUE #995: Provide default expression for conversion expressions in migrations
Example:
image

tailhook:
Fixes #995

Example of how it looks like. Type change:
```
did you alter the type of property 'firstname' of object type 'default::Name'? [y,n,l,c,b,s,q,?]
> y
Please specify a conversion expression to alter the type of property 'firstname':
cast_expr> <std::int32>.firstname
```

Required field:
```
did you make property 'something' of object type 'default::Name' required? [y,n,l,c,b,s,q,?]
> y
Please specify an expression to populate existing objects in order to make property 'something' of object type 'default::Name' required:
fill_expr> <std::str>{}
```

This uses the same algorithm as the dev-mode. Which means it works. But the empty set might be a problem if there are values in the database. And that might only be figured out on production.

Alternative, is probably do not provide default for "required", only for type conversion. Or we can provide dictionary for default values for every type (i.e. empty string for strings and zero for numerics), but that might be not the best idea too.


tailhook:
Now relies on edgedb/edgedb#5506

fantix (approved):

PR #1086, opened 2023-06-15
Record and ask for the database name in `project init`

tailhook:
Relies on https://github.com/edgedb/edgedb-rust/pull/257 and https://github.com/edgedb/shared-client-testcases/pull/28

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1016,7 +1016,7 @@ dependencies = [
 [[package]]
 name = "edgedb-derive"
 version = "0.5.0"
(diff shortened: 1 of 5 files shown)


fantix (approved):
Ran into an issue: I have an EdgeDB cloud instance running with schema applied in database `edgedb`, then I tried to initialize a new empty project with the same instance but with a different database, it failed at:

```
Specify the name of the database: [default: edgedb]:
> yy
...
Applying migrations...
edgedb error: there are migrations in the database, but no migrations in "/tmp/yy/dbschema/migrations"
```

tailhook:
Oh, thank you @fantix. Fixed #1087

PR #1093, opened 2023-06-23
Add `version` to server info option

scotttrinh:
Want to add the server version in the `setup-edgedb` GitHub action and instead of needing to parse the JSON output, I thought it would be easier to add a `--get version` option to the `server info` command.

I didn't see any tests for bin-path, but if you can help point me at a good place for tests, I'm happy to add them!

### DIFF
--- a/src/portable/info.rs
+++ b/src/portable/info.rs
@@ -48,6 +48,14 @@ pub fn info(options: &Info) -> anyhow::Result<()> {
                 }
 
             }
+            "version" => {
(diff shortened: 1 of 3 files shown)


tailhook:
Perhaps we don't have a test for bin-path. You can add one around [here](https://github.com/edgedb/edgedb-cli/blob/master/tests/portable_smoke.rs#L35-L39).

elprans (approved):

PR #1100, opened 2023-07-10
Handle missing shared tests submod

scotttrinh:
The generic message you get isn't helpful for debugging _why_ the build failed.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -87,7 +87,8 @@ fn main() {
         "cargo:rerun-if-changed={}",
         connection_testcases.to_str().unwrap()
     );
-    let connection_testcases = fs::read_to_string(connection_testcases).unwrap();
+    let connection_testcases = fs::read_to_string(connection_testcases)
+        .expect("Could not find connection test cases: Ensure git submodules are included in your checkout.");
     let connection_testcases: Value = serde_json::from_str(&connection_testcases).unwrap();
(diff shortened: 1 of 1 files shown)


fantix (approved):

build.rs
```diff
@@ -83,12 +83,15 @@ fn main() {
     }
 
     let connection_testcases = testcases.join("connection_testcases.json");
+    if !connection_testcases.exists() {
+        return;
```
fantix, line 87: Maybe we can give some warning here, so that we know the shared test cases are not running for a reason?

scotttrinh: Seems reasonable! I'll move these checks earlier and centralize the handling of this "missing submodules" case.

scotttrinh: @fantix [`d2ac70c` (#1100)](https://github.com/edgedb/edgedb-cli/pull/1100/commits/d2ac70ccf8bba56f67b15c43772492d7fae5523e) ⚡

fantix (approved):

scotttrinh:
Alright, based on feedback from @tailhook , we'd like to continue to treat the missing submodule as a panic. In the future, we might possibly have it skip this whole thing if there are contexts where we're not running the tests.

⚡ 10b1d12

fantix (approved):

PR #1117, opened 2023-09-01
User-facing output changes

Dhghomon:
I'm currently working on a branch to tidy up the CLI but it has gotten quite large due to including both user-facing and developer-facing touchups (e.g. Clippy suggestions that were no brainers) which would make a review quite a challenge even if no behavior is being changed anywhere. So I've pulled out all the user-facing changes which form the entirety of this PR. It is essentially all documentation, and will allow Devon to take an undistracted look too.

Going forward I'm still doing a lot of fiddling around inside the built CLI to see what else I can find that is awkward or confusing from the POV of a user, and those changes will be in similar but much smaller PRs.

A bit of discussion about later activity: despite wanting to do so, I'm thinking it might make sense for me to give up on the developer-facing changes despite their tiny scope considering the delicateness of the CLI. To cite one example, the IsTerminal trait from the is_terminal crate we use is now part of the standard library:

https://crates.io/crates/is-terminal

> As of Rust 1.70, most users should use the [IsTerminal](https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html) trait in the Rust standard library instead of this crate.

Same goes for backtraces and OnceCell (though not Lazy yet). But even a small change like moving away from is_terminal requires building from a minimum of 1.70 (we build from 1.65, I think) and it's in no way crucial. So for the time being I am simply putting together a growing list of nice-to-haves for when the time comes.

If anyone who hasn't used Rust before wants to build this to try it out as a binary, you can do the following:

- Install Rust
- Clone the repo
- Change `edgedb` to `cargo run`. So edgedb help becomes cargo run help, and so on.

It takes a few minutes to compile and the very end (at least for me) when the linking it done it takes quite a while. So when a single change is done it'll recompile right away but still take a minute or two for the linker to finish. But if you aren't making any changes then the one build will be enough and afterwards it'll run as fast as any other binary.


fantix (approved):
:+1:

(I'll push a PR to fix the failing CI.)

elprans (approved):
A very nice cleanup, @Dhghomon!

PR #1119, opened 2023-09-07
point workflows at new uploader

RobertBlumen:
this may be temporary until the docker images catch up.


elprans:
This isn't strictly necessary. Landing edgedb/edgedb-pkg#59 is sufficient, as images get rebuilt on every workflow run.

PR #1126, opened 2023-09-18
Add `migration extract` command to sync `dbschema/migrations` with database

### LINKED ISSUE #270: Implement command to synchronize migrations in the filesystem with database
See edgedb/edgedb#2356 for context.

fantix:
Fixes #270

- [ ] Add test
- [ ] Add hint to `edgedb migrate`
- [ ] Add more hints

### DIFF
--- a/src/migrations/db_migration.rs
+++ b/src/migrations/db_migration.rs
@@ -1,0 +1,123 @@
+use std::collections::{BTreeSet, BTreeMap};
+use indexmap::IndexMap;
+
+use crate::connect::Connection;
+
+#[derive(Debug, Clone, edgedb_tokio::Queryable)]
+// TODO(tailhook) this has to be open-ended enumeration
+pub(crate) enum MigrationGeneratedBy {
+    DevMode,
+    DDLStatement,
+}
+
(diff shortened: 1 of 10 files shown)


src/migrations/db_migration.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::connect::Connection;
+
+#[derive(Debug, Clone, edgedb_tokio::Queryable)]
+// TODO(tailhook) this has to be open-ended enumeration
+pub(crate) enum MigrationGeneratedBy {
+    DevMode,
+    DDLStatement,
+}
+
+pub(crate) trait SortableMigration {
```
fantix, line 13: I think we don't need this trait any more as we are merging the impls - using the struct directly in `linearize_db_migrations()` would be enough.

fantix (commented):
LGTM otherwise. (should be an approval but GH doesn't allow me to approve my own PR)

src/migrations/db_migration.rs
elprans: I decided to leave it in because I plan further work to unify FS migrations with DB migrations at the trait level.

PR #1127, opened 2023-09-20
Drop an unnecessary unsafe

msullivan:


Dhghomon:
Yeah, time to get rid of that warning now that it's not unsafe anymore. Relevant PR for reference: https://github.com/edgedb/edgedb-rust/pull/272

Dhghomon (approved):

PR #1129, opened 2023-09-20
Dump progress bar

Dhghomon:
This adds a spinner bar when doing a database dump that updates each time a package comes in, which is 10MB at a time.

I did some looking around beforehand to see if there is any way to get the total database size and as far as I can tell there isn't, and even with pg_database_size it adds a performance penalty (quote from Elvis 3 years ago):

> The concern here is that pg_database_size is basically just a loop over a directory with a bunch of stat calls. It's not cached or updated/maintained by postgres. Hence, on somewhat slow (read: cloud) storage with a bunch of tables you can easily add tens of milliseconds of latency penalty per statement.

But a spinner bar is much better than the current behavior which just starts the dump and displays nothing until it's done.

Also changes u32::max_value() to u32::MAX as the method is set to be deprecated: https://doc.rust-lang.org/std/primitive.u32.html#method.max_value

I had originally added a prompt per database which looked pretty nice (output below) but it interferes with the current tests so removed that code for the moment as adding some sort of output to dump and restore is the most urgent priority. Might add something like an --interactive flag in a later PR though as being able to dump some databases and not others was nice to have.

```
     Running `target\debug\edgedb.exe dump --all --format=dir somedir`
Connecting to EdgeDB instance at localhost:10704...
Connecting to EdgeDB instance at localhost:10704...
Start dump for database edgedb? [y/n]
> n
Skipped dump for edgedb.

Connecting to EdgeDB instance at localhost:10704...
Start dump for database empty_db? [y/n]
> y
Starting dump for new_thing...
  Finished dump for empty_db. Total size: 0B
```

### DIFF
--- a/src/commands/dump.rs
+++ b/src/commands/dump.rs
@@ -1,17 +1,19 @@
 use std::path::{Path, PathBuf};
 
 use anyhow::Context;
+use indicatif::{ProgressBar, HumanBytes, ProgressStyle};
 use tokio::fs;
 use tokio::io::{self, AsyncWrite, AsyncWriteExt};
-use sha1::{Digest};
+use sha1::Digest;
 
-use tokio_stream::{StreamExt};
+use tokio_stream::StreamExt;
 
 use crate::commands::Options;
 use crate::commands::list_databases::get_databases;
 use crate::commands::parser::{Dump as DumpOptions, DumpFormat};
 use crate::connect::Connection;
 use crate::platform::tmp_file_name;
+use crate::question::Confirm;
 
 
 type Output = Box<dyn AsyncWrite + Unpin + Send>;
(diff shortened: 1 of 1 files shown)


raddevon:
I'm gonna let someone else do a proper code review, but everything looks and sounds good to me!

Dhghomon:
Thanks! Yeah, I put you on these to keep you in the loop, no need to learn
Rust and delve into the rest ^^

2023년 9월 20일 (수) 오후 8:19, Devon Campbell ***@***.***>님이 작성:

src/commands/dump.rs
```diff
@@ -1,11 +1,12 @@
 use std::path::{Path, PathBuf};
 
 use anyhow::Context;
+use indicatif::{ProgressBar, HumanBytes, ProgressStyle};
```
fantix, line 4: Like the warning says, `ProgressStyle` is an unused import.

fantix (approved):

src/commands/dump.rs
Dhghomon: Weird, rust-analyzer is blanking on me with the unused warnings. Thanks!

PR #1137, opened 2023-10-11
Add `instance create` options for compute and storage size

elprans:
The new `--compute-size` and `--storage-size` options allow specifying
the amount of compute and storage respectively.  The new `--free-tier`
(exclusive with `...-size`) can be used to explicitly request a Free
tier instance.

### DIFF
--- a/src/portable/create.rs
+++ b/src/portable/create.rs
@@ -190,10 +190,57 @@ fn create_cloud(cmd: &Create, org_slug: &str, name: &str, client: &cloud::client
     )?;
 
     let server_ver = cloud::versions::get_version(&query, client)?;
+    let tier = if cmd.free_tier {
+        String::from("Free")
+    } else {
+        String::from("Pro")
+    };
+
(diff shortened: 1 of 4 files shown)


src/portable/create.rs
```diff
@@ -190,10 +190,54 @@ fn create_cloud(cmd: &Create, org_slug: &str, name: &str, client: &cloud::client
...
+                },
+                cloud::ops::CloudInstanceResourceRequest{
+                    name: "storage".to_string(),
+                    value: storage_size,
+                },
+            ])
+        },
+    };
+
+    let resources_display = if req_resources.is_some() {
```
zackelan, line 220: we might want to display this even for free-tier? that way people know what they're getting in terms of compute / storage, and whets their appetite for "ooh, I want more storage..." and converting them to paid customers

elprans: Done

fantix (approved):

PR #1150, opened 2023-10-17
Add `instance create` options for compute and storage size (take two)

elprans:
This reverts commit 410d344cb2338d2fafde6b935cd31ae84da7a32d and uses
org payment method status to determine if `Free` or `Pro` should be the
default.

### DIFF
--- a/edgedb-cli-derive/src/into_args.rs
+++ b/edgedb-cli-derive/src/into_args.rs
@@ -16,9 +16,17 @@ pub fn structure(s: &types::Struct) -> TokenStream {
     for field in &s.fields {
         let ident = &field.ident;
         if field.attrs.flatten {
-            args.push(quote! {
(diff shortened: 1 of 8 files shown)


Dhghomon (approved):
Looks good from what I can see, some small clippy-like things you could change like `Tier: {tier:?}` instead of making a tier_str and `UsageError{kind: kind` to `UsageError{kind` if you want.

Maybe out of the scope of this PR but I'm curious, are there any plans to convert properties like preferred_payment_method and region into enums? I find those work well as semi-documentation for anyone poking through our code and wondering what the options are without having to look elsewhere.

elprans:
> Maybe out of the scope of this PR but I'm curious, are there any plans to convert properties like preferred_payment_method and region into enums?

Neither is an enum, those are database values.

Issue #1138, opened 2023-10-12
Add hint to `edgedb migrate`/`edgedb migration apply` for the case when user has dev mode migrations and is ready to break out of dev mode


raddevon:
When a user has dev mode migrations and tries to apply migrations, we give this error:

```
edgedb error: Database contains Dev mode migrations.
  Hint: Use `edgedb migrate --dev-mode` or `edgedb watch`
```

This hint is useful if the user wants to continue in dev mode, but it's more likely that they're running `migrate` because they're done with dev mode and need to apply proper migrations… they just haven't created a migration yet. Could we either replace this hint with one that reflect the more likely intent of the user or add a hint for that case? (If we add one, I think maybe the new hint should be the first one we present, but I'm open to alternatives.)

Dhghomon:
Currently working on this. One other item I noticed while doing so is that due to edgedb watch creating migrations in dev mode, and dev mode passing on a `true` when it comes to quiet output, any migration applied after edgedb watch is over don't show the usual `Applied (migration hash) 00001.edgeql` type output you usually see.

https://github.com/edgedb/edgedb-cli/blob/master/src/migrations/dev_mode.rs#L211

https://github.com/edgedb/edgedb-cli/blob/master/src/migrations/dev_mode.rs#L237

I'm curious, why is dev mode quiet by default in the first place?

PR #1151, opened 2023-10-18
Dev mode migration hints

Dhghomon:
Resolves https://github.com/edgedb/edgedb-cli/issues/1138

Adds a hint when starting up, from:

directory> edgedb watch
⠴ connecting
Connecting to EdgeDB instance at localhost:10704...
Initialized. Monitoring "C:\\directory".

To:

directory> edgedb watch
⠴ connecting
Connecting to EdgeDB instance at localhost:10704...
EdgeDB Watch initialized.
  Hint: Use `edgedb migration create` and `edgedb migrate --dev-mode` to apply changes once done.
Monitoring "C:\\directory".

Sometimes users might hit ctrl-c instead of opening a new terminal window, so changes 

edgedb error: Interrupted

to:

edgedb error: Interrupted by user

(Just to ensure that the user is aware that they caused the interruption as opposed to *something else* being interrupted when they hit ctrl-c)

And when user tries to apply migrations with the wrong command, from:

edgedb error: Database contains Dev mode migrations.
  Hint: Use `edgedb migrate --dev-mode` or `edgedb watch`

to:

edgedb error: Database contains Dev mode / `edgedb watch` migrations.
  Hint: Use `edgedb migration create` followed by `edgedb migrate --dev-mode`, or resume `edgedb watch`

I'm also curious about why dev mode is set to quiet by default as commented in the issue. Any thoughts or other context to add while the PR is open?

### DIFF
--- a/src/migrations/migrate.rs
+++ b/src/migrations/migrate.rs
@@ -263,8 +263,8 @@ async fn fixup(cli: &mut Connection, ctx: &Context,
         match db_migration.generated_by {
             Some(MigrationGeneratedBy::DevMode) => {
                 return Err(
-                    anyhow::anyhow!("Database contains Dev mode migrations.")
-                    .hint("Use `edgedb migrate --dev-mode` or `edgedb watch`")
+                    anyhow::anyhow!("Database contains Dev mode / `edgedb watch` migrations.")
+                    .hint("Use `edgedb migration create` followed by `edgedb migrate --dev-mode`, or resume `edgedb watch`")
                 )?;
             }
             Some(MigrationGeneratedBy::DDLStatement) => {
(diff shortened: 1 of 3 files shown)


elprans (approved):

src/migrations/migrate.rs
```diff
@@ -263,8 +263,8 @@ async fn fixup(cli: &mut Connection, ctx: &Context,
         match db_migration.generated_by {
             Some(MigrationGeneratedBy::DevMode) => {
                 return Err(
-                    anyhow::anyhow!("Database contains Dev mode migrations.")
-                    .hint("Use `edgedb migrate --dev-mode` or `edgedb watch`")
+                    anyhow::anyhow!("Database contains Dev mode / `edgedb watch` migrations.")
+                    .hint("Use `edgedb migration create` followed by `edgedb migrate --dev-mode`, or resume `edgedb watch`")
```
raddevon, line 267: @Dhghomon In the case that you've decided to leave dev mode and create the actual migration, wouldn't your `edgedb migrate` exclude the `--dev-mode` flag? You'd only need that to continue in dev mode, I think…

Issue #1152, opened 2023-10-18
\e panics CLI (clap regression)


msullivan:
Running `\e` now panics the CLI. Pretty sure this is a clap-related regression

```
edgedb> \e
thread 'tokio-runtime-worker' panicked at 'edit:entry: `Arg::trailing_var_arg` must accept multiple values', /home/sully/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.4.6/src/builder/debug_asserts.rs:548:17
stack backtrace:
   0: rust_begin_unwind
             at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/core/src/panicking.rs:67:14
   2: clap_builder::builder::debug_asserts::assert_app
   3: clap_builder::builder::command::Command::_build_self
             at /home/sully/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.4.6/src/builder/command.rs:3985:13
   4: clap_builder::builder::command::Command::_build_subcommand
             at /home/sully/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.4.6/src/builder/command.rs:4072:9
   5: clap_builder::parser::parser::Parser::parse_subcommand
             at /home/sully/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.4.6/src/parser/parser.rs:679:27
   6: clap_builder::parser::parser::Parser::get_matches_with
             at /home/sully/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.4.6/src/parser/parser.rs:458:17
   7: clap_builder::builder::command::Command::_do_parse
             at /home/sully/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.4.6/src/builder/command.rs:3862:29
   8: clap_builder::builder::command::Command::try_get_matches_from_mut
             at /home/sully/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.4.6/src/builder/command.rs:791:9
   9: clap_builder::builder::command::Command::try_get_matches_from
             at /home/sully/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.4.6/src/builder/command.rs:709:9
[...]

PR #1153, opened 2023-10-18
Fix some fallout from clap 4 port

elprans:
Enum-valued args should not use `value_parser`, it's both redundant and
wrong, and a `ValueEnum` derive should be used instead.

Another issue is that `trailing_var_arg` now requires `num_args` to be
set explicitly, which is also an improvement as `\e 10 12` would now get
rejected properly.

Fixes: #1152


msullivan (approved):

msullivan:
Thanks!

PR #1156, opened 2023-10-26
Make "Connecting to..." line more informative

elprans:
Include the instance name, and, if the instance is a Cloud instance,
don't show the host name as it's not very informative.

Also, in interactive mode, erase the "Connecting to..." line once the
connection has been established for a cleaner initial look.


fantix (approved):
Love the extraction of dup code!

PR #1157, opened 2023-10-26
Include instance name in prompt

elprans:
Make it evident what the current REPL session is connected to by showing

    org/instance:db>

or

    local-instance:db>

instead of just

    db>


1st1:
+1

1st1 (approved):

fantix (approved):

PR #1159, opened 2023-10-29
chore(nix): initial flake

MrFoxPro:
I was struggling to update nixpkgs's edgedb package, so I decied to nixify this repo instead.
Why you could consider to use Nix:
1. Declarative configuration of build system and dev shell
2. Repoducable packages from any machine
3. Same developer environment for every contributor. For example, share same edgedb version(s), rust toolchain etc.
4. Make onboarding easier for new contributors, as they will not have problems with setup. All will run with few commands
5. Attract users from Nix community
6. Simplify CI setup, as it can use or share configuration from your dev shell: https://determinate.systems/posts/nix-github-actions

Right now it's already possible to build edgedb-cli package via `nix build .?submodules=1`. Result binary will be located in `./result/bin/edgedb`

Current limitations:
- There is a little inconvinience because of using git submodules. Nix requires to pass `?submodules=1` to handle this. For using in consumer flake.nix, following syntax is required:
```nix
edgedb-cli = {
  type = "git";
  url = "https://github.com/mrfoxpro/edgedb-cli.git";
  submodules = true;
  ref = "nix";
};
```
Related PR: https://github.com/NixOS/nix/pull/7862
- Tests should be adjusted to run in isolated environment. I disabled tests for now (https://github.com/MrFoxPro/edgedb-cli/blob/9e793dbba528a61612a72780c5a048796d3c162d/flake.nix#L55)

I also created draft PR with developer environment setup with edgedb-server (https://github.com/edgedb/edgedb-cli/pull/1160). Related: https://github.com/NixOS/nixpkgs/issues/179635#issuecomment-1667764435

Current used Nix libs:

**flake-parts**: framework for convinient cross-platform flake shape: https://flake.parts. Allows splitting flake to modules, also provides opportunity to confiure some usefull tools declaratively

**fenix**: convinient rust toolchain managment: https://github.com/nix-community/fenix. Allows creating single package from multiple toolchains (cargo, rustc, rustc-src, rustfmt, rust-analyzer, clippy and so on)

**crane**: https://crane.dev. Usefull tool for effecient building of Rust project. Caches cargo artifacts. 

I think it would be nice to configure integration with cachix later, so users will download cached result instead of rebuilding themselfs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 /.vagga
 /target
 .vscode
+
+# Nix artifacts
+result
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,107 @@
+{
+  "nodes": {
+    "crane": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1698166613,
+        "narHash": "sha256-y4rdN4flxRiROqNi1waMYIZj/Fs7L2OrszFk/1ry9vU=",
+        "owner": "ipetkov",
+        "repo": "crane",
+        "rev": "b7db46f0f1751f7b1d1911f6be7daf568ad5bc65",
+        "type": "github"
+      },
+      "original": {
+        "owner": "ipetkov",
+        "repo": "crane",
+        "type": "github"
+      }
+    },
+    "fenix": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-analyzer-src": []
+      },
+      "locked": {
+        "lastModified": 1698474008,
+        "narHash": "sha256-t7uGih9Q/ucsXUnb7oWbI+J24+VmhP4wGmrs+4Fbp6c=",
+        "owner": "nix-community",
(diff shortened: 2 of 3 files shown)


aljazerzen:
@MrFoxPro this looks promising. Can I take over this effort (including #1160), as I also need a dev shell with appropriate tooling?

In my opinion, this PR is good as it is, I would only split the crane modules up a bit, so downstream flakes don't need to run all of the tests (and thus don't need the git submodule).

MrFoxPro:
> @MrFoxPro this looks promising. Can I take over this effort (including #1160), as I also need a dev shell with appropriate tooling?
> 
> In my opinion, this PR is good as it is, I would only split the crane modules up a bit, so downstream flakes don't need to run all of the tests (and thus don't need the git submodule).

for sure, you can push your changes if you're maintainer

scotttrinh:
I think this has been superceded by #1160 . Going to close.

aljazerzen:
This flake also contains derivations for building edgedb-cli itself, not just a shell for development.

Although, I'm planning to adapt PR to edgedb/nix-packages, so it can remain closed.

PR #1164, opened 2023-10-31
Fix configure set from CLI

Dhghomon:
Resolves https://github.com/edgedb/edgedb-cli/issues/1140

Currently most configure commands sent in directly from the CLI fail because the expected vs. actual types don't match. These are all simple queries under our control so seems easiest to just build a string for all of them with a cast in case of a scalar type like duration.

It also looks like clap was unhappy with ConfigBool which holds a bool, which it treats as a SetTrue action:

Argument 'value` is positional and it must take a value but action is SetTrue

https://docs.rs/clap/latest/clap/enum.ArgAction.html#variant.SetTrue

Which means that it assumes that `true` or `false` will not be following the parameter to set. You can modify the clap Command to allow this but ConfigBool itself isn't necessary either if we can just build a String and be done with it.

allow_bare_ddl is another interesting case as it feels like it's a bool but is actually an enum with the valiants AlwaysAllow and NeverAllow, so it handles the input ahead of time to let the user know which input is acceptable. Could even accept true or false on the user's end and convert it to the enum since there are only the two variants, would that be a good idea?

There could also be more input handling on the CLI side which could be nice since it doesn't require a database connection to happen first. (e.g. memory configs which have a minimum and maximum number of bytes) On Windows everything is a bit slower so connecting first to check an input always takes a few seconds. Could those be added as well or is it icky to have the CLI doing the same checks that the database then proceeds to do?

### DIFF
--- a/src/commands/configure.rs
+++ b/src/commands/configure.rs
@@ -1,41 +1,17 @@
+use std::fmt::Display;
+
 use edgeql_parser::helpers::{quote_string, quote_name};
 use crate::commands::Options;
 use crate::print;
 use crate::connect::Connection;
-use crate::commands::parser::{Configure, ConfigStr, ConfigBool};
-use crate::commands::parser::{AuthParameter};
+use crate::commands::parser::{Configure, ConfigStr, ListenPort, ListenAddresses, AuthParameter};
 
-
-async fn set_string(cli: &mut Connection, name: &str, value: &ConfigStr)
+async fn set(cli: &mut Connection, name: &str, cast: Option<&str>, value: impl Display)
     -> Result<(), anyhow::Error>
 {
-    print::completion(&cli.execute(
-        &format!("CONFIGURE INSTANCE SET {} := {}",
-            name, quote_string(&value.value)),
-        &(),
-    ).await?);
-    Ok(())
-}
-
-async fn set_bool(cli: &mut Connection, name: &str, value: &ConfigBool)
-    -> Result<(), anyhow::Error>
-{
-    print::completion(&cli.execute(
-        &format!("CONFIGURE INSTANCE SET {} := {}",
-            name, if value.value { "true" } else { "false" }),
-        &(),
-    ).await?);
-    Ok(())
-}
-
-async fn set_duration(cli: &mut Connection, name: &str, value: &ConfigStr)
-    -> Result<(), anyhow::Error>
-{
-    print::completion(&cli.execute(
-        &format!("CONFIGURE INSTANCE SET {} := <duration>{}",
-            name, quote_string(&value.value)),
-        &(),
-    ).await?);
+    let cast = cast.unwrap_or_default();
+    let query = format!("CONFIGURE INSTANCE SET {name} := {cast}{value}");
+    print::completion(&cli.execute(&query, &()).await?);
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


fantix (approved):
I think this is good and passes manual test. Please rebase and fix the warning tho.

> Could even accept true or false on the user's end and convert it to the enum since there are only the two variants, would that be a good idea?

I'm not so sure this is good, or even allowing arbitrary upper/lower cases for enum values. We used an enum instead of true/false or 1/0 in the first place, because (I think) we chose to define precise values for unambiguity.

> There could also be more input handling on the CLI side which could be nice since it doesn't require a database connection to happen first. (e.g. memory configs which have a minimum and maximum number of bytes) On Windows everything is a bit slower so connecting first to check an input always takes a few seconds. Could those be added as well or is it icky to have the CLI doing the same checks that the database then proceeds to do?

If our edgedb-rust codec got this covered, then sure, we could just handle the encoding error (?) then. But I wouldn't repeat a server-side schema-level constraint (like `listen_port`) as it might change in the future - not likely, but if it does, we won't want to change the CLI too as it's not a part of the binary-protocol.

Dhghomon:
> I think this is good and passes manual test. Please rebase and fix the warning tho.
> 
> > Could even accept true or false on the user's end and convert it to the enum since there are only the two variants, would that be a good idea?
> 
> I'm not so sure this is good, or even allowing arbitrary upper/lower cases for enum values. We used an enum instead of true/false or 1/0 in the first place, because (I think) we chose to define precise values for unambiguity.
> 
> > There could also be more input handling on the CLI side which could be nice since it doesn't require a database connection to happen first. (e.g. memory configs which have a minimum and maximum number of bytes) On Windows everything is a bit slower so connecting first to check an input always takes a few seconds. Could those be added as well or is it icky to have the CLI doing the same checks that the database then proceeds to do?
> 
> If our edgedb-rust codec got this covered, then sure, we could just handle the encoding error (?) then. But I wouldn't repeat a server-side schema-level constraint (like `listen_port`) as it might change in the future - not likely, but if it does, we won't want to change the CLI too as it's not a part of the binary-protocol.

Yes, agreed about taking over anything that the binary-protocol already handles.

Re: your "allowing arbitrary upper/lower cases for enum values", yes I think I'll take out the .to_ascii_lowercase() part then as the error message will still be nice and clear:

```
edgedb error: allow_bare_ddl may only be set to AlwaysAllow or NeverAllow
```

This is the one area where I do like checking the input on the CLI side as the message from the server is somewhat intimidating for a new user who assumes that the parameter is just a friendly bool and gets this:

```
Connecting to EdgeDB instance 'cli' at localhost:10704...
edgedb error: InvalidValueError: invalid input value for enum 'cfg::AllowBareDDL': "true"
```

fantix:
> ```
> invalid input value for enum 'cfg::AllowBareDDL': "true"
> ```

Ah, we should probably add an error hint on the server side here, in `def _interpret_invalid_text_repr()` where we know the schema type of the enum.

Dhghomon:
> > ```
> > invalid input value for enum 'cfg::AllowBareDDL': "true"
> > ```
> 
> Ah, we should probably add an error hint on the server side here, in `def _interpret_invalid_text_repr()` where we know the schema type of the enum.
> 
> ```diff
> diff --git a/edb/server/compiler/errormech.py b/edb/server/compiler/errormech.py
> index 607c3d0de..b804349df 100644
> --- a/edb/server/compiler/errormech.py
> +++ b/edb/server/compiler/errormech.py
> @@ -656,9 +656,19 @@ def _interpret_invalid_text_repr(
>      hint: Optional[str],
>      from_graphql: bool = False,
>  ):
> +    if hint is None:
> +        try:
> +            type_id = next(enum_re.finditer(err_details.message)).group("id")
> +            stype = schema.get_by_id(uuidgen.UUID(type_id), None)
> +            if stype.is_enum(schema):
> +                values = stype.get_enum_values(schema)
> +                hint = "Allowed values: " + ", ".join(values)
> +        except Exception:
> +            pass
>      return errors.InvalidValueError(
>          translate_pgtype(schema, err_details.message,
> -                         from_graphql=from_graphql)
> +                         from_graphql=from_graphql),
> +        hint=hint,
>      )
> ```
> 
> Produces:
> 
> ```
> edgedb> configure session  set allow_bare_ddl := 'abc';
> edgedb error: InvalidValueError: invalid input value for enum 'cfg::AllowBareDDL': "abc"
>   Hint: Allowed values: AlwaysAllow, NeverAllow
> ```

That looks great, much cleaner to have the server do it. I'll remove the lines validating the input then. Thanks!

PR #1173, opened 2023-11-06
Unbreak migration commands on old servers

elprans:
Changes in #1126 caused things to break when running against pre-3.x
servers because those lack the `Migration.generated_by` property.


fantix (approved):
👍

PR #1175, opened 2023-11-08
allow_bare_ddl applies to database not instance

Dhghomon:
I thought I had discovered a bug with `configure instance allow_bare_ddl` changing not actually allowing a user to use DDL but noticed inside this issue https://github.com/edgedb/edgedb/issues/4792 that it is set by database instead. (The InstanceConfig can indeed be set to AlwaysAllow or NeverAllow but that doesn't change the behavior when one actuall tries to use a DDL command).

### DIFF
--- a/src/commands/configure.rs
+++ b/src/commands/configure.rs
@@ -94,14 +94,18 @@ pub async fn configure(cli: &mut Connection, _options: &Options,
         C::Set(Set { parameter: S::QueryExecutionTimeout(ConfigStr { value }) }) => {
             set(cli, "query_execution_timeout", Some("<duration>"), format!("'{value}'")).await
         }
-        C::Set(Set { parameter: S::AllowBareDdl(ConfigStr { value }) }) => {
-            set(cli, "allow_bare_ddl", None, format!("'{value}'")).await
-        }
         C::Set(Set { parameter: S::ApplyAccessPolicies(ConfigStr { value }) }) => {
             set(cli, "apply_access_policies", None, value).await
         }
         C::Set(Set { parameter: S::AllowUserSpecifiedId(ConfigStr { value }) }) => {
             set(cli, "allow_user_specified_id", None, value).await
+        }
+        C::Set(Set { parameter: S::AllowBareDdl(ConfigStr { value }) }) => {
+            let query = format!("CONFIGURE CURRENT DATABASE SET allow_bare_ddl := '{value}'");
+            let res = &cli.execute(&query, &()).await?;
+            eprintln!("Current database within instance set to {value}.");
+            print::completion(res);
+            Ok(())
         }
         C::Reset(Res { parameter }) => {
             use crate::commands::parser::ConfigParameter as C;


msullivan:
I think there needs to be a way to specifically specify whether you are configuring the instance or a database.

Dhghomon:
> I think there needs to be a way to specifically specify whether you are configuring the instance or a database.

That's the funny part here, I thought this was a bug at first because it currently allows setting the instance to allow or not, but has no effect on actually allowing you to use DDL or not. I realized it wasn't a bug upon seeing this issue: https://github.com/edgedb/edgedb/issues/4792

So the user could be allowed to configure allow_bare_ddl by either instance or database but setting to instance wouldn't actually do anything as far as I can tell.

msullivan (changes requested):
Well, it *does* have an effect if you set it on an instance, but only if it hasn't also been set on the database, which overrides it.
I do agree that in this case it only really makes sense to set it on a database.

But I do not think that it makes sense to have `edgedb configure` silently decide whether to do an instance config or a database config, based on what the param is.

We can have instance configure of allow_bare_ddl produce an error, if we want, with a message directing the user to configure the database, but instance and database configuration commands must be clearly distinguished.

Dhghomon:
> Well, it _does_ have an effect if you set it on an instance, but only if it hasn't also been set on the database, which overrides it. I do agree that in this case it only really makes sense to set it on a database.
> 
> But I do not think that it makes sense to have `edgedb configure` silently decide whether to do an instance config or a database config, based on what the param is.
> 
> We can have instance configure of allow_bare_ddl produce an error, if we want, with a message directing the user to configure the database, but instance and database configuration commands must be clearly distinguished.

Agreed on silently deciding what to do. The tough part here is that when setting from the CLI you don't have the option to choose between one or the other: it's just `edgedb configure set` which then gets turned into `CONFIGURE INSTANCE SET` queries, so it's already silently deciding on instance over database. Not sure what the solution to that is though one small silver lining is that the output does tell you if it the configuration applied to the database or the instance (OK: CONFIGURE INSTANCE or OK: CONFIGURE DATABASE).

For the time being though I think it's safe to assume that a user that types `edgedb configure set allow_bare_ddl AlwaysAllow` *does* want to use DDL and will be surprised to find it not working which is what it currently does.

Or maybe the correct thing to do is a bigger change to have the CLI require `instance` or `database` between `configure` and `set`? Would users find it usefor to configure other parameters (like memory) by database instead of instance which is what they currently do? I guess it depends on whether the allow_bare_ddl is an outlier. If users will never want to set the other parameters by instance then the ddl setting would be a single exception, if not then we'd want to change the overall command to give them more fine-tuned control.

msullivan:
I think that we should allow writing `configure instance set <whatever>` and `configure database set <whatever>`, and continue defaulting to `instance` if you write bare `configure set` (for compatability).

Dhghomon:
> I think that we should allow writing `configure instance set <whatever>` and `configure database set <whatever>`, and continue defaulting to `instance` if you write bare `configure set` (for compatability).

Okay, that sounds good. Plus maybe output a small warning when a user goes for `configure set allow_bare_ddl` that configuring an instance to allow it won't actually allow DDL on the individual database(s) so they aren't confused.

PR #1184, opened 2023-12-04
Don't overwrite existing dump file by default, make database name clearer

Dhghomon:
Having a database dump instantaneously overwrite any existing files of the same name has always made me a bit nervous and I think we should require the user to add an `--overwrite-existing` to the command to make this behaviour explicit. Some current users may be annoyed but `edgedb dump --help` will quickly show the option to overwrite and I think it's best to get this out of the way now, much better than a panicked "I mistakenly overwrote some incredibly crucial file and is there any way to restore it...?" issue later on.

The current behaviour calls create() which makes an OpenOptions with the following parameters set:

https://doc.rust-lang.org/src/std/fs.rs.html#394

So keep that in case of overwrite-existing, and use .create_new() otherwise.

When using `--all` I think automatic overwriting is fine as is since the user has to specify a directory, and the PR adds a note to that effect that using --all will have this effect. Also dump_all() is used when upgrading an instance so best to leave that untouched.

Also noticed that having the default database be edgedb makes the output a bit unclear sometimes:

"Starting dump for edgedb", "Finished dump for edgedb" and "Database edgedb dump" all look like it might be referring to EdgeDB in general. So just add some ticks around the database name.


src/commands/dump.rs
```diff
@@ -24,29 +24,39 @@ pub struct Guard {
...
-            let file = fs::File::create(&filename).await
+            let file = options
+                .open(&filename).await
                 .context(filename.display().to_string())?;
             Ok((Box::new(file), Guard { filenames: None }))
         } else {
             let tmp_path = filename.with_file_name(
                 tmp_file_name(filename.as_ref()));
-            if fs::metadata(&tmp_path).await.is_ok() {
+            if fs::metadata(&tmp_path).await.is_ok() && overwrite_existing {
```
fantix, line 48: I think this doesn't seem right - the temporary file doesn't usually exist, and we should instead protect the `fs::rename()` in `fn commit()`.

mmastrac (approved):
LGTM!

PR #1187, opened 2023-12-12
Extra help when config incorrect

Dhghomon:


src/connect.rs
```diff
@@ -176,7 +176,12 @@ impl Connector {
             _ =>
                 format!("EdgeDB instance at {}", cfg.display_addr())
         };
-        format!("Connecting to {}...", desc)
+        let wait = cfg.wait_until_available();
+        let wait_msg = match wait.as_secs() {
+            seconds if seconds < 1 => format!("{}ms", wait.as_millis()),
+            seconds => format!("{seconds}s")
+        };
+        format!("Connecting to {desc} (will try up to {wait_msg})...\n")
```
raddevon, line 184: Is this telling me the number of retries or the time? I assume it's the time since `wait_msg` is set to `wait.as_secs()`, but it's not entirely clear based on the message. I.e., is the next word "times" or "seconds?" 

Maybe `wait_msg` includes "seconds," and if so, ignore this suggestion.
```suggestion
        format!("Connecting to {desc} (will try for up to {wait_msg} seconds)...\n")
```

raddevon (commented):
Love these changes!

src/connect.rs
Dhghomon: Yeah, it'll be seconds or milliseconds (though far more likely to be seconds), written s or ms. Could change s to seconds though.

Interestingly during the 30 seconds you can see it trying over and over again if you set the RUST_LOG to something like debug so we could technically put those messages in if it makes sense. It's noisier but the advantage would be that the user gets to see the problem within five seconds instead of twiddling thumbs until the 30 seconds have passed / killing the process.

Here's the output with debug, could pull out those debug messages and print them a bit nicer:

```
PS C:\rust\edgedb-cli> cargo run
    Finished dev [optimized + debuginfo] target(s) in 0.45s
     Running `target\debug\edgedb.exe`
[2023-12-12T13:58:31Z DEBUG edgedb::version_check] Cached version None
[2023-12-12T13:58:31Z INFO  edgedb_tokio::raw::connection] Connecting via TCP localhost:10757
[2023-12-12T13:58:31Z DEBUG rustyline] PathInfo("C:\\Users\\mithr\\AppData\\Local\\edgedb\\edgeql.history", SystemTime { intervals: 133468273415897816 }, 100)
Connecting to EdgeDB instance 'edgedb_cli' at localhost:10757 (will try up to 30s)...
[2023-12-12T13:58:35Z DEBUG edgedb_tokio::raw::connection] Temporary connection error: ClientConnectionError: No connection could be made because the target machine actively refused it. (os error 10061)
[2023-12-12T13:58:39Z DEBUG edgedb_tokio::raw::connection] Temporary connection error: ClientConnectionError: No connection could be made because the target machine actively refused it. (os error 10061)
[2023-12-12T13:58:43Z DEBUG edgedb_tokio::raw::connection] Temporary connection error: ClientConnectionError: No connection could be made because the target machine actively refused it. (os error 10061)
[2023-12-12T13:58:47Z DEBUG edgedb_tokio::raw::connection] Temporary connection error: ClientConnectionError: No connection could be made because the target machine actively refused it. (os error 10061)
[2023-12-12T13:58:51Z DEBUG edgedb_tokio::raw::connection] Temporary connection error: ClientConnectionError: No connection could be made because the target machine actively refused it. (os error 10061)
[2023-12-12T13:58:56Z DEBUG edgedb_tokio::raw::connection] Temporary connection error: ClientConnectionError: No connection could be made because the target machine actively refused it. (os error 10061)
[2023-12-12T13:59:00Z DEBUG edgedb_tokio::raw::connection] Temporary connection error: ClientConnectionError: No connection could be made because the target machine actively refused it. (os error 10061)
[2023-12-12T13:59:04Z DEBUG edgedb_tokio::raw::connection] Temporary connection error: ClientConnectionError: No connection could be made because the target machine actively refused it. (os error 10061)
edgedb error: ClientConnectionFailedError: cannot establish connection for 30s: No connection could be made because the target machine actively refused it. (os error 10061)
error: process didn't exit successfully: `target\debug\edgedb.exe` (exit code: 1)```

Issue #1185, opened 2023-12-05
Switch `edgedb migrate` away from running everything in a single transaction, add flag to opt back in


elprans:
Currently, `edgedb migrate` will open a transaction and run all outstanding migrations in it.  This is problematic with respect to certain operations that cannot coexist in a single transaction, such as alteration of enum types and usage of the new enum value.  Additionally, holding the transaction across all migrations prolongs the lifespan of various locks taken, reducing availability of the database while migrations are running.

PR #1189, opened 2023-12-12
Remove transaction from edgedb migrate

quinchs:
Removes the transaction block from `edgedb migrate`, as proposed in https://github.com/edgedb/edgedb-cli/issues/1185.

Closes https://github.com/edgedb/edgedb-cli/issues/1185

### DIFF
--- a/src/migrations/options.rs
+++ b/src/migrations/options.rs
@@ -120,7 +120,7 @@ pub struct Migrate {
     #[arg(long)]
     pub dev_mode: bool,
 
-    /// Runs the migration(s) in a single transaction (legacy behaviour).
+    /// Runs the migration(s) in a single transaction.
     #[arg(long="single-transaction")]
     pub single_transaction: bool,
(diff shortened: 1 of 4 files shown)


elprans (commented):
We should add an option to opt into the old behavior.  E.g `--single-transaction`.

src/migrations/options.rs
```diff
@@ -119,6 +119,10 @@ pub struct Migrate {
     /// a regular `.edgeql` file, after which the above query will return nothing.
     #[arg(long)]
     pub dev_mode: bool,
+
+    /// Runs the migration(s) in a single transaction (legacy behaviour).
```
elprans, line 123: Wanting a single transaction might be a legitimate case, so lets not scare users with "legacy".

```suggestion
    /// Runs the migration(s) in a single transaction.
```

elprans (approved):

PR #1200, opened 2024-01-31
Feat: Variable parsing rework (Array input)

quinchs:
## Summary
This PR reworks the variable input mechanism to use [`nom`](https://github.com/rust-bakery/nom/tree/main) as a parser; allowing us to parse array inputs and other complex inputs.

### Array Inputs
The array input works much like writing an array in EdgeQL, as proposed by Sully:
```edgeql
cli> select <array<int64>>$arg;
Parameter <array>$arg: [1, 3, 5]
{[1, 3, 5]}
```

and it also forces `str` to be quoted with either single quotes or double quotes in the input:
```edgeql
cli> select <array<str>>$arg;
Parameter <array>$arg: ["abc", 'def']
{['abc', 'def']}
```

### Error reporting and hints
With the change to nom, errors and hints have changed too as a side effect, in the current state of this PR it dumps the parser errors pretty verbosely, I would like some feedback on how we should format these errors to be as concise as possible, some examples:
image
image
image
image

Things get pretty messy with some of the parsers though:
image


1st1:
Can you add a few examples of how successful input looks like? What if you have an array of tuples which have arrays in them?

msullivan:
We definitely shouldn't be including things like `TakeWhileMN` in error messages

msullivan:
We need to figure out how to make check warnings not appear once for each platform

src/prompt/variable.rs
```diff
@@ -16,31 +35,200 @@ use rustyline::hint::Hinter;
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum StringFragment<'a> {
+    Literal(&'a str),
+    EscapedChar(char),
+    EscapedWS,
+}
+
+// heavily based on https://github.com/rust-bakery/nom/blob/main/examples/string.rs
+fn quoted_str_parser<'a>(input: &'a str, quote: char, esc: &str) -> IResult<&'a str, String, ParsingError> {
```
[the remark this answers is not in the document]
msullivan: I think the easiest (and best) way to do this would be to take advantage of some existing code in our tokenizer for this: https://github.com/edgedb/edgedb/blob/62d8fcfcbeded747106a8914e1b4ef4bac37aa61/edb/edgeql-parser/src/helpers/strings.rs#L63-L168

So I think the approach then would be to match a quoted string *uninterpreted* (only understanding escapes as much as is needed to allow an escaped quote), and then passing that uninterpreted quoted string to unquote_string)

src/prompt/variable.rs
```diff
@@ -49,8 +237,65 @@ pub struct Uuid;
...
+                    char('-'),
+                    nom::bytes::streaming::take_while_m_n(
+                        12,
+                        12,
+                        char::is_alphanumeric,
+                    ),
+                )),
+                |s| uuid::Uuid::from_str(&[s.0, s.2, s.4, s.6, s.8].join("")).context("UUID format is malformed")
+            )(i)
+        }
```
msullivan, line 287: I would just grab 36 characters and pass it to `Uuid::from_str` instead of trying to parse it out directly

src/prompt/variable.rs
```diff
@@ -134,10 +398,41 @@ pub struct Decimal;
...
-        let dec: BigDecimal = input.parse().map_err(no_pos_err)?;
-        let dec = dec.try_into().map_err(no_pos_err)?;
-        Ok(Value::Decimal(dec))
+    fn parse<'a>(&self, input: &'a str, _flags: InputFlags) -> ParseResult<'a> {
+        context(
+            "decimal",
+            map(
+                map_res(
+                    map_res(
+                        recognize_float_parts,
```
msullivan, line 407: I think that if you use `recognize` on `recognize_float_parts`, you can get back the string and then parse that directly.

https://docs.rs/nom/latest/nom/combinator/fn.recognize.html

src/prompt/variable.rs
```diff
@@ -146,27 +441,82 @@ pub struct Json;
...
-        match serde_json::from_str::<serde_json::Value>(input) {
-            Err(e) if e.classify()  == serde_json::error::Category::Eof
-            => Err(Error::Incomplete),
-            Err(e) => Err(no_pos_err(e)),
-            Ok(_) => Ok(Value::Json(
-                model::Json::new_unchecked(input.into())
-            )),
-        }
+    fn parse<'a>(&self, input: &'a str, _flags: InputFlags) -> ParseResult<'a> {
+        // treat as string, and then validate the strings content
```
msullivan, line 445: This is not right; the json might not be a string.

This one is kind of a pain in the ass. serde does not seem to care about making this easy.
I think to do it we'd want to use https://docs.rs/serde_json/latest/serde_json/struct.StreamDeserializer.html, where we can parse one value out of the stream and then use `byte_offset` to figure out how much to advance the parser

Another thing we *could* do is requiring it be quoted in the nested quotes required mode. We could even make the syntax in that mode be `to_json('...')`.

src/prompt/variable.rs
```diff
@@ -146,27 +441,82 @@ pub struct Json;
...
+
+impl VariableInput for Array {
+    fn type_name(&self) -> &str { "array" }
+    fn parse<'a>(&self, input: &'a str, _flags: InputFlags) -> ParseResult<'a> {
+        context(
+            "array",
+            map(
+                preceded(
+                    char('['),
+                    cut(
```
msullivan, line 473: Why is `cut` needed?

src/prompt/variable.rs
```diff
@@ -186,43 +536,57 @@ impl Hinter for VarHelper {
...
     fn highlight<'l>(&self, line: &'l str, _pos: usize) -> Cow<'l, str> {
-        match self.var_type.parse(line) {
-            Ok(_) => line.into(),
+        match self.var_type.parse(line, InputFlags::None) {
+            Ok(r) => {
+                if r.0.len() == 0 {
+                    return line.into()
+                }
+
+                let mut str = line[..(line.len() - r.0.len())].to_string();
```
msullivan, line 562: explain?

msullivan (commented):
I think the overall approach is solid, but have a bunch of suggestions about various parsers.

Also, we need some tests

src/prompt/variable.rs
```diff
@@ -16,31 +35,200 @@ use rustyline::hint::Hinter;
...
 use rustyline::{Helper, Context};
 
+type ParseResult<'a, I = &'a str, R = Value> = IResult<I, R, ParsingError>;
 
 #[derive(Debug, thiserror::Error)]
-pub enum Error {
+pub enum ParsingError {
     #[error("{}", description)]
-    Mistake { offset: Option<usize>, description: String },
+    Mistake { kind: Option<ErrorKind>, description: String },
```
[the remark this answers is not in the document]
quinchs: No, `ParsingError` cant directly calculate the offset since its constructed with only the context of the current slice. An offset can be calculated later when we format `ParsingError`, which sounds like a good idea.

src/prompt/variable.rs
quinchs: should we allow this kind of input?
```
33d3ca20-149c-4a49-b8ac-a5a8ff9cc6b3 // OK
33d3ca20149c4a49-b8ac-a5a8ff9cc6b3-- // OK
----33d3ca20149c4a49b8aca5a8ff9cc6b3 // OK
```

quinchs: or wait~ does `Uuid::from_str` validate all that? iirc it doesn't allow `-`

quinchs: oh it does allow, will fix

src/prompt/variable.rs
quinchs: Mainly for if one element fails to parse it should be unrecoverable

src/prompt/variable.rs
quinchs: slices the original input to remove the remaining unparsed content from it, to then color the remaining red

msullivan:
(Get rid of all the unused imports)

src/prompt/variable.rs
msullivan: put that in a comment please

src/prompt/variable.rs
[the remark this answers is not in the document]
msullivan: (That is what we do in our actual lexer)

src/prompt/variable.rs
quinchs: I guess it works without

quinchs: It just makes it error out with `nom::internal::Err::Failure` instead of `nom::internal::Err::Error` but it won't change any of the parent parsers behaviour

src/prompt/variable.rs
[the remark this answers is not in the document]
quinchs: image
I've updated the parser to follow the same rules of unicodes and such, I tried integrating `unquote_string` but I failed at it. Maybe if you already know how to integrate it you can.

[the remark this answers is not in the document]
msullivan: Could you elaborate on what the snags were? What I would probably recommend is writing a combinator that matches a quoted string but doesn't really worry about what the return value is or if the escapes are valid; then you can use that with `recognize`, and pass the result of that to `unquote_string`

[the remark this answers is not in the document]
msullivan: https://github.com/edgedb/edgedb/blob/master/edb/edgeql-parser/src/tokenizer.rs#L659-L669 is the tokenizer code for matching a quoted string; one option (though not the most idiomatic for parser combinators) is to just write a combinator directly that does that logic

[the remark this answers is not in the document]
quinchs: > Could you elaborate on what the snags were?

Getting it to match a quoted string while ignoring other escape sequences, ex `\x33` or `\u236A`, kept either overrunning the quoted string (greedily) or not matching anything at all because of non recognizable escapes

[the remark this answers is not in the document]
msullivan: What did the combinators you were producing look like?
I *think* that for single quote strings, it should work to match a sequence of:
 * any non `'` non `\` character
 * `\` followed by any other character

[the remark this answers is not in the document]
quinchs: something like:
```rs
    context(
        "quoted_string",
        delimited(
            char(quote),
            map_res(
                fold_many0(
                    alt((
                        map(
                            verify(is_not(esc), |s: &str| !s.is_empty()),
                            StringFragment::Literal
                        ),
                        map(
                            preceded(
                                char('\\'),
                                alt((
                                    value('\\', char('\\')),
                                    value('/', char('/')),
                                    value('"', char('"')),
                                    value('\'', char('\'')),
                                ))
                            ),
                            StringFragment::EscapedChar
                        ),
                        value(
                            StringFragment::EscapedWS,
                            preceded(char('\\'), nom::character::streaming::multispace1)
                        ),
                    )),
                    String::new,
                    |mut string, fragment| {
                        match fragment {
                            StringFragment::Literal(s) => string.push_str(s),
                            StringFragment::EscapedChar(c) => string.push(c),
                            StringFragment::EscapedWS => {}
                        }
                        string
                    }
                ),
                |s| unquote_string(&s).map(|v| v.into()).context("Failed to unquote string")
            )
            char(quote),
        )
    )(input)
```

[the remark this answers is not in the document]
msullivan: What is `esc` here? The relevant quote and backslash?

One problem here is that `unquote_string` expects to get the full quoted string, *including* the quote characters. I also think the escape handling is busted (it doesn't match all escape characters (I think) and it *interprets* some of the escape characters, which will do the wrong thing once passed to unquote_string).

I think the best approach remains the one I outlined above: write a combinator that *recognizes* a quoted string without interpreting it at all (it should probably just produce `()` as its value), then use `recognize` to get the string out and call `unquote_string`.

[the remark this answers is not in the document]
msullivan: You can follow up with me on slack if you have any questions

src/prompt/variable.rs
```diff
@@ -146,27 +402,111 @@ pub struct Json;
...
-            )),
-        }
+    fn parse<'a>(&self, input: &'a str, flags: InputFlags) -> ParseResult<'a> {
+        context(
+            "json",
+            |s: &'a str| {
+                let de = serde_json::Deserializer::from_str(s);
+                let mut stream = de.into_iter::<serde_json::Value>();
+
+                // let the stream do the tokenization gradually until we hit EOF or a syntax error
```
msullivan, line 412: I think that the right thing to do here is to *always* do exactly one iteration of the loop.

If there is anything not consumed in the input, that will be an error up at a higher level, right, so there is no need to handle that here?

I think this should try to always parse exactly one json value, and always validate that it parsed correctly, and then leave any remainder to be handled by calling code.

(Either to be parsed, if it is an array or a tuple, or to be errored out on, if this was a top level value)

quinchs:
> Can you add a few examples of how successful input looks like? What if you have an array of tuples which have arrays in them?

#### Array of tuples
image

#### Array of tuples with arrays
image

#### Array of json
image

src/prompt/variable.rs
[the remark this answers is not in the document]
quinchs: I think I got a working impl using `unquote_string`

src/prompt/variable.rs
```diff
@@ -131,106 +133,70 @@ fn quoted_str(input: &str) -> IResult<&str, String, ParsingError> {
...
+                        if str.is_empty() {
+                            return Err(Error(ParsingError::Incomplete));
+                        }
+
+                        if str.len() == 1 && str.chars().nth(0).unwrap() != quote {
+                            return Err(Error(ParsingError::Mistake {
+                                kind: None,
+                                description: "Missing end quote on string".to_string()
+                            }))
+                        }
```
msullivan, line 166: This logic shouldn't be necessary for correctness, right? Since if either of those conditions hold, the loop below should exit with `pos == 0` and then the `char(quote)` should cause the error.

quinchs: Seems to be the case, I'll remove the redundant code

src/prompt/variable.rs
```diff
@@ -409,28 +409,18 @@ impl VariableInput for Json {
...
-                    }
-
-                    // if there is any errors at all with the input, reflect that back out and stop parsing
-                    match r {
-                        Ok(_a) => {},
+                // consume a single json value
+                match stream.next() {
+                    Some(r) => match r {
+                        Ok(_) => {},
+                        Err(_) if flags.contains(InputFlags::JSON_REMAINDER) => {},
```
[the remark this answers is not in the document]
quinchs: It was a precursor of the old implementation, I think its safe to remove

src/prompt/variable.rs
```diff
@@ -131,106 +133,70 @@ fn quoted_str(input: &str) -> IResult<&str, String, ParsingError> {
...
+                        }
+
+                        if str.len() == 1 && str.chars().nth(0).unwrap() != quote {
+                            return Err(Error(ParsingError::Mistake {
+                                kind: None,
+                                description: "Missing end quote on string".to_string()
+                            }))
+                        }
+
+                        let chars = str.char_indices();
```
msullivan, line 157: nit: I'd just inline this below into line 172

src/prompt/variable.rs
```diff
@@ -131,106 +133,70 @@ fn quoted_str(input: &str) -> IResult<&str, String, ParsingError> {
...
+                            }
+
+                            if c == quote {
+                                // is it escaped?
+                                if prev.unwrap() == '\\' {
+                                    prev = Some(c);
+                                    continue;
+                                }
+
+                                // not escaped, its the end of the quoted string
```
msullivan, line 174: grammar nit: it's

src/prompt/variable.rs
```diff
@@ -131,106 +133,70 @@ fn quoted_str(input: &str) -> IResult<&str, String, ParsingError> {
...
-                    match fragment {
-                        StringFragment::Literal(s) => string.push_str(s),
-                        StringFragment::EscapedChar(c) => string.push(c),
-                        StringFragment::EscapedWS => {}
-                    }
-                    string
-                }
             ),
-            char(quote),
+            |s| unquote_string(s).map(|v| v.into()).context("Failed to unquote string")
```
msullivan, line 188: 'Failed to unquote string' is pretty obscure for a user-facing message, I think

src/prompt/variable.rs
```diff
@@ -131,106 +133,70 @@ fn quoted_str(input: &str) -> IResult<&str, String, ParsingError> {
...
+                                // not escaped, its the end of the quoted string
+                                pos = i;
+                                break;
+                            }
+
+                            prev = Some(c)
+                        }
+
+                        // slice out the actual string part, and the remainder based on the 'pos'
+                        Ok((&str[pos..], &str[..pos]))
```
msullivan, line 183: Probably pointless nit: maybe return `()` instead of `&str[pos..]`, since we don't need the real string. (It doesn't really matter, since it isn't like a string slice is expensive to take, but it seems a little cleaner)

src/prompt/variable.rs
quinchs: It will include the message from the `Result` of `unquote_string` in the actual error message, I'll still change it to something more direct

src/prompt/variable.rs
quinchs: `&str[pos..]` will be needed if the parent parser has more to do, ex `tuple<str>` or `array<str>` need to match `)` and `]` respectfully after parsing a string

src/prompt/variable.rs
```diff
@@ -131,106 +133,70 @@ fn quoted_str(input: &str) -> IResult<&str, String, ParsingError> {
...
+
+                        let chars = str.char_indices();
+
+                        let mut prev = None;
+
+                        for (i, c) in chars {
+                            if prev == None {
+                                prev = Some(c);
+                                continue;
+                            }
```
msullivan, line 161: This will cause us to fail to parse `""` correctly, since the trailing `"` will get skipped. I think you should drop this and turn the check below into something like `prev == Some('\\')`.

And then, if you merge the escape check into the quote check, like `c == quote && prev != Some('\\')`, then we don't need any nonlocal control flow except the loop break

src/prompt/variable.rs
```diff
@@ -131,106 +133,70 @@ fn quoted_str(input: &str) -> IResult<&str, String, ParsingError> {
...
+
+                            if c == quote {
+                                // is it escaped?
+                                if prev.unwrap() == '\\' {
+                                    prev = Some(c);
+                                    continue;
+                                }
+
+                                // not escaped, its the end of the quoted string
+                                pos = i;
```
msullivan, line 171: I *think* that we ought to unconditionally update `pos`, so we always consume however much of the string we actually processed, even when we run off the end without a quote. The best way to do that might be to get rid of `i` and just increment `pos` at the bottom of the loop.
I'm not sure if it will actually matter, but it should allow the parser to have a better understanding of where the error happened?

src/prompt/variable.rs
msullivan: Oh, sorry, I meant instead of `&str[..pos]`!

msullivan:
The big remaining thing needed is tests. There probably *ought* to have been some tests before, but now the parsing is much more involved so we definitely need them.

src/prompt/variable.rs
quinchs: It doesn't fail to parse since `pos` is never updated in that condition, parsing `""` looks like the following:
- first quote gets captured by `char(quote)`
- `str.char_indices()` has 1 iteration; being the second quote
- the `if prev == None` case triggers since its the first iteration
- loop exits because theres no more elements in the iterable
- `Ok` is returned, and since we didn't update `pos`, its remainder is `"` which is the same as the input.

src/prompt/variable.rs
quinchs: We could, but since the closure doesn't have an error case I don't think its necessary, we can just let either `char(quote)` fail or have `unquote_string` fail with its error message.

msullivan: Well, the weird thing is that `char(quote)` will fail at the "wrong position"

src/prompt/variable.rs
msullivan: That's only true in isolation, where the `"` is at the end of the input. Consider parsing `["", "foo"]`.
You'll skip the closing `"` on the first string, then carry on until you hit the opening quote of `"` which will be incorrectly interpreted as a close quote.

msullivan:
```
_localdev:edgedb> select <array<str>>$0;
Parameter <array>$0: ["", "asdf"] -- array -> Expected ']' in "asdf\"]"
```

quinchs: Adding
```rs
 if c == quote {
     break;
 }

```

into the condition fixes this

src/prompt/variable.rs
quinchs: Could you give an example input where it would fail in the wrong position?

msullivan: If you have `"abcdef`, `char(quote)` will fail because `a` is not a `"`, right? Which is kind of weird, because that doesn't really seem to be what the problem is here.

quinchs: image

since quoted strings are never alone and are always wrapped it propigates it to the parent parser, in the case of `array<str>` it just complains about the missing `]` since the string doesn't consume anything until a quote.

src/prompt/variable.rs
```diff
@@ -156,6 +156,11 @@ fn quoted_str_parser<'a>(input: &'a str, quote: char) -> IResult<&'a str, String
 
                         for (i, c) in str.char_indices() {
                             if prev == None {
+                                // first iteration, we want to make sure that this is not a quote
```
msullivan, line 157: I think this works but it has gotten very tangled. (Two breaks, two continues, plus the normal loop termination condition.)

I'd recommend trying out my suggestion from the earlier comment, which I think gets it down to one break and no continues

src/cloud/ops.rs
```diff
@@ -85,19 +98,49 @@ pub struct Version {
     pub version: String,
 }
 
+#[derive(Debug, serde::Deserialize)]
```
msullivan, line 101: hm..., weird? why is github including this in the diff?

src/prompt/variable.rs
```diff
@@ -16,31 +28,182 @@ use rustyline::hint::Hinter;
...
+                    move |str: &'a str| {
+                        let mut pos = 0;
+                        let mut prev = None;
+                        let mut complete = false;
+
+                        for c in str.chars() {
+                            // check for a quote, if prev is none then its the quote is the first char of the string
+                            // meaning it isn't escaped; but if we have a previous char then check whether it's the
+                            // escape char '\'
+                            if c == quote && (prev == None || (prev.is_some() && prev.unwrap() != '\\')) {
```
msullivan, line 160: Could we do `prev != Some('\\')`.

src/prompt/variable.rs
```diff
@@ -146,27 +361,178 @@ pub struct Json;
...
+                    ),
+                ),
+                |v| Value::Array(v),
+            ),
+        )(input)
+    }
+}
+
+#[derive(Debug)]
+pub struct Tuple {
```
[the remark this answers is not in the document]
quinchs: What should the syntax be for this?
`(NAME := VALUE, ...)`?

src/cloud/ops.rs
quinchs: no clue

src/prompt/variable.rs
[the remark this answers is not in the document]
quinchs: Regular tuples are supported https://github.com/edgedb/edgedb-cli/pull/1200/files#diff-4f169bfb17a93405713417c53e1b55bcc7a4162f1e7a6b2f5144846278e5518fR439-R478

quinchs:
Added named tuples, might wanna gloss over it

msullivan:
Oh, clean up the unused imports please

msullivan:
Do the tuples and array support trailing commas? They should, and we should make sure to test it

src/prompt/variable.rs
```diff
@@ -505,6 +510,143 @@ impl VariableInput for Tuple {
...
+    pub element_types: HashMap<String, Arc<dyn VariableInput>>,
+    pub shape: NamedTupleShape
+}
+
+pub struct NamedTupleParser<'a> {
+    tuple: &'a NamedTuple,
+}
+
+impl Parser<&str, Value, ParsingError> for NamedTupleParser<'_> {
+    fn parse<'a>(&mut self, mut input: &'a str) -> IResult<&'a str, Value, ParsingError> {
```
msullivan, line 524: I don't feel strongly about this, but if you want to try doing named tuples in a more "combinator-y" way, I think you could use https://docs.rs/nom/latest/nom/branch/fn.permutation.html

msullivan: Or maybe https://docs.rs/nom/latest/nom/combinator/fn.flat_map.html

msullivan: Oh, `permutation` doesn't work, for a couple reasons, the most insurmountable being that it takes a tuple, not any sort of list. I think doing a [separated_list0](https://docs.rs/nom/latest/nom/multi/fn.separated_list0.html) where the elements are [flat_map](https://docs.rs/nom/latest/nom/combinator/fn.flat_map.html)s that consume the name and map it to the right type parser could work.

The funny thing is that it might actually be easier to do a "pure combinator" approach for named tuples than for regular tuples. For regular tuples, I think we probably could do it without needing to manually implement any combinators, but I think we would need to have an explicit loop where we compose combinators. I guess that's still "pure combinator", though.

quinchs:
> Do the tuples and array support trailing commas?

No, any extra comma errors. I'll change that

msullivan:
Hm, actually now that I think about it more, maybe we should require the named tuples to be in order, also? That matches edgeql better (where the order is important), though (for example) the Python bindings will accept dictionaries and namedtuples without caring about order. @elprans thoughts on this?

We should probably also support writing non-named tuples when a named tuple is expected also, which I now realize is what @elprans was probably saying earlier

msullivan:
Oh, and *leading* spaces after an opening `(` or `[`

quinchs:
> We should probably also support writing non-named tuples when a named tuple is expected

this will be exclusive right? ex: cant have 1 named and 1 unnamed part; its either all named or no named?

msullivan:
> > We should probably also support writing non-named tuples when a named tuple is expected
> 
> this will be exclusive right? ex: cant have 1 named and 1 unnamed part; its either all named or no named?

Yeah

src/prompt/variable.rs
```diff
@@ -16,31 +31,188 @@ use rustyline::hint::Hinter;
...
+    fn parse<'a>(&self, input: &'a str, flags: InputFlags) -> ParseResult<'a>;
+}
+
+fn white_space<'a, O, E: ParseError<&'a str>, F: Parser<&'a str, O, E>>(
+    f: F,
+) -> impl Parser<&'a str, O, E> {
+    delimited(multispace0, f, multispace0)
+}
+
+fn space(i: &str) -> IResult<&str, &str, ParsingError> {
```
msullivan, line 124: This is the same as multispace0, right?

src/prompt/variable.rs
```diff
@@ -146,27 +370,289 @@ pub struct Json;
...
+                                    format!(
+                                        "Expecting one of the following element name(s): {}",
+                                        self.shape.elements.iter()
+                                            .filter(|v| !result.iter().any(|t| t.0 == v.name.as_str()))
+                                            .map(|v| v.name.clone())
+                                            .collect::<Vec<String>>()
+                                            .join(", ")
+                                    )
+                                )
+                            })
```
msullivan, line 604: I think we need to handle the case with duplicate field names also

msullivan:
I just pushed commits that support trailing commas for tuples, named tuples, and arrays. Mostly I did this as a personal experiment with `nom`.

The combinator-y tuple that works by building up a new boxed parser isn't too bad. It took some fiddling because I'm not that good at rust anymore, but is basically reasonable, I think.

The array/namedtuple version I found very frustrating. I implemented a function `trailing_separated_list0`, with the intention that it behave like `separated_list0` but allow trailing separators. This worked but I wasn't able to figure out a way to do it purely by composing combinators, and needed to invoke the parsers directly, which cuts against the benefit of parser combinators somewhat. The final result isn't actually that bad, though, just unsatisfying.

aljazerzen:
@quinchs what's the status of this?

quinchs:
> @quinchs what's the status of this?

I've got some remaining changes to fix with it which sully has mentioned.

PR #1203, opened 2024-02-06
Order schema files before reading them

diksipav:


diksipav:
I tried to use filter_map and collect instead iterator but couldn't make it work (still need time understanding all the traits and ways how to solve things). Maybe I can make it better in the future.

Also, I am not 100% sure how to test different order of schema files in the filesystem, I tried to google how mac does this but no success.

src/migrations/create.rs
```diff
@@ -945,3 +949,21 @@ fn add_newline() {
     assert_eq!(wrapper("(1 + 7) #xx"), "(1 + 7) #xx\n");
     assert_eq!(wrapper("(1 #one\n + 3 #three\n)"), "(1 #one\n + 3 #three\n)");
 }
+
+#[tokio::test]
+ async fn start_migration() {
+    let schema_dir = PathBuf::from("/Users/dijanapavlovic/Documents/projects/edgedb/edgedb-cli/tests/migrations/db5");
```
msullivan, line 955: This won't work on anyone else's machine (and is probably why the tests fail)

msullivan (commented):
Code change looks good!
Test needs to be fixed

src/migrations/create.rs
diksipav: ahaha omg yes :D will fix it :D

msullivan:
> I tried to use filter_map and collect instead iterator but couldn't make it work (still need time understanding all the traits and ways how to solve things). Maybe I can make it better in the future.
> 
Your version seems perfectly fine.

> Also, I am not 100% sure how to test different order of schema files in the filesystem, I tried to google how mac does this but no success.
>
There is probably no way to do this. What *could* maybe be done is making an "abstract" file system interface that would allow us to write unit tests that specify the files in different orders. I do not think it would be worth your time to do that.

msullivan (approved):
Ha, one test failing because of irritating windows line ending issues. I'd just replace all `\r\n`s with `\n` before doing the comparision, I guess.

(It seems like rust (unlike C and python) doesn't have APIs for reading "in text mode" that will convert windows line endings to normal ones)

diksipav:
> Ha, one test failing because of irritating windows line ending issues. I'd just replace all `\r\n`s with `\n` before doing the comparision, I guess.
> 
> (It seems like rust (unlike C and python) doesn't have APIs for reading "in text mode" that will convert windows line endings to normal ones)

Hey tnx, I was waiting last night for all tests to finish but felt asleep in the meantime. It takes quite some time to run tests.

PR #1205, opened 2024-02-06
Use pricing API to populate instance resource defaults

elprans:
Use `/pricing` for instance resource defaults instead of hardcoding
them.

While here, fix handling of tier transitions (free -> pro) and handle
simultaneous resizes of compute and storage (by sending resize requests
in sequence).

### DIFF
--- a/src/cloud/ops.rs
+++ b/src/cloud/ops.rs
@@ -1,4 +1,6 @@
 use std::time::{Duration, Instant};
+use std::collections::HashMap;
+use std::fmt;
 
 use anyhow::Context;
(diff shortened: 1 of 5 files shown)


src/cloud/ops.rs
```diff
@@ -96,19 +98,49 @@ pub struct Version {
     pub version: String,
 }
 
+#[derive(Debug, serde::Deserialize)]
+pub struct Price {
+	pub billable: String,
```
zackelan, line 103: indentation is off here

zackelan (approved):

PR #1214, opened 2024-02-19
Bump `env_logger` and `clicolors-control`

elprans:
Removes transitives dependencies on unmaintained `atty`.


fantix (approved):

aljazerzen:
This is not needed anymore, since env_logger does not depend on atty anymore.

PR #1224, opened 2024-03-01
Move the custom build.rs into a separate crate

aljazerzen:
This enables this crate to be built (but not fully tested) without cloning the git submodule.

### DIFF
--- a/tests/shared-client-tests/Cargo.toml
+++ b/tests/shared-client-tests/Cargo.toml
@@ -1,0 +1,32 @@
+[package]
+name = "shared-client-tests"
+license = "MIT/Apache-2.0"
+version = "0.1.0"
+authors = ["EdgeDB Inc. <hello@edgedb.com>"]
+edition = "2018"
+
+[features]
+portable_tests = []
+
+[build-dependencies]
+serde_json = "1.0"
+
+[lib]
+path = "lib.rs"
+test = false
+doctest = false
+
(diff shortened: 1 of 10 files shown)


tests/shared-client-tests/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+serde_json = "1.0"
+edgedb-protocol = { git = "https://github.com/edgedb/edgedb-rust/", features = [
+    "all-types",
+] }
+sha1 = "0.10.1"
+# we need slicing of indexmap, so need version 2.0
+indexmap = { git = "https://github.com/bluss/indexmap", rev = "11ac52c", features = [
+    "serde",
+] }
+hex = { version = "0.4.3", features = ["serde"] }
```
aljazerzen, line 32: This is the main downside: we have to re-declare all needed dependencies.

fantix: Perhaps we can move dependencies to `[workspace.dependencies]` in `$PROJECT_ROOT/Cargo.toml`, and use workspace dep-ref like `hex = { workspace = true }` here to mitigate the version redeclaration.

elprans: I would prefer we do this. Adding/removing deps is far less frequent than bumping versions.

quinchs (approved):

PR #1225, opened 2024-03-04
Branching: Rebase

quinchs:
## Summary
This PR adds the `branch rebase` command to the CLI, which will preform the necessary steps outlined in the [RFC](https://github.com/edgedb/rfcs/blob/master/text/1025-branches.rst) to rebase one branch ontop of another.

### Outline of flow
In short, this is what the CLI does to preform a rebase with the following:
```bash
> edgedb branch switch feature
> edgedb branch rebase target
```
1. The CLI creates a new branch `TEMP` thats derived from `target` (`create data branch TEMP from target`)
2. The diverging migration history is calculated by introspecting the `feature` branch and the `target` branch and finding all migrations that are apart of `feature` but not apart of `target`[^1]
3. The new migration history is written to disc and the IDs of the migrations are updated to match the new hashes.
4. The new migration history is applied to the `TEMP` branch.
5. The `feature` branch is dropped and `TEMP` is renamed to feature.

[^1]: The way this works is we find the oldest migration in `feature` that isn't apart of `target` and create a slice at that migration, taking everything on the right-side of that splice, and appending it to the migrations of `target`.


aljazerzen (approved):
I haven't reviewed this fully, but after a quick skim it looks ok.

Given the circumstances, let's merge this now so it can be released in the nightly build and we'll open followup PRs if needed.

src/migrations/branch.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    fn statements<'a>(&'a self) -> Self::StatementsIter<'a> {
+        std::iter::once(&self.migration.script)
+    }
+}
+
+pub struct RebaseMigrations {
+    source_migrations: IndexMap<String, DBMigration>,
+    target_migrations: IndexMap<String, DBMigration>,
+}
```
aljazerzen, line 56: Do I understand this correctly:

`source_migrations` contains all migrations that are on the source branch,
`target_migrations` contains migrations that are on the target branch, but not on the source branch

---

If this is so, I would suggest to refactor this into:
```
pub struct RebaseMigrations {
    /// Migrations initial..base
    base_migrations: IndexMap<String, DBMigration>,
    /// Migrations base..source
    source_migrations: IndexMap<String, DBMigration>,
    /// Migrations base..target
    target_migrations: IndexMap<String, DBMigration>,
}
```
This structure would be useful in `.flatten()`.

src/migrations/branch.rs
```diff
@@ -0,0 +1,216 @@
...
+    fn flatten(&self) -> anyhow::Result<Vec<RebaseMigration>> {
+        let mut result = Vec::new();
+
+        let mut key_index = 0;
+        let last_source_migration_id = self.source_migrations.last().map(|v| v.0.as_str()).unwrap_or("initial");
+        let mut iter = self.source_migrations.iter();
+
+        loop {
+            match iter.next() {
+                Some((id, migration)) => {
```
aljazerzen, line 68: Ditto, this is a `for` loop

src/migrations/branch.rs
```diff
@@ -0,0 +1,216 @@
+use std::iter::FromIterator;
```
aljazerzen, line 1: This file should probably be named `rebase`, no?

src/branch/option.rs
```diff
@@ -62,3 +63,8 @@ pub struct Rename {
 }
 #[derive(clap::Args, Debug, Clone)]
 pub struct List {}
+
+#[derive(clap::Args, Debug, Clone)]
```
aljazerzen, line 67: Add comment:

```
/// Creates a new branch that is based on the target branch, but also
/// contains any new migrations on the current branch.
/// Warning: data stored in current branch will be deleted.
```

aljazerzen (commented):
This is a solid PR, good work. I've left a few comments here and a suggestion on improving DX in #1230

Cargo.toml
```diff
@@ -7,7 +7,7 @@ members = [
 [package]
 name = "edgedb-cli"
 license = "MIT/Apache-2.0"
-version = "4.2.0-dev"
+version = "4.2.1-dev"
```
aljazerzen, line 10: This shouldn't have been changed, as we have not released a new version.

Issue #1230, opened 2024-03-06
branch rebase printing output could be improved


aljazerzen:
Currently, branch rebase output looks like this:

```
➜  demo10 git:(feat-links) edgedb branch rebase main
OK: CREATE BRANCH
Rebasing m1x56iv2ej7twarsf7sfs5ydornaawo2uvwj6xzzwyj6qakodd2mrq as base
Rebasing m1ce4fakd5rmb5gnxba4mmznjjyxcivekktgmtm3dxxbmy7hbq5x5q as base
Rebasing m1jg66zlozm4kt74rrnofcpwwiw7e43qnwpl5adj6mgv5t3xml3yja as base
Rebasing m1cl5vqruxmatwlbjyrtougnb6t7ze65ssuvkp6553iey4dbma3zqq as feature onto m1jg66zlozm4kt74rrnofcpwwiw7e43qnwpl5adj6mgv5t3xml3yja
Updating migration m1cl5vqruxmatwlbjyrtougnb6t7ze65ssuvkp6553iey4dbma3zqq to m14hkktv5sghhmvv5xn2db7rgkdtw2w7vev42prfr6n5hbda7i2pfa
Writing: /home/aljaz/EdgeDB/demo10/dbschema/migrations/00004.edgeql
Writing: /home/aljaz/EdgeDB/demo10/dbschema/migrations/00003.edgeql
Writing: /home/aljaz/EdgeDB/demo10/dbschema/migrations/00002.edgeql
Writing: /home/aljaz/EdgeDB/demo10/dbschema/migrations/00001.edgeql
Applied m14hkktv5sghhmvv5xn2db7rgkdtw2w7vev42prfr6n5hbda7i2pfa (00004.edgeql)
OK: DROP BRANCH
Recreating branch feat-links...
OK: ALTER BRANCH
```

My nits:
- the `OK: CREATE BRANCH` and `OK: DROP BRANCH` could be more clear in what they are doing.
- `Rebasing m1x56iv2ej7twarsf7sfs5ydornaawo2uvwj6xzzwyj6qakodd2mrq as base` is not necessary to show
- `Updating migration m1cl5vqruxmatwlbjyrtougnb6t7ze65ssuvkp6553iey4dbma3zqq to m14hkktv5sghhmvv5xn2db7rgkdtw2w7vev42prfr6n5hbda7i2pfa` is very verbose, is it even necessary?
- `Writing` should only be printed for new migrations, and not any common migrations.

[...]

PR #1231, opened 2024-03-06
Branching: Feedback items from #1225

quinchs:


quinchs:
Also addresses https://github.com/edgedb/edgedb-cli/issues/1230

src/migrations/rebase.rs
```diff
@@ -0,0 +1,286 @@
...
+
+pub async fn do_rebase(connection: &mut Connection, context: &Context, rebase_migrations: RebaseMigrations) -> anyhow::Result<()> {
+    let temp_dir = tempfile::tempdir()?;
+    let temp_ctx = Context {
+        schema_dir: temp_dir.path().to_path_buf(),
+        edgedb_version: None,
+        quiet: false,
+    };
+
+    // write all the migrations to disc.
```
aljazerzen, line 227:
```suggestion
    // write all the migrations to disk.
```

src/migrations/rebase.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    // move the new migrations from temp to schema dir
+    let mut last: Option<&RebaseMigrationKind> = None;
+    for (index, from) in migration::read_names(&temp_ctx).await?.iter().enumerate() {
+        let rebase_migration = flattened_migrations.get(index);
+        let to = context
+            .schema_dir
+            .join("migrations")
+            .join(from.file_name().expect(""));
```
aljazerzen, line 249: `.expect("")` will produce confusing panics.

```suggestion
            .join(from.file_name().unwrap());
```

I think that it is safe to assume that `from` will have a file name, because this temporary dir will only contain files that we have written, which will never be a directory.

src/migrations/rebase.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        last = rebase_migration.map(|v| &v.kind);
+    }
+
+    // apply the new migrations
+    let migrations: IndexMap<String, MigrationFile> = migration::read_all(context, true).await?.into_iter()
+        .filter(|(id, _)| rebase_migrations.target_migrations.contains_key(id))
+        .collect();
+
+    migrate::apply_migrations(connection, &migrations, context, true).await?;
```
aljazerzen, line 283: Just a suggestion: it would make sense to have this chunk moved into `branch::rebase::rebase` function and rename this function to something like `write_rebased_migration_files`.

src/migrations/rebase.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        let mut next_key = || -> MigrationKey {
+            key_index += 1;
+            MigrationKey::Index(key_index)
+        };
+
+        for (_, migration) in &self.base_migrations {
+            result.push(RebaseMigration {
+                key: next_key(),
+                parent: None,
```
aljazerzen, line 99: This is not correct, because you are iterating over all `base_migrations`.

```suggestion
                parent: result.last().map(|v| v.migration.name.as_str()),
```

src/migrations/rebase.rs
```diff
@@ -0,0 +1,286 @@
...
+    fn parent(&self) -> anyhow::Result<&str> {
+        if let Some(parent) = self.parent {
+            return Ok(parent)
+        }
+
+        if self.migration.parent_names.is_empty() {
+            return Ok("initial");
+        }
+
+        Ok(&self.migration.parent_names[0])
```
aljazerzen, line 50: This shouldn't be needed. We are iterating over all migrations anyway, so our parent should be set correctly.

If it is not, that is a bug.

src/migrations/rebase.rs
```diff
@@ -0,0 +1,286 @@
...
+                parent: base_last.map(|v| v.migration.name.as_str()),
+                kind: RebaseMigrationKind::Source,
+            });
+
+
+            for (_, migration) in &self.source_migrations[1..] {
+                result.push(RebaseMigration {
+                    key: next_key(),
+                    migration,
+                    parent: None,
```
aljazerzen, line 121:
```suggestion
                    parent: result.last().map(|v| v.migration.name.as_str()),
```

src/migrations/rebase.rs
```diff
@@ -0,0 +1,286 @@
...
+        if self.source_migrations.len() > 0 {
+            let (_, first_migration) = self.source_migrations.first().unwrap();
+            let base_last = result.last();
+
+            result.push(RebaseMigration {
+                key: next_key(),
+                migration: first_migration,
+                parent: base_last.map(|v| v.migration.name.as_str()),
+                kind: RebaseMigrationKind::Source,
+            });
```
aljazerzen, line 114: Is there a reason for special-casing the first migration?

aljazerzen (approved):
This is great.

`src/migrations/migrate.rs` is missing two `parent_id_range: (0, 0),`

Also:
```
cargo clippy --fix -- --allow=clippy::all
````

src/migrations/rebase.rs
quinchs: If `parent` is `None` on `RebaseMigration`, the `DBMigration`s' `parent_names` field is used, https://github.com/edgedb/edgedb-cli/blob/master/src/migrations/branch.rs#L32-L42

quinchs: which yields correct behavior for `base_migrations`

src/migrations/rebase.rs
quinchs: Yes, since the 3 sets need to be "joined" by the parent id for the rebase to work properly, thats what `parent` override does

src/migrations/rebase.rs
aljazerzen: But don't we iterate over all migrations, so we could have all `.parent`s set correctly, even for `base_migrations`?

It seems like unnecessary complication, but you are right, it does fallback correctly.

quinchs: > But don't we iterate over all migrations

In `get_diverging_migrations`?

src/migrations/rebase.rs
aljazerzen: Ahh ok, I see now. It is `parent_override`! Can you rename it to `parent_override` (only if you decide against my other comment about setting all `parent`s)

src/migrations/rebase.rs
quinchs: That requires `migrate` and `migration` become public modules, is that okay?

quinchs: Oh nvm, I see what you mean 💯

PR #1233, opened 2024-03-06
Branching: Check if branch exists before trying to create it

quinchs:


aljazerzen (commented):
This roundtrip is unnecessary, I think the error that the server returns is good enough.

I'm sorry I threw you off with my bug report, I though some other error was caused by an existing branch.

PR #1236, opened 2024-03-07
Branching: Create no longer switches; Switch has --create

quinchs:


src/branch/switch.rs
```diff
@@ -16,7 +17,12 @@ pub async fn main(
         .await?;
```
aljazerzen, line 18: Just remembered: let's also check that we are not already on this branch, so you don't get messages like:

```
Switching from `hello` to `hello`
```

aljazerzen (approved):

PR #1238, opened 2024-03-08
Branching: fix migration order and swap target and source connections

aljazerzen:
Edit: I didn't see you had something similar up in #1235

This PR does two things:
- sorts the list of migration files that are read from the temp dir, so they are printed in order
- swaps target and source connections - I think you had this double swapped so it worked, but printed incorrectly.

One thing I noticed was that we definitively need tests for this.


quinchs (approved):

quinchs:
I'll close https://github.com/edgedb/edgedb-cli/pull/1235 and reopen with the migration fix

PR #1240, opened 2024-03-08
Branches: Fix migration file calculation

quinchs:
Since `rebase_migrations` were never updated by `rebase_migration_ids`, the check for what migration files to apply would always produce incorrect results since the the ID of the migration file didn't match anything within `rebase_migration`. This PR fixes that by updating `rebase_migrations` in `rebase_migration_ids`


aljazerzen (approved):
Make more sense yes.

Hey, why are we even writing migrations into the temporary directory? Wouldn't it be easier to read them from the database, flatten, replace ids (in memory) and then write directly into the real migration dir?

quinchs:
> Hey, why are we even writing migrations into the temporary directory?

Migration IDs are based off of content hashes, and currently there's no in-memory way to generate them without also writing them.

I could refactor this to do the ID stuff in-memory, but I mean both solutions do the same thing.

aljazerzen:
I see. I'm thinking about what happens on a project with 500+ migration files and a HDD.

PR #1243, opened 2024-03-08
Don't print server backtraces when database doesn't exist

### LINKED ISSUE #1218: Don't print a server traceback on UnknownDatabaseError
We should maybe only print server tracebacks for InternalServerErrors and nothing else.
But we definitely shouldn't print it for UnknownDatabaseError

msullivan:
We were just always printing a stack trace for top level errors. No
reason to do that; we should only print stack traces on ISEs
typically.

Fixes #1218.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -67,7 +67,7 @@ fn main() {
                 err = arc.inner();
             }
             if let Some(e) = err.downcast_ref::<edgedb_errors::Error>() {
-                print::edgedb_error(e, true);
+                print::edgedb_error(e, false);
             } else {
                 print::error(err);
             }


src/main.rs
```diff
@@ -67,7 +67,7 @@ fn main() {
                 err = arc.inner();
             }
             if let Some(e) = err.downcast_ref::<edgedb_errors::Error>() {
-                print::edgedb_error(e, true);
+                print::edgedb_error(e, false);
```
msullivan, line 70: This is the "verbose" argument.

Also, I consider boolean arguments to functions to be an anti-pattern in languages without named arguments. We should probably make an enum for opting in to Verbose.

quinchs (approved):

PR #1244, opened 2024-03-08
Branches: switch doesn't rely on old branch to exist

### LINKED ISSUE #1241: Unable to switch branches with `edgedb branch switch` after force dropping the current branch
Trying to switch branches after force-dropping the current branch produces an error:
```
[2024-03-08T17:20:52Z WARN  edgedb_tokio::raw::connection] Error received from server: database <branch-name> does not exist. Severity: Error. Code: 0x4030005
edgedb error: ClientConnectionFailedError: end of stream while reading message
```

This error is also produced when running other branch commands like `list`.

Workaround: Switch branches while passing a still existing branch for -b. `edgedb -b edgedb branch switch edgedb`

- EdgeDB Version: 5.0-beta.1+8477d12
- EdgeDB CLI Version: 4.2.1-dev.1092+b2a234f
- OS Version: macOS 14.0

Steps to Reproduce:

1. Create a branch
2. Switch to the branch (done as part of create in current nightly)
3. Get a REPL to a different branch, for instance with `edgedb -b edgedb`
4. Drop the new/current branch in that REPL with `drop branch <branch-name> force;`
5. Try switching the current branch with `edgedb branch switch edgedb`

quinchs:
Fixes #1241

### DIFF
--- a/src/branch/connections.rs
+++ b/src/branch/connections.rs
@@ -33,7 +33,10 @@ impl BranchConnection<'_> {
     }
 }
 
-pub async fn try_connect(connector: &Connector) -> anyhow::Result<Option<Connection>> {
+/// Attempts to connect a provided connection, returning `Ok(Some)` if the connection was
+/// established, `Ok(None)` if the connection couldn't be established because of a
+/// `ClientConnectionFailedError`, and `Err` if any other type of error occurred.
+pub async fn connect_if_branch_exists(connector: &Connector) -> anyhow::Result<Option<Connection>> {
     match connector.connect().await {
         Ok(c) => Ok(Some(c)),
         Err(e) => {
--- a/src/branch/connections.rs
+++ b/src/branch/connections.rs
@@ -1,6 +1,7 @@
 use std::ops::Deref;
+use edgedb_tokio::Error;
 use uuid::Uuid;
-use crate::connect::Connection;
+use crate::connect::{Connection, Connector};
 use crate::options::Options;
 use crate::print;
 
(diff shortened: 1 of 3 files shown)


src/branch/connections.rs
```diff
@@ -32,6 +33,23 @@ impl BranchConnection<'_> {
     }
 }
 
+pub async fn try_connect(connector: &Connector) -> anyhow::Result<Option<Connection>> {
```
aljazerzen, line 36: Please document what does `Ok(None)` response here mean. That the database does not exists?

If this is the case, I think the naming convention would be `connect_if_branch_exists`. `try_connect` is used for stuff that returns a `Result` (i.e. `into()` vs `try_into()`).

src/branch/switch.rs
```diff
@@ -1,39 +1,56 @@
...
+            }
         }
+    } else {
+        // try to connect to the target branch
+        let target_branch_connector = connector.database(&options.branch)?;
+        match try_connect(&target_branch_connector).await? {
+            Some(mut connection) => {
+                verify_server_can_use_branches(&mut connection).await?;
+            },
+            None => anyhow::bail!("The target branch doesn't exist, and cannot be created because the current branch cannot be connected to.")
```
aljazerzen, line 44: > cannot be created

If we are not trying to create the branch, we should just print:

> "The target branch doesn't exist"

src/branch/switch.rs
```diff
@@ -1,39 +1,56 @@
...
     eprintln!(
         "Switching from '{}' to '{}'",
         context.branch, options.branch
     );
+
     context.update_branch(&options.branch).await?;
 
     Ok(())
-}
+}
```
aljazerzen, line 56: newline

aljazerzen (approved):
Looks good.

fantix (approved):

Issue #1228, opened 2024-03-06
Branching: cannot rename current branch


aljazerzen:
Tracked by: #1204 

Probably we will have to use some other branch (or create a temporary one) and issue the ALTER via that one instead.

PR #1245, opened 2024-03-11
Branches: Allow renaming of current branch

quinchs:
Fixes https://github.com/edgedb/edgedb-cli/issues/1228

### DIFF
--- a/src/branch/rename.rs
+++ b/src/branch/rename.rs
@@ -1,14 +1,19 @@
+use crate::branch::connections::get_connection_to_modify;
 use crate::branch::context::Context;
 use crate::branch::option::Rename;
+use crate::options::Options;
 use crate::connect::Connection;
 use crate::print;
 
 pub async fn main(
     options: &Rename,
-    _context: &Context,
+    context: &Context,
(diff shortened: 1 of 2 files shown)


src/branch/rename.rs
```diff
@@ -1,14 +1,19 @@
...
 
 pub async fn main(
     options: &Rename,
-    _context: &Context,
+    context: &Context,
     connection: &mut Connection,
+    cli_opts: &Options,
 ) -> anyhow::Result<()> {
-    let status = connection
+    let mut modify_connection = get_connection_to_modify(&context.branch, &cli_opts, connection).await?;
```
aljazerzen, line 14: In the case when current connection is not on the source branch, `get_connection_to_modify` should not open any new connections or even do any new roundtrips.

So what we need is something like this:

```
fn get_connection_to_modify(branch, ..., connection) -> ... {
    if connection.branch != branch {
        return branch;
    }
    ...
}
```

aljazerzen (approved):

fantix (approved):

PR #1253, opened 2024-03-15
Branches: Merge

quinchs:


src/migrations/merge.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub target_migrations: IndexMap<String, MergeMigration>
+}
+
+impl MergeMigrations {
+    fn flatten(&self) -> IndexMap<&String, &MergeMigration> {
+        let mut result = IndexMap::new();
+
+        for migration in self.base_migrations.iter() {
+            result.insert(migration.0, migration.1);
+        }
```
aljazerzen, line 22:
```suggestion
        result.insert(self.base_migrations);
```
(I'm not sure this will work, but it would with normal `Map` or `Vec`)

src/migrations/merge.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        for migration in self.target_migrations.iter() {
+            result.insert(migration.0, migration.1);
+        }
+
+        return result;
```
aljazerzen, line 28:
```suggestion
        result
```

src/migrations/merge.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    for (index, (base_migration_id, _)) in base_migrations.iter().enumerate() {
+        if let Some((target_migration_id, _)) = target_migrations.get_index(index) {
+            if target_migration_id != base_migration_id {
+                anyhow::bail!("Migration histories of base and target diverge");
+            }
+        } else {
+            anyhow::bail!("Target branch doesn't contain the migration {} (at index {})", base_migration_id, index)
+        }
+    }
```
aljazerzen, line 75: These error messages do explain what's going on in our merge algorithm, but I don't think people who know git would find them useful.

What I would expect is only the following error message:
```
Cannot fast-forward.
Current branch contains migrations that are not target branch:
- 00003.edgeql
- 00004.edgeql
- 00005.edgeql

Hint: you must rebase the target branch onto current branch first:
    edgedb branch switch {target-branch}
    edgedb branch rebase {current-branch}
    edgedb branch switch {current-branch}
    edgedb branch merge {target-branch}
```

One way to do this is to:
- first find the merge base, same as you have done with rebase.
- check that `merge_base == current_branch`, otherwise throw the error above,
- `target_migrations = merge_base..target_branch`

Note that if there is no common migrations, merge base is `initial`. This means that all migrations on target branch are `target_migrations` and that if there are any migrations on current branch, we throw the error above.

src/migrations/merge.rs
quinchs: seems to work fine? why would it not normally work?

quinchs: No it doesn't seem to work with `IndexMap`

src/migrations/merge.rs
quinchs: Have a look at the new error messages

src/migrations/merge.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    if target_missing.len() > 0 {
+        eprintln!("\nThe migration history of {} is missing the following migrations", target.database());
+
+        for (index, missing) in &target_missing {
+            eprintln!("{}. {}", index, &missing[..7])
+        }
+    }
```
aljazerzen, line 95: Hmm, in what case will this happen?

src/migrations/merge.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+pub async fn get_merge_migrations(base: &mut Connection, target: &mut Connection) -> anyhow::Result<MergeMigrations> {
+    let base_migrations = read_all(base, true, false).await?;
+    let mut target_migrations = read_all(target, true, false).await?;
+
+    if base_migrations.len() > target_migrations.len() {
+        anyhow::bail!("Source branch contains more migrations than the target branch");
+    }
```
aljazerzen, line 66: Can we drop this special message? I goes against my mental model of how merge works.

src/migrations/merge.rs
quinchs: This happens when the target branch doesn't contain a specific migration; ex:
**Base Branch**
- A
- B
- C

**Feature**
- A
- B

The message will appear with missing migration of C

src/migrations/merge.rs
quinchs: Should the if case be removed entirely? or should the message just change

src/migrations/merge.rs
aljazerzen: Well, this is ok - it's a no-op: there is nothing from feature branch to merge into base. Git says:

```
➜  edgedb git:(refactor-parser-context) git merge HEAD~1
Already up to date.
```

src/migrations/merge.rs
aljazerzen: I'd say remove the whole if

aljazerzen (approved):

Issue #1250, opened 2024-03-13
We should populate the prompt with `sys::get_current_database()`


msullivan:
Currently:
```
_localdev:edgedb> select sys::get_current_database();
{'main'}
```

It would be better if it said `_localdev:main> `

1st1:
Yes, with branches it's now essential

quinchs:
Seems like this is already in the CLI:
```bash
> edgedb -I test -b test_branch
EdgeDB 5.0-dev.8154+3acd9a4 (repl 4.3.0-dev.1110+93964fb)
Type \help for help, \quit to quit.
test:test_branch>

...

> edgedb -I test -d test_branch
database connection argument is deprecated in favor of 'branch'
EdgeDB 5.0-dev.8154+3acd9a4 (repl 4.3.0-dev.1110+93964fb)
Type \help for help, \quit to quit.
test:test_branch>
```

msullivan:
@quinchs The point is that the *default* branch should be correct, whether it is `edgedb` or `main`, even if the user didn't specify it

PR #1258, opened 2024-03-22
Print current branch/database in REPL

quinchs:
Fixes https://github.com/edgedb/edgedb-cli/issues/1250

### DIFF
--- a/src/interactive.rs
+++ b/src/interactive.rs
@@ -136,6 +136,7 @@ pub fn main(options: Options, cfg: Config) -> Result<(), anyhow::Error> {
         initial_text: "".into(),
         edgeql_state_desc: RawTypedesc::uninitialized(),
         edgeql_state: State::empty(),
+        current_database: None,
     };
     let runtime = tokio::runtime::Builder::new_multi_thread()
         .enable_all()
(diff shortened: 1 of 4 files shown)


aljazerzen (approved):
This basically doubles the time between user hitting enter and CLI being ready for next command.

But there is no other way, so I guess ok.

msullivan:
> This basically doubles the time between user hitting enter and CLI being ready for next command.
> 
> But there is no other way, so I guess ok.

Can't we fetch it only once at connection time?

aljazerzen:
Oh, yes - branch cannot be changed with EdgeQL queries, so it can be cached for the whole duration of a connection.

That said: we could just the the "branch" connection parameter and to make any requests to the database. I guess this would be more fragile since we both "database" param and "branch" param and some non-trivial logic deciding which branch we actually open the connection to.

Bottom line: current approach is good

aljazerzen:
The interactive tests are failing because they expect to have "edgedb>" command prompt, but get "main>" instead.

quinchs:
> The interactive tests are failing because they expect to have "edgedb>" command prompt, but get "main>" instead.

I'll fix that

Issue #1257, opened 2024-03-21
`edgedb branch` commands don't work from outside a project at all


msullivan:
```
$ edgedb branch wipe main -I _localdev
error: unexpected argument '-I' found
```
(`edgedb database wipe -I _localdev` works)

```
$ edgedb branch wipe main
thread 'main' panicked at src/branch/main.rs:37:58:
Missing project
stack backtrace:
   0: rust_begin_unwind
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:72:14
   2: core::panicking::panic_display
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:178:5
   3: core::panicking::panic_str
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:152:5
   4: core::option::expect_failed
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/option.rs:1985:5
   5: core::option::Option<T>::expect
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/option.rs:894:21
   6: edgedb::branch::main::create_context::{{closure}}
             at /home/sully/.cargo/git/checkouts/edgedb-cli-6a9c62fe43dd9873/7af21d2/src/branch/main.rs:37:23
   7: edgedb::branch::main::branch_main::{{closure}}
             at /home/sully/.cargo/git/checkouts/edgedb-cli-6a9c62fe43dd9873/7af21d2/src/branch/main.rs:11:36
```

```
$ edgedb -I _localdev branch wipe main
thread 'main' panicked at src/branch/main.rs:37:58:
Missing project
stack backtrace:
```

I understand that `switch` obviously needs a project, but `create`, `drop`, `wipe`, `rename`, and `list` must work without a project, using a specified instance. (The database commands do still work, but they whine at you.)

PR #1259, opened 2024-03-22
Branches: dont require projects unless explicitly needed

quinchs:
Fixes https://github.com/edgedb/edgedb-cli/issues/1257

### DIFF
--- a/src/branch/context.rs
+++ b/src/branch/context.rs
@@ -8,7 +8,6 @@ use crate::portable::project::{instance_name, stash_path};
 
 pub struct Context {
     pub project_config: Option<Config>,
-    pub project_branch: Option<String>,
     pub branch: Option<String>,
 
     project_dir: Option<PathBuf>,
@@ -21,8 +20,6 @@ impl Context {
             None => None
         };
 
-        let project_branch = project_config.as_ref().map(|v| v.edgedb.branch.clone());
-
         let credentials = match project_dir {
             Some(path) => Some(credentials::read(&credentials::path(&get_instance_name(&path)?)?).await?),
             None => None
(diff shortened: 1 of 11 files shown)


src/branch/context.rs
```diff
@@ -1,41 +1,73 @@
...
 use crate::credentials;
 use crate::portable::options::InstanceName;
 use crate::portable::project::{instance_name, stash_path};
 
 pub struct Context {
-    pub project_config: Config,
-    pub project_branch: String,
-    pub branch: String,
+    pub project_config: Option<Config>,
+    pub project_branch: Option<String>,
```
aljazerzen, line 11: I though that we abandoned the concept of "projects having branches" at all.

Right now, I see that we still read "edgedb.toml", property "edgedb.branch" and default to that when "credentials.json" doesn't contain the "database".

aljazerzen (approved):

src/branch/context.rs
```diff
@@ -1,41 +1,73 @@
...
-    pub async fn new(project_dir: &Path) -> anyhow::Result<Context> {
-        let project_config = crate::portable::config::read(&project_dir.join("edgedb.toml"))?;
-        let project_branch = project_config.edgedb.branch.clone();
+    pub async fn get_branch_or_default(&self, connection: &mut Connection) -> anyhow::Result<String> {
+        if self.branch.is_some() {
+            return Ok(self.branch.as_ref().unwrap().clone());
+        }
+
+        return Ok(self.get_default_branch_name(connection).await?.to_string())
+    }
```
aljazerzen, line 24: This function is unused. Is this intentional?

src/branch/context.rs
quinchs: IIRC we're still keeping that, I'll double check on this

PR #1268, opened 2024-03-28
Add info for install path

scotttrinh:
This allows getting both the containing path to the currently executing binary under `binary-dir`, but also the path that the self-installer would install the binary to under `install-dir`.

There are some slight naming confusions happening here since we were calling the install path the "binary path". Open to doing a naming refactor here to clean that up.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -415,6 +415,8 @@ pub struct UI {
 #[derive(clap::Args, Debug, Clone)]
 pub struct Info {
    #[arg(long, value_parser=[
+        "install-dir",
+        "binary-dir",
         "config-dir",
         "cache-dir",
         "data-dir",
@@ -422,6 +424,8 @@ pub struct Info {
     ])]
     /// Get specific value:
     ///
+    /// * `install-dir` -- Directory where EdgeDB CLI is installed
+    /// * `binary-dir` -- Directory where the current binary is located
     /// * `config-dir` -- Base configuration directory
     /// * `cache-dir` -- Base cache directory
     /// * `data-dir` -- Base data directory (except on Windows)
--- a/src/options.rs
+++ b/src/options.rs
@@ -416,7 +416,6 @@ pub struct UI {
 pub struct Info {
    #[arg(long, value_parser=[
         "install-dir",
-        "binary-dir",
         "config-dir",
         "cache-dir",
         "data-dir",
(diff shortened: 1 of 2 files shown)


src/options.rs
```diff
@@ -415,13 +415,17 @@ pub struct UI {
 #[derive(clap::Args, Debug, Clone)]
 pub struct Info {
    #[arg(long, value_parser=[
+        "install-dir",
+        "binary-dir",
```
scotttrinh, line 419: I went with `binary-dir` over the discussed `bin-dir`. The longer form just seemed to fit with the other options here better than the shorter form.

scotttrinh: Furthermore, what we're actually interested in for the npx wrapper is the _install_ directory (which is confusingly called `binary_path`)

elprans: Hm. Wait. `install-dir` is what we actually want here. Why do we need the directory of the current binary?

scotttrinh: Mostly just because I started with `binary-dir` before I realized what we want is `install-dir`, so I just kept it. Also, this makes it so that everything you get in the `info` command* is available one-by-one using `--get`.

*Well, we actually show the path to the binary rather than the directory it's in in the info table, but this still seems right.

---

I can just remove the `binary-dir` option here if you'd prefer to only expose what we need.

scotttrinh: This is what the info table now shows for me running this with cargo on macOS, which shows the difference between `binary` and `install` more obviously:

```
EdgeDB uses the following local paths:
┌───────────────┬────────────────────────────────────────────────────────────┐
│ Cache         │ /Users/scotttrinh/Library/Caches/edgedb/                   │
│ Config        │ /Users/scotttrinh/Library/Application Support/edgedb/      │
│ Custom Binary │ /Users/scotttrinh/edgedb/edgedb-cli/target/debug/edgedb    │
│ Install       │ /Users/scotttrinh/Library/Application Support/edgedb/bin/  │
│ Data          │ /Users/scotttrinh/Library/Application Support/edgedb/data/ │
│ Service       │ /Users/scotttrinh/Library/LaunchAgents/                    │
└───────────────┴────────────────────────────────────────────────────────────┘
```

elprans: I think `binary-dir` is redundant.  You can always use `which edgedb` if you really need to find where the binary is located.

elprans (approved):

PR #1273, opened 2024-04-01
Ask branch for edgedb instance link if >=5.0

### LINKED ISSUE #1269: `edgedb instance link` asks for the branch name even when the target instance is running a version without branches
The `edgedb instance link` command now asks for a branch and defaults to `main`, but it does so indiscriminately, even when the instance is pre-v5, does not have branches, and when the default value of `main` would not make sense.

Example output:
```
$ edgedb instance link
[2024-03-29T18:15:22Z WARN  edgedb::version_check] Newer version of edgedb tool exists 4.3.0-dev.1120+d5587b7 (current 4.3.0-dev.1117+2dfb53b). To upgrade run `edgedb cli upgrade`
Specify server host [default: localhost]:
> localhost
Specify server port [default: 10915]:
> 10887
Specify database user [default: edgedb]:
> edgedb
Specify branch [default: main]:
> main
…
```
The instance on localhost:10887 is a 4.5 instance.

Is it possible to use the host and port to discover the version somehow? This would tee up the ideal DX where we could ask the user for a database name with `main` as the default for pre-v5 while asking for a branch and defaulting to `main` for 5+.

If not, maybe it would make sense to change the prompt to something like this:
> Specify branch (or database name for pre-v5 instances):
If we need to go this route, we may also want to log out a note or a warning of some kind to say that the default of `main` likely will not work on a pre-v5 database and that they should manually set their database name. Maybe even just do away with a default value altogether to avoid confusion.

- EdgeDB Version: N/A
- EdgeDB CLI Version: 4.3.0-dev.1120+d5587b7
- OS Version: macOS 14.0

Steps to Reproduce:

1. Run `edgedb instance link` and observe output

quinchs:
Fixes #1269

**Remarks**
I made the `instance_main` entry as `tokio::main` to fix some async issues later in the call chain, seems to have no outstanding negative effects.


src/portable/link.rs
```diff
@@ -162,26 +165,38 @@ fn gen_default_instance_name(input: impl fmt::Display) -> String {
...
+        let branch = question::String::new("Specify branch")
+            .default("main")
+            .ask()?;
+
+        creds.database = Some(branch.clone());
+        return Ok(config.with_database(&branch)?)
+    }
+
+    Ok(config)
 }
```
aljazerzen, line 184: Tests are failing because you are not respecting `--non-interactive` option.

aljazerzen: (I think)

src/portable/link.rs
```diff
@@ -198,32 +213,43 @@ pub fn link(cmd: &Link, opts: &Options) -> anyhow::Result<()> {
...
-            let password;
-
-            if opts.conn_options.password_from_stdin {
-                password = tty_password::read_stdin()?
-            } else if !cmd.non_interactive {
-                password = tty_password::read(format!(
+    let mut conn = connect(&config).await;
+
+    if let Err(e) = conn {
+        if let Some(err) = e.downcast_ref::<edgedb_tokio::Error>() {
```
aljazerzen, line 219: I would suggest not to bring Rust async into this fight :D It I very convenient in some use cases, but I think those use cases don't apply to simple CLI app that is running synchronously.

So my suggestion is to add `#[tokio::main]` to `connect()` & `ask_branch()` and to keep `link()` sync.

src/portable/link.rs
```diff
@@ -198,32 +213,43 @@ pub fn link(cmd: &Link, opts: &Options) -> anyhow::Result<()> {
...
+            else {
+                return Err(e.into());
             }
-            connect(&config)?;
         } else {
             return Err(e.into());
         }
+    } else if let Ok(mut conn) = conn {
+        config = ask_branch(&mut conn, &mut creds, config).await?
     }
```
aljazerzen, line 251: Here, you have:

```rust
if let Err(...) = x {
   ...
} else if let Ok(...) = x {
   ...
}
```

The idiomatic way to express this pattern would use `match`:

```rust
match x {
    Err(...) => ...
    Ok(...) => ...
}
```

(I think rust-analizer even has a "quick action" for this)

aljazerzen (approved):
Ok, I see what the fundamental problem is:

We ask for connection parameters before we connect and at that point, the CLI has no idea about the version of the server, so I cannot affect the prompt and the default.

The PR makes it so we connect to default database first (without specifying any branch), query the version and only then `ask_branch`.

This works fine.

---

An alternative approach would be to:
- use version-ambiguous wording in the prompt ("Specify database branch:") and
- not have any default set. This would allow the server to provide its default for the branch. We can then read it with another query, so that it can be written into the config.

quinchs:
Blocked on the changes to the `edgedb-rust` binding regarding branch connection

src/portable/link.rs
aljazerzen: In #1286, I've replaced all `#[tokio::main]` with `#[tokio::main(flavor = "current_thread")]`, which should be done here as well. It might even solve the issue with "shutdown context".

aljazerzen:
This is now failing only because we are not respecting `--non-interactive`.

Cargo.toml
```diff
@@ -34,10 +34,10 @@ heck = "0.4.0"
...
 [dependencies]
 edgeql-parser = {git = "https://github.com/edgedb/edgedb"}
-edgedb-protocol = {git = "https://github.com/edgedb/edgedb-rust/", features=["all-types"]}
-edgedb-derive = {git = "https://github.com/edgedb/edgedb-rust/"}
-edgedb-errors = {git = "https://github.com/edgedb/edgedb-rust/"}
-edgedb-tokio = {git = "https://github.com/edgedb/edgedb-rust/", features=["admin_socket", "unstable"]}
+edgedb-protocol = {git = "https://github.com/edgedb/edgedb-rust/", features=["all-types"], rev = "cf5043a"}
+edgedb-derive = {git = "https://github.com/edgedb/edgedb-rust/", rev = "cf5043a"}
+edgedb-errors = {git = "https://github.com/edgedb/edgedb-rust/", rev = "cf5043a"}
+edgedb-tokio = {git = "https://github.com/edgedb/edgedb-rust/", features=["admin_socket", "unstable"], rev = "cf5043a"}
```
quinchs, line 40: we should update `edgedb-rust`

aljazerzen: This dependency here uses the last commit from the repo or whatever ref is saved in `Cargo.lock`.

So I think you wanted to do `cargo update edgedb-tokio` here, and not pin these deps on one specific commit.

src/portable/link.rs
```diff
@@ -163,14 +168,41 @@ fn gen_default_instance_name(input: impl fmt::Display) -> String {
 }
 
 #[tokio::main]
```
aljazerzen, line 170: Can you also convert these to `#[tokio::main(flavor = "current_thread")]`?

aljazerzen (approved):

PR #1276, opened 2024-04-02
Rewrite some help text

raddevon:


quinchs (approved):
Should we remove all `.` at the end of all help texts/options? seems that we don't have them in the first place elsewhere:
https://github.com/edgedb/edgedb-cli/blob/master/src/portable/options.rs#L36-L130

but in some places we do:
https://github.com/edgedb/edgedb-cli/blob/master/src/cloud/options.rs#L13-L122

raddevon:
> Should we remove all . at the end of all help texts/options?
Yeah, I think so. I'll go through and make them consistent before I merge this. If we have any case where a single help message consists of more than one sentence, we'll use periods. Otherwise, we can just omit them, I think.

raddevon:
Now, as I'm applying that rule, I'm starting to second guess. Maybe since there are cases where we need periods to delineate multiple sentences, we should just have them everywhere instead of trying to decide when to have them and when not to?

PR #1283, opened 2024-04-03
Branches: Add current command

quinchs:


src/branch/option.rs
```diff
@@ -125,3 +126,9 @@ pub struct Merge {
     #[arg(long)]
     pub no_apply: bool,
 }
+
+#[derive(clap::Args, Clone, Debug)]
+pub struct Current {
+    #[arg(long)]
+    pub shell: bool
```
aljazerzen, line 133:
```suggestion
    pub plain: bool
```

src/branch/current.rs
```diff
@@ -0,0 +1,24 @@
...
+use crate::branch::option::Current;
+use crate::connect::Connection;
+
+pub async fn main(
+    options: &Current,
+    context: &Context,
+) -> anyhow::Result<()> {
+    if options.shell {
+        if let Some(branch) = &context.branch {
+            eprintln!("{}", branch);
```
aljazerzen, line 12:
```suggestion
            println!("{}", branch);
```

This one is meant to be used as "the output" of the command and not a message.

aljazerzen (approved):

src/branch/option.rs
```diff
@@ -125,3 +126,12 @@ pub struct Merge {
     #[arg(long)]
     pub no_apply: bool,
 }
+
+/// Prints the current branch.
+#[derive(clap::Args, Clone, Debug)]
+pub struct Current {
+    /// Print as plain output to stdout; printing nothing instead of erroring if the current branch
+    /// can't be resolved
```
raddevon, line 134:
```suggestion
    /// Print as plain text output to stdout. Prints nothing instead of erroring if the current branch
    /// can't be resolved.
```

Issue #1263, opened 2024-03-27
Add REPL backslash commands for branches


raddevon:
We currently have REPL backslash commands for databases. We should add new commands for branches.

Here are some commands that need branch analogs and my proposals for their counterparts:
- `\list databases` -> `\list branches`
- `\database create [name]` -> `\branch create [name]`
- `\c, \connect [dbname]` -> `\switch [branch-name]` and maybe `\s`?

raddevon:
One more interesting wrinkle here: we have `\l` as an alias for `\list databases`. This gets confusing when some instances have databases and others have branches. Is it possible for the `\l` shorthand to be adaptive to alias either `\list databases` or `\list branches`, whichever is applicable to the current database? This way, if anyone has built up muscle memory around `\l`, they can continue to use it across any of their instances, regardless of whether those instances have databases or branches.

quinchs:
I like the approach of `\l` being version-aware, ex `\l databases` prints warning in >=5.x server and `\l branches` prints warning in <5.0

Issue #1275, opened 2024-04-02
Add `edgedb list branches` as the branch analog to `edgedb list databases`


raddevon:
`list databases` does seem to work on v5 instances, but people may not know to use it since we are re-training them that databases are out and branches are in. Looks to me like `list branches` could just be an alias for `list databases` except that I guess it should probably fail on pre-v5.

1st1:
I think we should instead deprecate `list databases` and print a recommendation to use `branch list`

raddevon:
Would we then make that command work for users on versions prior to 5 to list their databases, or would that recommendation not apply to the users who would have wanted to use `edgedb list databases` to list databases on a pre-v5 database? Either of those seems like awkward DX to me, but maybe there's a third option I'm not considering…

PR #1284, opened 2024-04-04
Branches: REPL backslash commands for branches

quinchs:
Fixes https://github.com/edgedb/edgedb-cli/issues/1263

### DIFF
--- a/src/commands/execute.rs
+++ b/src/commands/execute.rs
@@ -29,6 +29,9 @@ pub async fn common(cli: &mut Connection, cmd: &Common, options: &Options)
             }
             ListCmd::Databases => {
                 commands::list_databases(cli, &options).await?;
+            },
+            ListCmd::Branches => {
+                commands::list_branches(cli, &options).await?;
             }
             ListCmd::Scalars(c) => {
                 commands::list_scalar_types(cli, &options,
--- a/src/branch/current.rs
+++ b/src/branch/current.rs
@@ -1,7 +1,7 @@
 use crossterm::style::Stylize;
 use crate::branch::context::Context;
 use crate::branch::option::Current;
-use crate::connect::Connection;
+
 
 pub async fn main(
     options: &Current,
--- a/src/commands/backslash.rs
+++ b/src/commands/backslash.rs
@@ -41,7 +41,7 @@ Introspection
 
   \d [-v] NAME              Describe schema object
   \ds                       Describe whole schema   (alias: \describe schema)
-  \l                        List databases          (alias: \list databases)
+  \l                        List databases/branches (alias: \list branches)
   \ls [-sc]  [PATTERN]      List scalar types       (alias: \list scalars)
   \lt [-sc]  [PATTERN]      List object types       (alias: \list types)
   \lr [-c]   [PATTERN]      List roles              (alias: \list roles)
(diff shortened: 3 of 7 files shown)


quinchs:
Also fixes https://github.com/edgedb/edgedb-cli/issues/1275

src/migrations/create.rs
```diff
@@ -106,12 +107,11 @@ pub enum MigrationKey {
...
 }
 
-pub trait MigrationToText {
-    type StatementsIter<'a>: Iterator<Item = &'a String> where Self: 'a;
+pub trait MigrationToText<'a, T: Iterator<Item = &'a String> = std::iter::Once<&'a String>> {
     fn key(&self) -> &MigrationKey;
     fn parent(&self) -> anyhow::Result<&str>;
     fn id(&self) -> anyhow::Result<&str>;
-    fn statements<'a>(&'a self) -> Self::StatementsIter<'a>;
+    fn statements(&'a self) -> T;
```
quinchs, line 114: For context of this change, see https://github.com/rust-lang/rust/issues/102211#issuecomment-1513931928 & https://edgedb.slack.com/archives/CNECHA9EW/p1712953250921689

quinchs:
Adding other branch commands is simple by adding them to the `BranchingCmd` enum

src/commands/execute.rs
```diff
@@ -1,15 +1,15 @@
...
-use crate::commands::{self, Options};
+use crate::commands::{self, branching, Options};
 use crate::migrations::options::{MigrationCmd};
 use crate::migrations;
 use crate::print;
 
 
-pub async fn common(cli: &mut Connection, cmd: &Common, options: &Options)
+pub async fn common(cli: &mut Connection, cmd: &Common, options: &Options, cli_opts: &options::Options)
     -> Result<(), anyhow::Error>
```
aljazerzen, line 13: Needing to add `cli_opts` here suggests that we should not be passing-in these options to branching commands, since no other command is doing that.

From a quick glance over the code, it seems that we only use `options.create_connector()`. So I think we should replace `options::Options` with `commands::Options` in branching commands.

I'll try to do that in a separate PR.

aljazerzen (approved):

src/commands/execute.rs
aljazerzen: It is possible! https://github.com/edgedb/edgedb-cli/pull/1288

src/migrations/create.rs
quinchs: _a lot of debugging_ ended up the codepath in branch commands that used that trait, and some cheeki googling

src/commands/execute.rs
quinchs: Should I revert the changes to the statements trait?

aljazerzen: Does this change get us around the rust bug? I though it was unrelated?

I'm -0.5 (on the scale from -1 to 1) for the change of `MigrationToText`'s associated type to a generic param. It makes more sense to me for it to be an associated type. But if rust cannot handle that, generic param is fine.

quinchs: I'll verify whether your changes fixes it, my guess is it _should_ since the bug derives from passing `options` in interactive

quinchs: Seems to still be an issue, even after your changes:
```
error: implementation of `std::marker::Send` is not general enough
   --> src\interactive.rs:144:18
    |
144 |     let handle = runtime.spawn(_main(options, state, cfg));
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ implementation of `std::marker::Send` is not general enough
    |
    = note: `std::marker::Send` would have to be implemented for the type `&()`
    = note: ...but `std::marker::Send` is actually implemented for the type `&'0 ()`, for some specific lifetime `'0`
```

aljazerzen: Don't bother with this then, generic param is fine.

PR #1294, opened 2024-04-18
Organize and document tests

aljazerzen:
Document different types of tests we have, what they do and their requirements.

Move "common util" files into "tests/common/" dir.

Rename a few things.


quinchs (approved):
🙏

Issue #1192, opened 2023-12-19
Command to generate hashed password


MrFoxPro:
There is only one way to set hashed password in edgedb: via CLI. When running edgedb it's often useful to configure  user in startup script, like this:
```nix
systemd.services.edgedb = {
  # ...
  path = ["/nix/var/nix/profiles/per-user/root/edgedb"];
  script = concatStringsSep " \\\n" [
    "edgedb-server"
    "--runstate-dir=/run/edgedb"
    "--backend-dsn=postgresql://?host=/run/postgresql"
  ];
  postStart = ''
    while [ ! -S ${edb.socket} ]; do sleep 2; done
    chmod g+rw ${edb.socket}
    ${edb.cmd} query 'ALTER ROLE edgedb SET password_hash := "${edb.password_hash}";' ||:
  '';
  # ...
};
```
And there is no way to do this. To achieve that, I'm running Rust tests from this repo with modified code to get generated hashed password:
```rs
#[test]
pub fn test_bootstrap_script(){
    let pass = "mypassword";
    let output = bootstrap_script("edgedb", "edgedb", pass);
    println!("initial script is: {}", output);
}
```
I think it's worth to add command to generate hashed password from CLI.

raddevon:
Maybe do this in your [`--bootstrap-command`](https://www.edgedb.com/docs/reference/environment#edgedb-server-bootstrap-command)?

MrFoxPro:
> Maybe do this in your [`--bootstrap-command`](https://www.edgedb.com/docs/reference/environment#edgedb-server-bootstrap-command)?

Then I forced to store actual password in plaintext if not using some special tool to read it from file.

raddevon:
I thought your goal was to store a hash, not a plain text password…

PR #1295, opened 2024-04-19
feat: add command to hash password

MrFoxPro:
https://github.com/edgedb/edgedb-cli/issues/1192


aljazerzen (approved):
I've checked with the team, this is a valid use-case and a nice thing to have in the CLI.

Thank you!

PR #1297, opened 2024-04-19
Remove non-recursive access policies

quinchs:
### Summary
This PR removes the creation of `futures.esdl` with non-recursive access policies because of a bug with checking whether or not a version should have `futures.esdl`.

see https://edgedb.slack.com/archives/CNECHA9EW/p1713554610426989

### DIFF
--- a/src/portable/project.rs
+++ b/src/portable/project.rs
@@ -51,13 +51,6 @@ const DEFAULT_ESDL: &str = "\
     }\n\
 ";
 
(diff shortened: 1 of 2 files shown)


aljazerzen:
As I've replies in slack, I think we should fix `is_nonrecursive_access_policies_needed` instead of removing this completely.

By "fix" i mean to make is so it returns `false` when release name does not contain a version number.

PR #1301, opened 2024-04-23
Bump version to 5.0.0-dev

quinchs:


aljazerzen (approved):

Cargo.toml
```diff
@@ -8,7 +8,7 @@ members = [
 [package]
 name = "edgedb-cli"
 license = "MIT/Apache-2.0"
-version = "4.3.0-dev"
+version = "5.0.0-dev"
```
elprans, line 11: Shouldn't this be 5.1.0-dev?

elprans (changes requested):
`5.1.0-dev` is the correct bump for `master` since `5.0.0` has been released already.

elprans (approved):

PR #1309, opened 2024-04-29
Allow cloud instances for branches

quinchs:
Cloud instances don't have a `credentials.json` file to store the current branch information in, instead we allow cloud instances to be used *if* they're mapped to a project, we then update the `database` file in that projects stash directory to keep track of the current branch.

### DIFF
--- a/src/branch/context.rs
+++ b/src/branch/context.rs
@@ -1,6 +1,5 @@
 use edgedb_tokio::get_project_dir;
 
-
 use crate::commands::Options;
 use crate::connect::Connection;
 use crate::credentials;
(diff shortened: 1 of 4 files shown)


src/branch/context.rs
```diff
@@ -19,44 +22,55 @@ pub struct Context {
...
+        // read from credentials
+        if project_dir.is_some()
+            && instance_name.as_ref().map_or(false, |v| match v {
+                InstanceName::Local(_) => true,
+                _ => false,
+            })
+        {
+            let instance_name = match instance_name.as_ref().unwrap() {
+                InstanceName::Local(instance) => instance,
+                unsupported => anyhow::bail!(format!("Cannot use instance type {}", unsupported)),
```
elprans, line 56: A better error message is needed:

```suggestion
                unsupported => anyhow::bail!(format!("cannot switch branches on Cloud instance {org}/{slug}: instance is not linked to a project", unsupported)),
```

elprans: Wait, why does this apply to _all_ branch commands?  Shouldn't only the `switch` be affected?

quinchs:
> Wait, why does this apply to all branch commands? Shouldn't only the switch be affected?

I guess it can work for non switch/rename commands, assuming the database default branch

quinchs:
Alright, all commands work for cloud instances except switch/rename if theres no project linked to the instance

quinchs:
if you rename the current branch, it will preform a switch to the new name

quinchs:
> So only gate the switching part

thats what this PR does now

elprans:
Please fix `rustfmt`

elprans (approved):

PR #1324, opened 2024-05-09
Fix force_database_error and get_current_branch

### LINKED ISSUE #1321: Getting into CLI after setting `force_database_error`?
- EdgeDB Version: 4.7
- EdgeDB CLI Version: 5.1.0
- OS Version: macos

Steps to Reproduce:

1. Set `force_database_error`, eg `configure current database set force_database_error := '{"type": "QueryError", "message": "ongoing maintenance"}';`
2. If you quit your CLI session, you are seemingly permanently locked out and can't run `configure current database reset force_database_error;`:

```sh
$ edgedb
EdgeDB 4.7+45e4acc (repl 5.1.0+8ff9f85)
edgedb error: ConfigurationError: invalid 'force_database_error' value'
```

Luckily was able to get in via the UI.

aljazerzen:
When opening REPL, we query `sys::get_current_database`.
If `force_database_error` is set, this query fail, preventing even
entering the REPL, which is bad for obvious reasons.

I've made sure to override `force_database_error` is these queries
and even made it so if we already have branch set in connection
options, we don't even make the query, but just use that query param
as the current branch.

Closes #1321

### DIFF
--- a/tests/func/non_interactive.rs
+++ b/tests/func/non_interactive.rs
@@ -251,3 +251,31 @@ fn hash_password() {
         .success()
         .stdout(predicates::str::starts_with("SCRAM-SHA-256$"));
 }
+
+#[test]
+fn force_database_error() {
+    SERVER
+        .admin_cmd()
+        .arg("query")
+        .arg(
+            r#"configure current database
+                set force_database_error := 
+                  '{"type": "QueryError", "message": "ongoing maintenance"}';
+            "#,
+        )
+        .assert()
+        .context("set force_database_error", "should succeed")
+        .success();
+
+    SERVER
+        .admin_cmd()
+        .arg("query")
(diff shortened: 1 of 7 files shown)


quinchs (approved):

tests/func/non_interactive.rs
```diff
@@ -251,3 +251,31 @@ fn hash_password() {
         .success()
         .stdout(predicates::str::starts_with("SCRAM-SHA-256$"));
 }
+
+#[test]
+fn force_database_error() {
```
msullivan, line 256: Could you add a test to `interactive.rs` that fires up a new *interactive* session once force_database_error has been set?

The `edgedb query` case actually already worked, it was connecting interactively that was broken

Issue #1195, opened 2024-01-04
Session config lost if error is returned on first query after two successive reconnects


msullivan:
Here, I set a session config variable. Then, I twice run a query that errors (here, `select asdf`), while restarting edgedb between commands so that the CLI needs to reconnect. If I do this *twice* in a row, the session config gets lost!
If I only do it once, then it manages to keep the config.

This is probably somewhat marginal for end users, but I hit it in my development workflow because I'm constantly running queries that don't work and then restarting the server to run a hopefully fixed version!

```
_localdev:edgedb> configure session set apply_access_policies := false;
OK: CONFIGURE SESSION

_localdev:edgedb> select cfg::Config.apply_access_policies;
[false]

_localdev:edgedb> select asdf;
Reconnecting...
error: InvalidReferenceError: object type or alias 'default::asdf' does not exist

_localdev:edgedb> select asdf;
Reconnecting...
error: InvalidReferenceError: object type or alias 'default::asdf' does not exist

_localdev:edgedb> select cfg::Config.apply_access_policies;
[true]
```

msullivan:
I'm not sure if *this* case was a regression or if this was always fiddly (see https://github.com/edgedb/edgedb-cli/pull/1183)

msullivan:
Aha! And it *also* loses the automatically set session_idle_transaction_timeout! I've been haunted by session idle transaction timeouts for years, and I had still been getting it even after we were pretty sure we had squashed it in the server!
```
_localdev:edgedb> select cfg::Config.session_idle_transaction_timeout;
{<duration>'0:00:10'}
_localdev:edgedb> start transaction;
OK: START TRANSACTION
_localdev:edgedb[tx]> 
Reconnecting...
_localdev:edgedb> 
```

PR #1340, opened 2024-06-11
Fix lost connections state during multiple reconnects

aljazerzen:
Closes https://github.com/edgedb/edgedb-cli/issues/1195


aljazerzen:
This is very hard to test as it would require us to restart the server or drop connection some other way.

msullivan (approved):
:fireworks:

PR #1347, opened 2024-07-02
Make migration apply more verbose

### LINKED ISSUE #1346: Have `migrate` indicate when we are compiling the migration and when we are applying it
Pure CLI component of https://github.com/edgedb/edgedb/issues/7490.

I'm pretty sure we send Parse and Execute separately and can report when Parse has finished.

aljazerzen:
Closes #1346

image

### DIFF
--- a/src/migrations/dev_mode.rs
+++ b/src/migrations/dev_mode.rs
@@ -220,7 +220,7 @@ pub async fn rebase_to_schema(
     execute(cli, "START MIGRATION REWRITE").await?;
 
     let res = async {
-        apply_migrations_inner(cli, migrations, true).await?;
+        apply_migrations_inner(cli, migrations, false).await?;
         migrate_to_schema(cli, ctx).await?;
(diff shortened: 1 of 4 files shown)


src/migrations/dev_mode.rs
```diff
@@ -220,7 +220,7 @@ pub async fn rebase_to_schema(
     execute(cli, "START MIGRATION REWRITE").await?;
 
     let res = async {
-        apply_migrations_inner(cli, migrations, true).await?;
+        apply_migrations_inner(cli, migrations, false).await?;
```
msullivan, line 223: Put comments saying `/*verbose*/` or some such?

I really miss named arguments.

msullivan (approved):

src/migrations/dev_mode.rs
aljazerzen: image

Issue #1211, opened 2024-02-16
Support `CONFIGURE CURRENT DATABASE`


msullivan:
Currently config always does instance. We should be able to do database too.

Issue #1176, opened 2023-11-08
`edgedb configure` should maybe get the list of config values via reflection


msullivan:
Otherwise we'll probably end up always missing some.
If we do it via reflection, it's easy to include extension configs too.

This should get all the config properties that can be set with SET
```
select schema::ObjectType {
    name,
    properties: { name, typ := .target.name },
}
filter .name = 'cfg::AbstractConfig'
or 'cfg::ExtensionConfig' in .ancestors.name;
```

PR #1350, opened 2024-07-12
Add config auto_rebuild_query_cache_timeout

fantix:
This should fix the failing CI


msullivan (approved):
Sure.

We really should do https://github.com/edgedb/edgedb-cli/issues/1176 and https://github.com/edgedb/edgedb-cli/issues/1211 at some point

PR #1352, opened 2024-07-15
Revert dump overwrite default to true

fantix:
A proper fix would be to treat `/dev/` differently than the current branch of disabling overwriting, but let's allow overwriting for all first.


mmastrac (approved):
LGTM

PR #1361, opened 2024-08-27
Allow cloud instance names to contain '-' and '_' in the org part to support vercel instances

jaclarke:
The org slug for vercel instances are prefixed with `-v` to prevent collisions with github org slugs, but in the cli we validate that org slugs are 'github like', so currently the cli can't connect to vercel instances. (Some older vercel instances orgs may be prefixed with `_v` and also contain `_`, so the org slug validation is also updated to support those.)

Depends on https://github.com/edgedb/edgedb-rust/pull/346

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,10 +34,10 @@ heck = "0.4.0"
 
 [dependencies]
 edgeql-parser = {git = "https://github.com/edgedb/edgedb"}
-edgedb-protocol = {git = "https://github.com/edgedb/edgedb-rust/", features=["all-types"]}
-edgedb-derive = {git = "https://github.com/edgedb/edgedb-rust/"}
-edgedb-errors = {git = "https://github.com/edgedb/edgedb-rust/"}
-edgedb-tokio = {git = "https://github.com/edgedb/edgedb-rust/", features=["admin_socket", "unstable"]}
+edgedb-protocol = {git = "https://github.com/edgedb/edgedb-rust/", rev="a2dc68a7a6e6400cd03a0a162b72cac686653f2c", features=["all-types"]}
+edgedb-derive = {git = "https://github.com/edgedb/edgedb-rust/", rev="a2dc68a7a6e6400cd03a0a162b72cac686653f2c"}
+edgedb-errors = {git = "https://github.com/edgedb/edgedb-rust/", rev="a2dc68a7a6e6400cd03a0a162b72cac686653f2c"}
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -34,10 +34,10 @@ heck = "0.4.0"
...
 [dependencies]
 edgeql-parser = {git = "https://github.com/edgedb/edgedb"}
-edgedb-protocol = {git = "https://github.com/edgedb/edgedb-rust/", features=["all-types"]}
-edgedb-derive = {git = "https://github.com/edgedb/edgedb-rust/"}
-edgedb-errors = {git = "https://github.com/edgedb/edgedb-rust/"}
-edgedb-tokio = {git = "https://github.com/edgedb/edgedb-rust/", features=["admin_socket", "unstable"]}
+edgedb-protocol = {git = "https://github.com/edgedb/edgedb-rust/", rev="dd0f4c945e22da4ef5ccc21016fb8ada2b40e398", features=["all-types"]}
+edgedb-derive = {git = "https://github.com/edgedb/edgedb-rust/", rev="dd0f4c945e22da4ef5ccc21016fb8ada2b40e398"}
+edgedb-errors = {git = "https://github.com/edgedb/edgedb-rust/", rev="dd0f4c945e22da4ef5ccc21016fb8ada2b40e398"}
+edgedb-tokio = {git = "https://github.com/edgedb/edgedb-rust/", rev="dd0f4c945e22da4ef5ccc21016fb8ada2b40e398", features=["admin_socket", "unstable"]}
```
jaclarke, line 40: I pinned the commit here, as it wasn't building for me with the latest master, but I'm not that familiar with rust/cargo so I may have been doing something wrong.

fantix: Yeah that looks like a recent upstream breaking change or something. Revert both `Cargo.toml` and `Cargo.lock` and it builds for me.

fantix (approved):
LGTM! We need to apply the same change to edgedb-rust under `edgedb-tokio/src/builder.rs` though.

Cargo.toml
aljazerzen: I've updated edgedb-cli dependencies, which should resolve your problem.

I suggest you first merge changes into edgedb-rust and then remove these `rev`s here and run `cargo update`. This will update the lock file to pick up latest changes from edgedb-rust.

jaclarke: Thanks, I can confirm that it builds for me with that PR.

PR #1362, opened 2024-08-29
Update bigdecimal and rustls dependencies

aljazerzen:
Update `bigdecimal` and `rustls` dependencies.

rustls 0.23 has [switched from crypto provider `ring` to `aws-lc-rs`](https://github.com/rustls/rustls/releases/tag/v%2F0.23.0). This was done mainly for performance reasons (from what I found). The downside is that `aws-ls-rs` does need additional build tools (make and a C compiler). This breaks our builds, because we use a version of gcc that has bug:

```
Your compiler (cc) is not supported due to a memcmp related bug reported in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189.We strongly recommend against using this compiler.EXECUTED: true ERROR:  OUTPUT:  
  stack backtrace:
     0: rust_begin_unwind
               at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/std/src/panicking.rs:645:5
     1: core::panicking::panic_fmt
               at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/panicking.rs:72:14
     2: build_script_main::cc_builder::CcBuilder::memcmp_check
     3: build_script_main::cc_builder::CcBuilder::compiler_checks
     4: build_script_main::cc_builder::CcBuilder::build_library
     5: <build_script_main::cc_builder::CcBuilder as build_script_main::Builder>::build
     6: build_script_main::main
     7: core::ops::function::FnOnce::call_once
```

My solution was to make sure that we explicitly state that we want to use `ring` as the cryto provider instead of `aws-lc-rs`.

The real solution would be to embrace `aws-lc-rs` and ensure correct build tools.

---

There is a dev-dep on warp that uses tokio-rustls, which uses an old version of rustls, which still defaults to using ring. When that's updated, we might encounter this problem again.


fantix (approved):
Thanks!

PR #1365, opened 2024-09-05
Print SNI from credentials file.

diksipav:
Changes for:
- adding SNI to the credentials file when linking remote instance (I just updated `Credentials` interface, `--tls-server-name` flag is already added as a global flag )
- reading SNI from credentials file together with other fields. 

Here is the [Rust PR](https://github.com/edgedb/edgedb-rust/pull/348).


aljazerzen (approved):

fantix (approved):

fantix (commented):
Please also inlcude the updates from `edgedb-rust` in `Cargo.lock` so that the CI passes.

Issue #1054, opened 2023-05-25
OS-level notifications for all local instances when upgrading a single instance to a new version


raddevon:
When upgrading a local instance to a new server version, I get one OS-level notification that a background item has been added for each instance running on my system. I would expect to get only one for the newly upgraded instance.

![image](https://github.com/edgedb/edgedb-cli/assets/2027711/6009b0d4-93de-4c5f-b220-3cfe15a808c1)

- EdgeDB Version: 3.0-rc.1+23c3462
- EdgeDB CLI Version: EdgeDB CLI 3.1.0-dev.994+eb21ca5
- OS Version: macOS 13.3.1

Steps to Reproduce:

1. With other instances on the system, upgrade an existing instance to a new version
2. Observe notifications

raddevon:
This is really strange. I just launched UTM (a macOS GUI for QEMU), and I got the same thing — notifications for all of my EdgeDB instances. 🤔 

Just closed it and re-launched it, and I didn't get the notifications. Seems it was only on initial launch of UTM.

tailhook:
I don't see any code that touches other services on upgrade. So probably this is some MacOS bug that shows notifications of all services when one is changed, as confirmed by UTM launch.

Can you also try upgrading an instance after UTM is installed and look whether UTM is also shown in notifications (i.e. confirm that this is not EdgeDB-specific issue).

raddevon:
I have only ever seen these "off-topic" notifications referencing the `edgedb` service although I have quite a few other background services running. Here's a partial list of some non-`edgedb` services I run:
![CleanShot 2023-05-30 at 07 19 27@2x](https://github.com/edgedb/edgedb-cli/assets/2027711/124a0a8d-bc53-48e8-8000-ce3d9116a733)

I run close to a dozen others, and I've only ever seen notifications referencing them when installing the associated apps.

tailhook:
Looks like #1058 should fix it. Would you mind testing it git version? We do rewrite `.plist` files on stop/start currently. So to test it you can stop and then start your instances (each of them separately, or just some of them to see that there are less notifications). And then do whatever was triggering this notification (i.e. the upgrade of some instance).

raddevon:
I don't mind testing it. How would I do that? I'm not sure how to run the CLI out of a particular branch.

tailhook:
If you have Rust installed run in the edgedb-cli repository:
```
git fetch
git checkout suppress_bg
[...]

PR #1367, opened 2024-09-15
Try to fix macOS bursting bubbles issue

fantix:
When creating a new instance on macOS, sometimes lots of bubble notifications show up at once. I'm not able to reliably reproduce the issue, so this is a guessed fix trying to avoid `libc::mkdir("~/Library/LaunchAgents", ...)` if unnecessary.

See also #1054

### DIFF
--- a/src/portable/macos.rs
+++ b/src/portable/macos.rs
@@ -137,16 +137,25 @@ fn plist_data(name: &str, info: &InstanceInfo) -> anyhow::Result<String> {
     ))
 }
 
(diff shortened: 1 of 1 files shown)


mmastrac:
I have a feeling this won't help. The extra mkdir is likely just erroring out without performing any modifications -- the underlying cause appears to be a background launch agent database that gets corrupted:

https://discussions.apple.com/thread/254341579?sortBy=rank

`sfltool resetbtm` is apparently the way to fix this. 

We could still land this as it's probably harmless either way.

PR #1372, opened 2024-09-30
Implement `instance restore` and `instance list-backups` commands

elprans:
Add support for listing and restoring instance backups

### DIFF
--- a/src/portable/options.rs
+++ b/src/portable/options.rs
@@ -529,7 +529,7 @@ pub struct ListBackups {
     #[command(flatten)]
     pub cloud_opts: CloudOptions,
 
-    /// Instance to resize.
+    /// Instance to list backups for.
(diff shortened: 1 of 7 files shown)


src/portable/options.rs
```diff
@@ -519,6 +523,55 @@ pub struct Resize {
...
+    #[arg(long)]
+    pub latest: bool,
+}
+
+#[derive(clap::Args, IntoArgs, Debug, Clone)]
+pub struct Restore {
+    #[command(flatten)]
+    pub cloud_opts: CloudOptions,
+
+    /// Instance to resize.
```
fantix, line 557: > resize

copy paste error?

fantix (approved):

src/portable/options.rs
elprans: yeah

PR #1385, opened 2024-10-21
Extension installation commands

mmastrac:
This PR adds two supported commands:

```
Manage local extensions

Usage: edgedb extension <COMMAND>

Commands:
  list-available  List available extensions for a local instance
  install         Install an extension for a local instance
  help            Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
```

install:

```
Install an extension for a local instance

Usage: edgedb extension install [OPTIONS] --extension <EXTENSION>

Options:
  -I, --instance <INSTANCE>    Specify local instance name
  -E, --extension <EXTENSION>  Name of the extension to install
  -h, --help                   Print help
```

Example run:

```
> extension install -E postgis -I testnightly --slot 6-dev8882
Found extension package: postgis version 3.5.0-dev.22+6c1775b
00:00:00 [====================] 22.51 MiB/22.51 MiB 32.22 MiB/s | ETA: 0s                                                                                     Extension 'postgis' installed successfully.
```

list-available:

```
List available extensions for a local instance

Usage: edgedb extension list-available [OPTIONS]

Options:
  -I, --instance <INSTANCE>  Specify local instance name
  -h, --help                 Print help
```

Example run:

```
> cargo run -- extension list-available -I testnightly --slot 6-dev8882
┌─────────┬──────────────────────┐
│ Name    │ Version              │
│ postgis │ 3.5.0-dev.22+6c1775b │
└─────────┴──────────────────────┘
```

Two hidden commands are added as well (`uninstall` and `list`) but these require server support before they can be unhidden.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -355,6 +355,8 @@ pub enum Command {
     Instance(portable::options::ServerInstanceCommand),
     /// Manage local EdgeDB installations
     Server(portable::options::ServerCommand),
+    /// Manage local extensions
+    Extension(portable::options::ServerInstanceExtensionCommand),
     /// Generate shell completions
     #[command(name = "_gen_completions")]
     #[command(hide = true)]
--- a/src/portable/repository.rs
+++ b/src/portable/repository.rs
@@ -219,7 +219,11 @@ where
     }
 }
 
-fn filter_package(pkg_root: &Url, pkg: &PackageData, expected_package_type: PackageType) -> Option<PackageInfo> {
+fn filter_package(
+    pkg_root: &Url,
+    pkg: &PackageData,
+    expected_package_type: PackageType,
+) -> Option<PackageInfo> {
     let result = _filter_package(pkg_root, pkg, expected_package_type);
     if result.is_none() {
         log::info!("Skipping package {:?}", pkg);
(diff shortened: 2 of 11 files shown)


src/portable/ver.rs
```diff
@@ -212,6 +213,16 @@ impl Specific {
...
+
+    pub fn slot(&self) -> String {
+        match self.minor {
+            MinorVersion::Minor(_) => self.major.to_string(),
+            MinorVersion::Dev(v) => format!("{}-dev{}", self.major, v),
+            MinorVersion::Alpha(v) => format!("{}-alpha{}", self.major, v),
+            MinorVersion::Beta(v) => format!("{}-beta{}", self.major, v),
+            MinorVersion::Rc(v) => format!("{}-rc{}", self.major, v),
+        }
+    }
```
aljazerzen, line 225: I'm not sure if I understand what's a "slot".

src/options.rs
```diff
@@ -355,6 +355,8 @@ pub enum Command {
     Instance(portable::options::ServerInstanceCommand),
     /// Manage local EdgeDB installations
     Server(portable::options::ServerCommand),
+    /// Manage local extensions
+    Extension(portable::options::ServerInstanceExtensionCommand),
```
aljazerzen, line 359: Does commands under `exension` always work some specific instance? If so, do you think it would fit better under the `instance` top-level command?

For example: `edgedb instance extension list-installed` and `edgedb instance extension install`.

src/portable/ver.rs
```diff
@@ -50,7 +50,8 @@ pub enum FilterMinor {
 }
 
 static BUILD: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(r#"^\d+\.\d+(?:-(?:alpha|beta|rc|dev)\.\d+)?\+(?:[a-f0-9]{7}|local)$"#).unwrap()
+    Regex::new(r#"^\d+\.\d+(?:\.\d+)?(?:-(?:alpha|beta|rc|dev)\.\d+)?\+(?:[a-f0-9]{7}|local)$"#)
+        .unwrap()
```
aljazerzen, line 54: The change here was to support the third number for patch versions? `(?:\.\d+)?`
Because extensions do publish patch releases?

aljazerzen (commented):
I've ran out of time for today, will get back to it.

Most of the command have to do with me not understanding stuff, so this review is more of "learn the codebase" and less of "checking if code is good enough" :D

src/portable/repository.rs
```diff
@@ -330,7 +387,31 @@ fn get_platform_server_packages(
...
+        Err(e) => return Err(e),
+    };
+    let packages = data
+        .packages
+        .iter()
+        .filter(|pkg| {
+            pkg.tags.contains_key("extension")
+                && pkg.tags.get("server_slot").map(|s| s.as_str()) == Some(slot)
+        })
+        .filter_map(|p| filter_package(pkg_root, p, PackageType::Zip))
```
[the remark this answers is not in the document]
mmastrac: Per slack -- we only support zip installation for extensions.

src/portable/ver.rs
mmastrac: That's how extensions are linked to server builds -- see the tags in the JSON manifest. I have a feeling we'll want to use semver ranges instead.

https://packages.edgedb.com/archive/.jsonindexes/x86_64-unknown-linux-gnu.json

src/options.rs
mmastrac: Per slack -- this was @elprans 's recommendation

src/portable/ver.rs
mmastrac: Our extension(s) have a third version number -- so we need to parse this out. Long term it would be nice to use a more structured form for this rather than strings and regexps, but I didn't want to duplicate the whole PackageInfo struct tree.

src/portable/repository.rs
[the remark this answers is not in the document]
mmastrac: I checked the manifest and the savings are quite significant -- it might be worth it for us to save ~20MB on nearly every extension installation.

aljazerzen (approved):

PR #1390, opened 2024-11-01
Prepare to rename EdgeDB -> Gel

mmastrac:
Paves the way for a rename patch where we can rename EdgeDB to Gel in one swoop. Each use of EdgeDB is parameterized.

`EdgeDB cloud` strings have become `BRANDING_CLOUD`.

This patch doesn't perform the rename yet. We also don't perform any directory or configuration file migrations yet. We do, however, set the default config file to `gel.toml`.

Review notes:

 - `format`/`println`/`cformat`/`log` support `{}`
 - `print::` and `echo!` must use `concatcp!`

### DIFF
--- a/src/commands/ui.rs
+++ b/src/commands/ui.rs
@@ -3,6 +3,7 @@ use std::io::{stdout, Write};
 use anyhow::Context;
 use const_format::concatcp;
 
+use crate::branding::BRANDING;
 use crate::cloud;
 use crate::commands::ExitCode;
 use crate::options::{Options, UI};
--- a/src/commands/ui.rs
+++ b/src/commands/ui.rs
@@ -119,7 +119,7 @@ fn _get_local_ui_url(cmd: &UI, cfg: &edgedb_tokio::Config) -> anyhow::Result<Str
                         " server."
                     ));
                     print::echo!(
-                        "  If you have {BRANDING} 2.0 and above, try running the \
+                        "  Try running the \
                         server with `--admin-ui=enabled`."
                     );
                     return Err(ExitCode::new(2).into());
--- a/src/commands/ui.rs
+++ b/src/commands/ui.rs
@@ -1,6 +1,7 @@
 use std::io::{stdout, Write};
 
 use anyhow::Context;
+use const_format::concatcp;
 
 use crate::cloud;
 use crate::commands::ExitCode;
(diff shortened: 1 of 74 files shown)


src/commands/ui.rs
```diff
@@ -111,17 +113,21 @@ fn _get_local_ui_url(cmd: &UI, cfg: &edgedb_tokio::Config) -> anyhow::Result<Str
...
                 Ok(reqwest::StatusCode::NOT_FOUND) => {
-                    print::error("Web UI not served correctly by specified EdgeDB server.");
+                    print::error(concatcp!(
+                        "Web UI not served correctly by specified ",
+                        BRANDING,
+                        " server."
+                    ));
                     print::echo!(
-                        "  If you have EdgeDB 2.0 and above, try running the \
+                        "  If you have {BRANDING} 2.0 and above, try running the \
```
msullivan, line 122: Let's drop the first half of this sentence. "Gel 2.0" doesn't really exist and nobody is running EdgeDB 1.0 anymore.

src/portable/link.rs
```diff
@@ -422,15 +423,15 @@ async fn prompt_conn_params(
...
-    eprintln!("  edgedb instance unlink -I {:?} --force", name);
+    eprintln!("  {BRANDING_CLI_CMD} instance unlink -I {:?} --force", name);
 }
 
 pub fn unlink(options: &Unlink) -> anyhow::Result<()> {
     let name = match instance_arg(&options.name, &options.instance)? {
         InstanceName::Local(name) => name,
         inst_name => {
             return Err(anyhow::anyhow!(
-                "cannot unlink cloud instance {}.",
```
msullivan, line 433: I feel like I'd prefer not to change all of the unqualified "cloud" references to "Gel Cloud". But also I don't feel strongly about it, especially because I'm sure preparing this PR was already enough of a nightmare without having to undo some of it.

src/portable/windows.rs
```diff
@@ -747,7 +750,7 @@ pub fn list_versions(options: &options::ListVersions) -> anyhow::Result<()> {
     } else {
         log::warn!(
             "WSL distribution is not installed, \
-                   so no EdgeDB server versions are present."
+                   so no {BRANDING} server versions are present."
```
msullivan, line 753: Maybe we shouldn't duplicate this one message 10 times?

msullivan: Doesn't really need to be fixed now though

msullivan (approved):
Thanks!
What a pain.

src/portable/windows.rs
mmastrac: I'm hoping to fix this in a later PR whilest fixing the duplicated print functions but totally agree.

src/commands/ui.rs
mmastrac:
```suggestion
                        "  Try running the \
```

PR #1394, opened 2024-11-12
WIP: Improve error message in the case where project is not initialized

mmastrac:


src/options.rs
```diff
@@ -901,9 +901,9 @@ impl Options {
 
                 if e.is::<ClientNoCredentialsError>() {
                     let project_dir = get_project_path(None, true).await?;
-                    let message = if project_dir.is_some() {
+                    let message = if let Some(project_dir) = project_dir {
                         format!(
-                            "project is not initialized and no connection options \
+                            "project is not initialized at {project_dir:?} and no connection options \
```
1st1, line 906: Perhaps we need to clarify that it's not initialized in the specific directory or in any of the parent directories

mmastrac: Agreed -- although I think I should close this PR in favour of the more verbose project init message discussed on Slack.

aljazerzen: I think I've made it clear what's the problem in the error message: project has been found, but it is not initialized

PR #1398, opened 2024-11-13
Install new executable name (and link old executable)

mmastrac:
This installs (when `--features gel` is passed) to both edgedb and gel executables. The alternative executable is hard linked, symlinked or copied (in that order of preference).

Note that the `gel` executable is installed even w/o `--features gel` right now, but it is installed as the "alternate" executable.

We also check the name of the executable that is invoked and warn if the alternative is there. If the two executables are out of sync (determined by checking executable length), we warn as well.

### DIFF
--- a/src/branding.rs
+++ b/src/branding.rs
@@ -1,8 +1,40 @@
 #![allow(unused)]
 
-pub const BRANDING: &str = "EdgeDB";
-pub const BRANDING_CLI: &str = "EdgeDB CLI";
-pub const BRANDING_CLOUD: &str = "EdgeDB Cloud";
-pub const BRANDING_CLI_CMD: &str = "edgedb";
+/// The product name.
+pub const BRANDING: &str = if cfg!(feature = "gel") {
+    "Gel"
+} else {
+    "EdgeDB"
+};
+/// The CLI name.
+pub const BRANDING_CLI: &str = const_format::concatcp!(BRANDING, " CLI");
+/// The cloud name.
+pub const BRANDING_CLOUD: &str = const_format::concatcp!(BRANDING, " Cloud");
+/// The CLI command name.
+pub const BRANDING_CLI_CMD: &str = if cfg!(feature = "gel") {
+    "gel"
+} else {
+    "edgedb"
+};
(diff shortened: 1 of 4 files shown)


src/branding.rs
```diff
@@ -1,8 +1,44 @@
...
+/// The executable file name for the CLI alternative.
+pub const BRANDING_CLI_CMD_ALT_FILE: &str = if cfg!(windows) {
+    concatcp!(BRANDING_CLI_CMD_ALT, ".exe")
+} else {
+    BRANDING_CLI_CMD_ALT
+};
+
+/// The WSL distribution name.
 pub const BRANDING_WSL: &str = "EdgeDB.WSL.1";
+
```
1st1, line 42: Do we want to update this one too?

src/cli/install.rs
```diff
@@ -595,6 +588,111 @@ fn _main(options: &CliInstall) -> anyhow::Result<()> {
...
+    let exe_dir = exe_path.parent().unwrap();
+    let old_executable = exe_dir.join(BRANDING_CLI_CMD_ALT_FILE);
+    let new_executable = exe_dir.join(BRANDING_CLI_CMD_FILE);
+    log::debug!("exe_path: {exe_path:?}");
+    log::debug!("old_executable: {old_executable:?}");
+    log::debug!("new_executable: {new_executable:?}");
+
+    if exe_path.file_name().unwrap() == BRANDING_CLI_CMD_ALT_FILE {
+        if new_executable.exists() {
+            log::warn!("{exe_path:?} is the old name for the {BRANDING_CLI_CMD_FILE} executable. \
```
1st1, line 661: I'd surround names with quotes for better readability

1st1 (approved):

src/branding.rs
mmastrac: TBH I don't know -- I'm not terribly familiar with WSL. Is there someone who understands what that change might affect? I am not sure if we'd orphan older databases on Windows.

mmastrac:
Punted Windows issues here: https://github.com/edgedb/edgedb-cli/issues/1401

PR #1400, opened 2024-11-14
Add support for SQL as input query language

elprans:
The new `\set language sql` toggle switches the CLI into SQL mode.  The
`query` subcommand also gained the `--input-language` argument here,
though it only works with `--tab-separated` output mode.

### DIFF
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -10,7 +10,9 @@ use tokio::sync::oneshot;
 
 use edgedb_errors::{ClientError, ProtocolEncodingError};
 use edgedb_errors::{Error, ErrorKind};
-use edgedb_protocol::common::{RawTypedesc, State as EdgeqlState};
+use edgedb_protocol::common::{
+    InputLanguage as ServerInputLanguage, IoFormat, RawTypedesc, State as EdgeqlState,
+};
 use edgedb_protocol::model::Duration as EdbDuration;
 use edgedb_protocol::model::Uuid;
 use edgedb_protocol::server_message::TransactionState;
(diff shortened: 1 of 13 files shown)


1st1:
We should have `\sql` and `\edgeql` aliases.

.gitignore
```diff
@@ -1,6 +1,7 @@
 /tmp
 /.vagga
 /target
+/.cargo
```
1st1, line 4: Is intentional?

src/repl.rs
```diff
@@ -284,7 +294,12 @@ impl State {
             _ => format!("{current_database}"),
         };
 
-        let prompt = format!("{location}{txstate}> ");
+        let lang = match self.input_language {
+            InputLanguage::EdgeQL => "",
+            InputLanguage::SQL => "[sql]",
```
1st1, line 299: Should we display a tag for EdgeQL too?

src/repl.rs
```diff
@@ -397,6 +412,26 @@ impl std::str::FromStr for InputMode {
     }
 }
 
+impl std::str::FromStr for InputLanguage {
+    type Err = anyhow::Error;
+    fn from_str(s: &str) -> Result<InputLanguage, anyhow::Error> {
+        match s {
+            "edgeql" => Ok(InputLanguage::EdgeQL),
```
1st1, line 419: Can you lowercase before the match?

1st1 (approved):

mmastrac (approved):
Looks good, but the shared-client-testcases changes should probably be reverted.

src/repl.rs
elprans: SQL in CLI is marginal, I would not think that cluttering the default mode is worth it.

src/repl.rs
elprans: OK.

PR #1404, opened 2024-11-21
Ensure that paths are branded

mmastrac:


msullivan (changes requested):
I don't think we will be able to just update these in place, if users already have edgedb installed. We can't lose all the previous local instances and edgedb installs.

I think we need to either:
 - Leave the directories named `edgedb` for now
 - Use `edgedb` if it exists, otherwise use `gel`
 - Have a migration path of some sort that involves moving the directories

The last option is a bit complicated since there may be active instances running that we'll need to stop first.

mmastrac:
> Leave the directories named `edgedb` for now

Agree -- I will add an explicit comment here so that the directory names are extracted, but are not changed during rebranding. It'll make it easier to fix in the future.

src/branding.rs
```diff
@@ -42,3 +42,11 @@ pub const BRANDING_WSL: &str = "EdgeDB.WSL.1";
...
 /// The display name for the configuration file.
 pub const CONFIG_FILE_DISPLAY_NAME: &str = "`gel.toml` (or `edgedb.toml`)";
+
+/// The database/OS username.
+// TODO: This should become "admin"
+pub const BRANDING_USERNAME: &str = "edgedb";
+
+/// The OS pathname for data files.
+// TODO: Should this be "gel" as well?
+pub const BRANDING_PATH: &str = if cfg!(windows) { "EdgeDB" } else { "edgedb" };
```
mmastrac, line 52:
```suggestion
// TODO: Should this be "gel" as well?
// WARNING: This requires will require a migration strategy if we decide to change it
pub const BRANDING_PATH: &str = if cfg!(windows) { "EdgeDB" } else { "edgedb" };
```

PR #1408, opened 2024-11-25
Send query tag "gel/cli" or "gel/repl"

fantix:
- [x] Depends on edgedb/edgedb-rust#364

### DIFF
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -185,6 +185,7 @@ impl State {
         let mut params = self.conn_params.clone();
         params.branch(branch)?;
         let mut conn = params.connect_interactive().await?;
+        conn.set_tag("gel/repl");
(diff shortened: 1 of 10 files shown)


fantix:
The test fails because the current server (protocol 3.0 without edgedb/edgedb#8036) is not correctly processing annotations.

elprans (approved):

src/repl.rs
```diff
@@ -185,6 +185,7 @@ impl State {
         let mut params = self.conn_params.clone();
         params.branch(branch)?;
         let mut conn = params.connect_interactive().await?;
+        conn.set_tag("gel/repl");
```
mmastrac, line 188: REPL_QUERY_TAG?

mmastrac (approved):
LGTM, one small item

PR #1417, opened 2024-12-02
Bump version to 6.1.0-dev

fantix:
Also add temporary allow(dead_code) refs #1416

### DIFF
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -136,6 +136,7 @@ pub fn binary_path() -> anyhow::Result<PathBuf> {
     Ok(dir.join(BRANDING_CLI_CMD_FILE))
 }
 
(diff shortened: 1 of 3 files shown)


src/platform.rs
```diff
@@ -136,6 +136,7 @@ pub fn binary_path() -> anyhow::Result<PathBuf> {
     Ok(dir.join(BRANDING_CLI_CMD_FILE))
 }
 
+#[allow(dead_code)]
```
scotttrinh, line 139: Or, I guess this is so trivial we can delete it until we need it?

scotttrinh (approved):

src/platform.rs
fantix: sure!

PR #1423, opened 2024-12-05
Fix "admin" role issue with different servers

fantix:
* Check 6.0 alpha versions correctly (early 6.0 nightlies are no longer supported)
* Still bootstrap 5.x and earlier instances with "edgedb" as the default default_user

Refs #1412, #1413

### DIFF
--- a/src/portable/create.rs
+++ b/src/portable/create.rs
@@ -442,10 +442,14 @@ pub fn bootstrap(
     fs::create_dir_all(&tmp_data).with_context(|| format!("creating {:?}", &tmp_data))?;
 
     let password = generate_password();
+    let version = info.get_version()?;
     let script = bootstrap_script(
         user,
         &password,
-        if info.get_version()? >= &Build::from_str("6.0-dev.9024+4b89273").unwrap() {
+        if &Build::from_str("6.0-alpha.2+8fb3f01").unwrap() <= version
+            && version < &Build::from_str("6.0-dev.8321+d6c575a").unwrap()  // lowest nightly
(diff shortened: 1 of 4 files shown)


src/portable/create.rs
```diff
@@ -445,7 +447,7 @@ pub fn bootstrap(
     let script = bootstrap_script(
         user,
         &password,
-        if info.get_version()? >= &Build::from_str("6.0-dev.9024+4b89273").unwrap() {
+        if info.get_version()? >= &Build::from_str("6.0-alpha.2+8fb3f01").unwrap() {
```
msullivan, line 450: Do we want the `+8fb3f01` bit?

msullivan (approved):

src/portable/create.rs
fantix: no we don't ;)

src/portable/project.rs
```diff
@@ -410,6 +410,14 @@ pub fn get_default_branch_name(version: &Specific) -> String {
...
 }
 
+pub fn get_default_user_name(version: &Specific) -> &'static str {
+    if version.major >= 6 {
+        "admin"
+    } else {
+        "edgedb"
+    }
+}
+
```
mmastrac, line 420: Could we move this to branding.rs as `BRANDING_DEFAULT_USERNAME_LEGACY` (edgedb) and `BRANDING_DEFAULT_USERNAME`? This helps with searchability of the codebase as this exact string has other purposes.

mmastrac (approved):
LGTM, one small request

PR #1425, opened 2024-12-11
Support printing postgis data.

dnwpark:
No special handling for Box2D or Box3D is included. It may be useful to do this in the future.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,6 +150,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7992085ec035cfe96992dd31bfd495a2ebd31969bb95f624471cb6c0b349e571"
 
 [[package]]
+name = "approx"
+version = "0.5.1"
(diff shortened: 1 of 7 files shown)


msullivan (approved):
Looks reasonable.
Were you able to test it against a running instance?

Supporting inputting geometry would be good also, but is a lower-priority follow-up.

dnwpark:
@msullivan 

The tests run on a local instance and pass.

The query `with module ext::postgis select <geometry>'POINT(0 1)';` returns
```
{POINT(0 1)}
```

The box types have some special handling, which doesn't get properly handled in the CLI, which might be confusing.

The query `with module ext::postgis select <box2d>'box(0 0, 1 1)';` returns:
```
{POLYGON((0 0,0 1,1 1,1 0,0 0))}
```

The query `with module ext::postgis select <str><box2d>'BOX(0 0, 1 1)';` (note cast to str) returns:
```
{'BOX(0 0,1 1)'}
```

The javascript binding PR has some special checking for the boxes, but I haven't reproduced it here.

PR #1428, opened 2024-12-12
Add rudimentary support for rendering SQLRow typedesc

1st1:
In tandem with https://github.com/edgedb/edgedb-rust/pull/368


msullivan (approved):
Are we going to want it to be a table printer eventually?

1st1:
> Are we going to want it to be a table printer eventually?

100% we must before 6.0. This is just a temporary rendering to let us debug / develop

aljazerzen (approved):

PR #1440, opened 2025-01-15
Refactor portable commands and their options

aljazerzen:
Module `portable` contains many different commands. I've segmented them into `server`, `instance`, `extension` and `project`. They don't define their options (aka clap command / args) in `portable::options` anymore, but instead have them as close to where the main `run` function is located.

This is not perfect yet, but it makes more sense to me now as it did before.


src/repl.rs
```diff
@@ -34,8 +34,8 @@ pub const FAILURE_MARKER: &str = "[tx:failed]";
 #[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
 #[value(rename_all = "lowercase")]
 pub enum InputLanguage {
-    EdgeQL,
-    SQL,
+    EdgeQl,
+    Sql,
 }
```
msullivan, line 39: Sigh. Do linters feel really strongly about this one? I think this is an anti-improvement

msullivan (approved):

src/repl.rs
aljazerzen: No, not strongly. We could of course disable this lint. It was mostly automatic. And I like that all things are cased the same.

PR #1444, opened 2025-01-21
Add a retry to dump-restore tests

aljazerzen:
It was failing on my machine, because `dump --all` queried the databases
which were then deleted by another test.
Ideally, we'd isolate tests better, but that would be too costly.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -35,11 +35,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1730270645,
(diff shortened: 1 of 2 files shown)


msullivan (approved):
Ugh

msullivan:
How many tests do deletes? Could we isolate just those?

aljazerzen:
I've added a comment: if we want isolation, we could spawn a separate edgedb-server, so `dump --all` is not interfering with any other branches. But that would be slow, probably.

Issue #1418, opened 2024-12-03
Cannot update a stable version to nightly on macOS


scotttrinh:
### Steps to reproduce

On macOS:

1. `edgedb project init` and select the current stable (`5.7` as of today)
2. `edgedb project upgrade --to-nightly`

Observe:

```
$ edgedb project upgrade --to-nightly
Downloading package...
00:00:01 [====================] 51.77 MiB/51.77 MiB 34.88 MiB/s | ETA: 0s
Successfully installed 6.0-dev.9043+4c05a1a
The schema is forward compatible. Ready for upgrade.
Upgrading to a major version 6.0-dev.9043+4c05a1a
Dumping the database...
Starting dump for database `main`...
  Finished dump for `main`. Total size: 0 B                                                                                                                                                                                                   Restoring the database...
[edgedb] CRITICAL 15986 - 2024-12-03T14:23:18.475 postgres: the database system is starting up
edgedb error: cannot restore "tt": Connection error: ClientConnectionFailedError: cannot establish connection for 300s
To undo run:
  edgedb instance revert -I "tt"
```

### Notes

Initializing with a nightly instance works fine, so it's something with the actual upgrade flow that is breaking.

Running the `RUST_LOG=trace`, you get this:

```
  Finished dump for `main`. Total size: 0 B                                                                                                                            [2024-11-26T15:42:36Z TRACE mio::poll] deregistering event source from poller
[2024-11-26T15:42:36Z TRACE mio::poll] deregistering event source from poller
[2024-11-26T15:42:36Z INFO  edgedb::portable::upgrade] Stopping instance before executable upgrade
[2024-11-26T15:42:36Z TRACE mio::poll] registering event source with poller: token=Token(1), interests=READABLE
[2024-11-26T15:42:36Z INFO  edgedb::process] Running detect launchd: Command { std: LANG="C.UTF-8" LC_ALL="C.UTF-8" LC_CTYPE="UTF-8" "/bin/launchctl" "print-disabled" "gui/501", kill_on_drop: false }
[2024-11-26T15:42:36Z TRACE mio::poll] registering event source with poller: token=Token(4485832320), interests=READABLE | WRITABLE
[2024-11-26T15:42:36Z TRACE mio::poll] registering event source with poller: token=Token(4485832576), interests=READABLE | WRITABLE
[2024-11-26T15:42:36Z DEBUG edgedb::process] Writing pid file None (pid: 60666)
[...]

PR #1450, opened 2025-01-23
macOS: clear runstate dir after instance stop

fantix:
Refs edgedb/edgedb#8248, #1418

Manually tested with the repro in #1418

### DIFF
--- a/src/portable/macos.rs
+++ b/src/portable/macos.rs
@@ -294,8 +294,8 @@ pub fn stop_and_disable(name: &str) -> anyhow::Result<bool> {
     // Clear the runstate dir of socket files and symlinks - macOS wouldn't
     // delete UNIX domain socket files after server shutdown, which may lead to
     // issues in upgrades
-    let dir_path = runstate_dir(name)?;
-    for entry in fs::read_dir(dir_path)? {
+    #[cfg(unix)]
+    for entry in fs::read_dir(runstate_dir(name)?)? {
         let entry = entry?;
         let path = entry.path();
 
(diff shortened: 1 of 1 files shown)


msullivan (commented):
What goes in the runstate dir? When running the server locally in dev mode, the socket is in the postgres data dir, which would obviously be wrong here.

We're 100% sure this is safe?

src/portable/macos.rs
```diff
@@ -288,6 +289,12 @@ pub fn stop_and_disable(name: &str) -> anyhow::Result<bool> {
         log::info!("Removing unit file {}", unit_path.display());
         fs::remove_file(unit_path)?;
     }
+
+    // Clear the runstate dir - macOS wouldn't delete UNIX domain socket
+    // files after server shutdown, which may lead to issues in upgrades
+    fs::remove_dir_all(runstate_dir(name)?).ok();
```
mmastrac, line 295: I think we should filter and remove only domain sockets if that is the problem. Something like this?

```
    for entry in fs::read_dir(dir_path)? {
        let entry = entry?;
        let path = entry.path();

        if let Ok(metadata) = path.metadata() {
            if metadata.file_type().is_socket() {
                fs::remove_file(&path)?;
            }
        }
    }
```

fantix: Yup, that'll do! Do you mind pushing that to the PR? I'm kinda in the middle of home moving. Thank you!

msullivan (approved):

fantix:
Thank you all!

PR #1452, opened 2025-01-28
Pass default branch when bootstrapping instances

aljazerzen:


msullivan (approved):
Seems probably fine, but what is the specific rationale for the change?

aljazerzen:
I've updated the original PR comment after you've asked that question.

PR #1453, opened 2025-01-29
Colors on light terminals

aljazerzen:
Motivation for this PR is printing on terminals with light color scheme.

Before:

image

After:

image

To be honest, I'm not sure if I improved the situation.

---

The code has been improved though. Instead of using `colorful::Colorful` all over the codebase, I've replaced the usages with our `print::Highlight`, which applies colors from the current theme. That theme is:
- `None` if we detect that terminal does not support colors, has NO_COLOR or is not even a terminal,
- `Some(light_theme)` if [terminal-light](https://docs.rs/terminal-light/latest/terminal_light/) says that the "luma" of the terminal is >0.6,
- `Some(light_theme)` otherwise.

So light theme can now be set here: https://github.com/edgedb/edgedb-cli/pull/1453/files#diff-fa9f9b1716877750c185f94c61aba7d11d29aa88968cd9e5eb937954bf702da1R62-R81

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1237,6 +1237,7 @@ dependencies = [
  "term 1.0.1",
  "termcolor",
  "termimad",
+ "terminal-light",
  "terminal_size",
  "test-case",
  "test-utils",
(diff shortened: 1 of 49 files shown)


1st1:
> To be honest, I'm not sure if I improved the situation.

That's something we can quickly do after this lands by just adjusting the exact color mapping, right?

aljazerzen:
Yes. All printed colors can now be changed via the `print::colors::THEME`.

aljazerzen:
Part of the problem with the ugly screenshots is that I've used the default "Black on white" theme in KDE's Konsole. We might need input from someone who seriously uses light color scheme in their terminal.

1st1:
> Part of the problem with the ugly screenshots is that I've used the default "Black on white" theme in KDE's Konsole. We might need input from someone who seriously uses light color scheme in their terminal.

I can play with that once I'm through my todo a bit. You can land this PR - don't wait on the exact colors approval

msullivan (approved):

PR #1457, opened 2025-01-31
Refactor filesystem watcher

aljazerzen:
- extract filesystem watcher into its own module,
- improve it so it emits the set of changed paths


src/watch/fs_watcher.rs
```diff
@@ -0,0 +1,94 @@
...
+                        return changed_paths;
+                    }
+                    if changed_paths.is_empty() {
+                        timeout = None;
+                    } else {
+                        timeout = Some(STABLE_TIME);
+                    }
+                },
+            }
+        }
```
msullivan, line 64: Add a comment about the algorithm/rationale here.
(It is waiting for quiescence, right?)

msullivan (approved):

PR #1458, opened 2025-01-31
Watch files

aljazerzen:


src/watch/main.rs
```diff
@@ -16,7 +16,11 @@ use crate::print::AsRelativeToCurrentDir;
 use crate::watch::WatchCommand;
 
 #[tokio::main(flavor = "current_thread")]
-pub async fn run(options: &Options, _watch: &WatchCommand) -> anyhow::Result<()> {
+pub async fn run(options: &Options, cmd: &WatchCommand) -> anyhow::Result<()> {
+    if cmd.files {
+        return super::files::run().await;
+    }
```
scotttrinh, line 22: I think we want _both_ behaviors, right? The existing behavior of `watch` and also this file watcher behavior.

PR #1460, opened 2025-01-31
Implement a tabular output format and make it the default for SQL

### LINKED ISSUE #1451: Tabular printing for SQL mode
Currently we produce a list of free objects, sometimes with janky autogenerated column names:
```
_localdev:main[sql]> select 1,2,3;
{(col~1 := 1, col~2 := 2, col~3 := 3)}
```

We probably want to print this in some sort of nice tabular format instead, like `psql` does.
I think we already have a tabular printing library we use.

msullivan:
It uses the Printer/Formatter mechanism, since that is what actually
has fully implemented formatters.

I would like to take advantage of the coloring that Printer/Formatter
does, but the table library we use breaks if there are control
characters.  I have a plan I might do to fork the library and get
something working, but that would be a follow-up.

This adds a new `tabular` `output-format` that works for both edgeql
and SQL. We then add a `sql-output-format` mode that defaults to
`tabular`, but it can still be set back to `default`.

Fixes #1451.

### DIFF
--- a/src/print/mod.rs
+++ b/src/print/mod.rs
@@ -35,7 +35,6 @@ use formatter::Formatter;
 use native::FormatExt;
 use stream::Output;
 
-// XXX?
 use crate::table::{self, Cell, Row, Table};
 use gel_protocol::value::Value;
 
@@ -329,14 +328,18 @@ where
     I: FormatExt + Into<Value>,
     E: fmt::Debug + Error + 'static,
 {
+    // This is kind of hacky, but for ~performance~ and to avoid
+    // needing to pass around enough config info to recreate new ones,
+    // we repeatedly invoke a single Printer and then pull the strings
+    // out and put them in a table we are building.
     let mut buf = String::new();
     let mut prn = Printer {
-        // colors,
+        // We don't use colors yet because the table library gets
+        // confused.
         colors: false,
         indent: config.indent,
         expand_strings: config.expand_strings,
-        max_width: usize::max_value(), // lol
-        // max_width,
+        max_width: usize::MAX,
         implicit_properties: config.implicit_properties,
         max_items: config.max_items,
         max_vector_length: config.max_vector_length,
--- a/src/print/mod.rs
+++ b/src/print/mod.rs
@@ -30,15 +30,14 @@ use gel_errors::display::display_error;
 use crate::branding::BRANDING_CLI_CMD;
 use crate::repl::VectorLimit;
 
-use buffer::{Delim, Exception, UnwrapExc, WrapErr, fix_infallible};
+use buffer::{fix_infallible, Delim, Exception, UnwrapExc, WrapErr};
 use formatter::Formatter;
 use native::FormatExt;
 use stream::Output;
 
 // XXX?
-use gel_protocol::value::Value;
 use crate::table::{self, Cell, Row, Table};
-
+use gel_protocol::value::Value;
 
 #[derive(Snafu, Debug)]
 #[snafu(context(suffix(false)))]
@@ -287,7 +286,6 @@ where
                     };
                     table_row.push(Cell::new(&get_printer_string(prn)?));
                 }
-
             }
             Value::Object { shape, fields } => {
                 for (s, vi) in shape.elements.iter().zip(fields) {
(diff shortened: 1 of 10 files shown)


msullivan:
Screenshot from 2025-01-31 15-15-27

1st1:
Can you show how the output looks like if the terminal isn't wide enough to fit the table?

msullivan:
> Can you show how the output looks like if the terminal isn't wide enough to fit the table?

It looks bad!

msullivan:
Screenshot from 2025-01-31 15-23-45

Postgres also looks pretty bad when it doesn't fit.

1st1:
Yeah, i was just checking if somehow we look even worse than postgres, but this is pretty much the same. Looks good to me.

I the future, adding coloring would be one improvement, detecting overly long/wide output and launching pager would be another.

1st1:
Do you truncate super long strings/bytes? Otherwise long scalars like some postgis/vector stuff?

msullivan:
> Do you truncate super long strings/bytes? Otherwise long scalars like some postgis/vector stuff?

We truncate vectors but not strings, I think; (same behavior as our normal printer).

src/print/mod.rs
```diff
@@ -226,6 +229,184 @@ where
...
+        v,
+        Value::Int16(_)
+            | Value::Int32(_)
+            | Value::Int64(_)
+            | Value::Float32(_)
+            | Value::Float64(_)
+            | Value::BigInt(_)
+            | Value::Decimal(_)
+    )
+}
```
mmastrac, line 256: This should probably get upstreamed at some point -- seems like a useful helper.

1st1:
We'll need to start truncating for both modes, should be a configurable setting... not as part of this or obviously, and isn't a high priority

src/print/mod.rs
```diff
@@ -226,6 +229,184 @@ where
...
 }
 
+fn get_printer_string<E>(
+    prn: &mut Printer<&mut String>,
+) -> Result<String, Exception<PrintError<E, Infallible>>>
+where
+    E: fmt::Debug + Error + 'static,
+{
+    prn.commit().map_err(fix_infallible)?;
+    prn.flush_buf().map_err(fix_infallible)?;
```
mmastrac, line 239: This feels a little off -- I can take a look at the CLI code a bit closer but the nested Result<Exception<Error<E>>> suggests we may need to refactor a bit.

mmastrac (commented):
Looks good but I think the error handling can be simplified. I can try to take a closer look.

src/print/mod.rs
msullivan: `Exception`  comes specifically from the `Printer`, and is used to signal when it couldn't fit things in one line and needs help from the outside to reflow. That doesn't match *this* use case (where we pass MAX_INT as the max line length and where `Printer.flow` is always set to false), but it's pretty deeply embedded in all the Printer function types.

It's all pretty hinky, especially in this use case. The printers return `Result<..., Exception<T::Error>>`, where `T` is the output stream type; but for this, since the output is into a `String`, the output stage can't actually fail and so the Error is Infallible.

And so we have a `Result<..., buffer::Exception<Infallible>>` where the `DisableFlow` part of the `buffer::Exception` shouldn't ever happen, but we can't verify that statically.

An earlier take just used `unwrap` on all of these, but I felt bad about that.
Maybe I should just go back to it, though.
(Since the caller outside panics on a `DisableFlow` anyway!)

src/print/mod.rs
msullivan: Another option would be to have a never-failing `unwrap_infallible` method, and then do `.unwrap_exc().unwrap_infallible()`, to make it clear that we are only *really* doing an unsafe unwrapping of the `DisableFlow` case, not of any other potential errors.

That's noisier but still gets the errors out of the `to_cell` and `get_printer_string` functions.

msullivan: Though, also: who bothers error checking your writes to stdout anyway? We don't anywhere else in our codebase.

mmastrac (approved):
Latest changes lgtm

PR #1464, opened 2025-02-05
Watch files

aljazerzen:


src/watch/files.rs
```diff
@@ -0,0 +1,67 @@
...
+
+        // run all matching scripts
+        for (glob_str, matcher, script) in &globs {
+            let is_matched = changed_paths.iter().any(|x| matcher.is_match(x));
+            if !is_matched {
+                continue;
+            }
+
+            print::msg!("{}", format!("--- watch.files \"{glob_str}\" ---").muted());
+            print::msg!("{}", format!("  $ {script}").muted());
```
scotttrinh, line 57: Could we show which files matched this glob? Maybe just as debug-level output?

src/watch/main.rs
```diff
@@ -16,7 +16,11 @@ use crate::print::AsRelativeToCurrentDir;
 use crate::watch::WatchCommand;
 
 #[tokio::main(flavor = "current_thread")]
-pub async fn run(options: &Options, _watch: &WatchCommand) -> anyhow::Result<()> {
+pub async fn run(options: &Options, cmd: &WatchCommand) -> anyhow::Result<()> {
+    if cmd.files {
+        return super::files::run().await;
+    }
```
scotttrinh, line 22: I think we want `watch` to respond to both schema changes _and_ file changes, rather than either/or.

aljazerzen: Right now, watching files has a few different code paths than watching schema.

If a project has the following config: 

```
[watch]
"dbschema/**" = "gel migration apply --dev-mode"
"gel.toml" = "gel migration apply --dev-mode"
```

... then `gel watch --files` would be similar to current behavior of `gel watch`. It would be different in:
- `gel watch` prints a few more messages about what's being done to your database,
- `gel watch` has a nicer status bar,
- on schema errors, `gel watch` will configure the database to reply with errors to all queries, so the user (who might not be looking at `gel watch` output) notices that there is a problem with the schema. When the problem is resolved, or `gel watch` exists, the "always reply with error" config of the database is cleared.

If we manage to iron out all of these wrinkles, then watching "schema" and "watching" files can be unified. If not, I'll spend the additional effort of merging the two, so we can trigger both from the same watch command.

aljazerzen:
It's debt from merging a previous PR without waiting for tests to finish :sweat_smile:

src/watch/main.rs
aljazerzen: > Which wrinkles?

The three points I've outlined above.

---

I'm saying that when I'm finished with this feature, users will be able to run just `gel watch` for both. The question is just if we are able to squeeze the "watching schema" behavior into the same implementation of "watching files". Or if we need to maintain two separate, slightly different codepaths.

aljazerzen: > It's also a little surprising that you'd have a watch toml table, but it only runs if you start watch-mode in this "file" mode.

That's current behavior, I'm not really sure why it is also watching gel.toml. Maybe because `project.schema-dir` might change? I don't know.

src/watch/files.rs
aljazerzen: I've added a `--verbose` flag, with the following output:

image

aljazerzen:
@scotttrinh This is now ready to be tried out.

Things to test:

- different glob patterns (`*.txt`, `**`, `*.{log,txt}`),
- a single change can trigger multiple scripts,
- multiple changes can trigger multiple scripts,
- multiple changes can trigger a single script (for example a `*.txt),
- `--migrate`,
- `--retry-sec=10` retries failed scripts (and only failed scripts) after 10 sec,
- `--retry-sec=10` retries failed scripts on new changes,
- `--verbose` reports files that triggered the script,

Do messages make sense? Should something be bold/muted?

aljazerzen:
Things to test:

- different glob patterns (`*.txt`, `**`, `*.{log,txt}`),
- a single change can trigger multiple scripts,
- multiple changes can trigger multiple scripts,
- multiple changes can trigger a single script (for example a `*.txt),
- `--migrate`,
- `--retry-sec=10` retries failed scripts (and only failed scripts) after 10 sec,
- `--retry-sec=10` retries failed scripts on new changes,
- `--verbose` reports files that triggered the script,

src/watch/mod.rs
```diff
@@ -1,15 +1,252 @@
...
+        connector: options.create_connector().await?,
+        is_force_database_error: false,
+    };
+
+    // determine what we will be watching
+    let mut matchers = assemble_matchers(cmd, &ctx)?;
+
+    if cmd.migrate {
+        print::msg!(
+            "Hint: --migrate will apply any schema changes to the database. When done use:"
```
scotttrinh, line 49: Maybe a bit more clear to point out that we're watching the file itself for changes? Otherwise "any schema changes" is a bit too ambiguous to understand what is happening, or what someone should do to trigger a "schema change".

How about:

```suggestion
            "Hint: --migrate will apply any changes in your schema module files to the database on save. When ready to commit your changes, use:"
```

src/watch/mod.rs
```diff
@@ -1,15 +1,252 @@
...
+            print::msg!(
+                "{}",
+                format!(
+                    "--- {}: {} ---",
+                    matcher.glob,
+                    matcher.target.to_string().muted()
+                )
+            );
+            if cmd.verbose {
+                print::msg!("{}", format!("  triggered by: {reason}").muted());
```
scotttrinh, line 141: I don't have a strong feeling here, but I don't think we _need_ any special coloring for verbose output. Feel free to keep it though, just putting it out there.

scotttrinh (commented):
Some initial feedback. Will also test the functionality and report back any feedback on using it as a tool.

src/migrations/dev_mode.rs
```diff
@@ -40,21 +40,33 @@ pub async fn migrate(cli: &mut Connection, ctx: &Context, bar: &ProgressBar) ->
             if !migrations.is_empty() {
                 bar.set_message("applying migrations");
                 apply_migrations(cli, migrations, ctx, false).await?;
+                bar.println("Migrations applied");
```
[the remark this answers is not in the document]
aljazerzen: Yes, `set_message` appears in the progress bar, after the `⣾` spinner. It used to be lower case, so I kept it.

It does look weird in the source code.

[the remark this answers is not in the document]
scotttrinh: Yeah, this is not a blocker at all, so I'm fine with shipping this as-is and we can talk updating it as needed based on our review of developer experience from a more holistic perspective.

src/watch/mod.rs
aljazerzen: Maybe I should cut the files after 10th to: "a.log, b.log, +3 others" and drop the `.muted()`.

scotttrinh: Yeah, but I think making it muted doesn't make it any less text, or take up any less screen real-estate. But it certainly would help make the non-verbose text stand out more. Not sure if that's a desirable outcome, tbh, so I'm fine to leave it as-is. Just wanted to raise the question since you asked for feedback on colors, etc. FWIW, I'm probably more minimalist than most about colorful output.

scotttrinh: In my opinion, if `--verbose` isn't telling me everything I want to know, it's a waste. Maybe that means we should show this truncated version in the non-verbose mode and the full-output in verbose mode?

scotttrinh: FWIW, I'm also 100% fine with not printing the glob matches that triggered the change, too. Just seemed like a helpful bit of information since our watch mode is already outputting activity text.

aljazerzen: > if --verbose isn't telling me everything I want to know, it's a waste

That's true.

> Maybe that means we should show this truncated version in the non-verbose mode and the full-output in verbose mode?

I don't feel like matched files are really all that useful in a normal workflow. You know which files you've just edited. You are interested in what is now being triggered.

>  the glob matches that triggered the change

I am printing globs that have been triggered.

PR #1473, opened 2025-02-12
Don't quote strings in tabular mode

msullivan:
Tabular mode isn't really for precision, anyway. This is what postgres
does, and I think it looks better.

Screenshot from 2025-02-12 13-10-47


1st1 (approved):
much better

mmastrac (approved):
LGTM, slick

PR #1475, opened 2025-02-14
Watch parallel execution

aljazerzen:
Makes gel watch execute scripts in parallel instead of sequentially.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -242,6 +242,11 @@ impl Native {
         self
     }
 
+    pub fn current_dir(&mut self, path: impl AsRef<Path>) -> &mut Self {
+        self.command.current_dir(path);
+        self
+    }
+
     pub fn env(&mut self, key: impl AsRef<OsStr>, val: impl AsRef<OsStr>) -> &mut Self {
         self.envs.insert(
             key.as_ref().to_os_string(),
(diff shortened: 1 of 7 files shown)


msullivan:
Add a commit message?

src/process.rs
```diff
@@ -652,6 +655,35 @@ impl Native {
...
+    #[cfg(windows)]
+    async fn signal_loop_tokio(&self, _: u32) -> io::Result<()> {
+        // on windows Ctrl+C signals are propagated automatically and no other
+        // signals are supported, so there is nothing to do here
+        wait_forever().await;
+        Ok(())
+    }
+
+    #[cfg(unix)]
+    async fn signal_loop_tokio(&self, pid: u32) -> io::Result<()> {
```
msullivan, line 667: Huh, I'm surprised that this is necessary. SIGINT gets sent to everything that is in the process group that owns the terminal, if I remember correctly; do the children get spawned in other process groups, I guess?

src/watch/migrate.rs
```diff
@@ -1,26 +1,67 @@
...
+    ) {
+        while let Some(order) = ExecutionOrder::recv(&mut input).await {
+            order.print(&matcher, self.ctx.as_ref());
+
+            let res = self.migration_apply_dev_mode().await;
+
+            if let Err(e) = &res {
+                print::error!("{e}");
+                // TODO
+                // matcher.should_retry = true;
```
msullivan, line 50: ?

msullivan (approved):
Seems reasonable, though losing the determinism is a bit of a downside

src/watch/migrate.rs
aljazerzen: This is for when `migration apply` fails due to network / non-schema-related reasons. We deemed this retry feature as unimportant so it is left unimplemented for now.

PR #1478, opened 2025-02-17
Remove duplicated `branching` command group

aljazerzen:
For the repl backslash commands, we had a separate `Branching` cmd struct. This was implemented via being in the `Common` cmd (which contains all "common" commands that can be executed in repl backslash or as normal CLI command).

But this meant that we also had CLI command named `branching`:
```
# gel branching current

# gel branch current
```

???

So I've removed `Branching` struct in favor of `Common::Branch`, renamed `\branching` to `\branch` and cleaned-up a lot of code duplication and generalization.


mmastrac (approved):
LGTM!

PR #1479, opened 2025-02-17
Add tests for projects with both gel.toml and edgedb.toml

aljazerzen:
- **project info schema-dir**
- **add tests for projects with both gel.toml and edgedb.toml**


aljazerzen:
Actually, we already have the test for that in edgedb-rust: https://github.com/edgedb/edgedb-rust/blob/master/gel-tokio/src/builder.rs#L2126-L2130

I'll keep just the `project info --get schema-dir` change.

mmastrac (approved):

PR #1481, opened 2025-02-18
Update [[watch]] in manifest to be an array

aljazerzen:
With this PR, I believe the watch & hooks are now in sync with [RFC 1028](https://github.com/edgedb/rfcs/blob/c9eea6c7be7d04d3acfb9319c7c962dadd69c20b/text/1028-cli-hooks.rst)

### DIFF
--- a/src/portable/project/init.rs
+++ b/src/portable/project/init.rs
@@ -632,7 +632,7 @@ fn init_new(
             },
             project: None,
             hooks: None,
-            watch: None,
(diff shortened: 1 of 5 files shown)


msullivan (approved):
Are there not any tests affected by this change?
There ought to be, right?

aljazerzen:
They ought to be, if we had any. Which is something I should change.

PR #1488, opened 2025-02-20
Update package name and authors to new name

msullivan:
We should probably also change the default binary, but I don't want to
have to chase down and fix the copies/symlinks that packaging/server
do right now.


mmastrac (approved):

aljazerzen:
> but I don't want to
have to chase down and fix the copies/symlinks that packaging/server
do right now.

... and all the tests that refer to "edgedb" binary in PATH.

PR #1496, opened 2025-02-24
Enable hook tests for windows

aljazerzen:
Ok, as I manually verified, hooks and watch mostly work on windows via WSL.

The problem we still face is that when we invoke a command in WSL, it will run within WLS and think it is on Linux. This means it will look at `/root/.config/edgedb` for config instead of `/mnt/c/Users/username/AppData/Roaming/edgedb`.

Which means that `gel instance list` it will return different list of instances within WSL and outside. That breaks the tests we have for hooks (well, a similar thing with project stash).

---

We now need to decide between two options:
- don't run hooks in WSL, but within windows instead. We should think how to execute scripts in this case.
- patch the CLI (and bindings) to understand that they are running on a linux within WSL and use the config from Windows FS instead.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -194,7 +194,7 @@ jobs:
             distribution: Debian
 
       - run: |
-          cargo test --workspace --test=${{ matrix.test }} --features portable_tests
+          cargo test --workspace --test=${{ matrix.test }} --features portable_tests -- --test-threads=1
 
   test-bin-installable:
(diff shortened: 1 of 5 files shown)


aljazerzen:
As discussed with @msullivan, we opted for not involving WSL at all and running scripts with `cmd.exe` instead. See https://github.com/geldata/rfcs/pull/108 for full justification.

aljazerzen:
Stack overflow? What?

PR #1508, opened 2025-02-26
Fix CLI install in WSL for aarch64

aljazerzen:
Fixes installation of the CLI in WSL on aarch64. I don't know how it worked before.


scotttrinh (approved):
I think I know the answer here, but do we have tests that could be added to catch this?

aljazerzen:
Testnig this should be as easy as it is adding aarch64-win32-unknown to our test matrix?

PR #1509, opened 2025-02-27
Use new link to scoping docs

scotttrinh:
We will also stick a redirect in, but this should be the correct link going forward.


src/portable/project/mod.rs
```diff
@@ -82,7 +82,7 @@ const FUTURES_SCHEMA: &str = "\
 const SIMPLE_SCOPING_SCHEMA: &str = "\
     # Use a simpler algorithm for resolving the scope of object names.\n\
     # This behavior will become the default in Gel 7.0.\n\
-    # See: https://docs.edgedb.com/database/edgeql/path_resolution#new-path-scoping\n\
+    # See: https://docs.edgedb.com/reference/edgeql/path_resolution#new-path-scoping\n\
```
elprans, line 85: Not `docs.geldata.com`?

scotttrinh: oh, good catch 🤦

PR #1513, opened 2025-02-28
Install completions for both edgedb and gel

### LINKED ISSUE #1511: Broken gel bash completion script
- EdgeDB CLI Version: 7.0.3+83ce17b and 7.1.0-dev.1291+ff53f4d
- OS Version: 22.04.1-Ubuntu

The bash completion script for the `gel` command is broken. It looks like it is just a copy of the `edgedb` command's script with out changing the expected command name. Nothing happens when I push tab.

aljazerzen:
Closes #1511

### DIFF
--- a/src/cli/gen_completions.rs
+++ b/src/cli/gen_completions.rs
@@ -57,15 +57,6 @@ fn write_completion(path: &Path, shell: Shell) -> anyhow::Result<()> {
 pub fn write_completions_home() -> anyhow::Result<()> {
     let home = home_dir()?;
     write_completion(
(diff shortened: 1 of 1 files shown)


scotttrinh (approved):
🆒

fmoor (commented):
Does the generated completion script contain the right executable name? The completion script contains the executable name and won't work if there is a miss match. It isn't clear to me from the diff that the `gel` completion wont be an exact copy of the `edgedb` script.

aljazerzen:
Right - writing to edgedb is pointless, because completions will always contain `gel` from now on. Good spot.

fmoor (approved):

PR #1514, opened 2025-02-28
Rename edgedb binary to gel

aljazerzen:
This will require tweaks in the build process and dev install in the main gel repo


msullivan:
I think I already did this on Monday: https://github.com/geldata/gel-cli/pull/1500

PR #1522, opened 2025-03-04
Usually eliminate ANSI spew on slow ssh connections

### LINKED ISSUE #1520: Terminal background detection sometimes spews ANSI control sequences on slowish ssh connections
```
$ gel instance list
^[]11;rgb:0000/0000/0000^G┌───────┬──────┬─────────────────┬─────────────┬─────────┐
│ Kind  │ Name │ Location        │ Version     │ Status  │
├───────┼──────┼─────────────────┼─────────────┼─────────┤
│ local │ test │ localhost:10700 │ 6.1+4db24c0 │ running │
└───────┴──────┴─────────────────┴─────────────┴─────────┘
sully@moroso:~
$ 11;rgb:0000/0000/0000
```

The problem is that the library we use (https://docs.rs/terminal-light/latest/terminal_light/) to detect background color does it by emitting an ANSI control sequence asking for the terminal to write it back, but it does so with a timeout, in case the terminal doesn't support it. If the timeout expires but the terminal *does* support it, we'll get spew like this.

`terminal_light` apparently uses a 20ms timeout, per its docs, which is just not that much if you are running the command over ssh.

Possible solutions
* Use a substantially higher timeout. https://crates.io/crates/termbg is another library doing the same thing that exposes the timeout as a parameter.
* Configure the theme in the config file instead?
* Do both of those: try to detect the color on first startup and store it in the config file

Thoughts?

msullivan:
At @mmastrac's suggestion, I made the library that queries the
terminal use a "fence" technique where we send a universally-supported
query after the background color query, so that we can detect when the
BG color query isn't handled.

This lets us turn up the timeout quite high (we *maybe* could
eliminate it, but that makes me a little nervous).

I tested on a slow SSH connection using gnome-terminal (which does
support querying BG color) and pterm (which does not).

See:
 * https://github.com/msullivan/xterm-query/pull/1
 * https://github.com/msullivan/terminal-light/pull/1
 * https://github.com/msullivan/terminal-light/pull/2

Fixes #1520.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4232,9 +4232,8 @@ dependencies = [
 
 [[package]]
 name = "terminal-light"
-version = "1.7.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4a9474484d1a0c031cd7d065c6f027a376859c9fedb32c94df3d7a797218bbb7"
+version = "1.6.0"
+source = "git+https://github.com/msullivan/terminal-light#d077ce4ff1906c5c47b2b438a043a1868bd8c5d4"
 dependencies = [
  "coolor",
  "crossterm",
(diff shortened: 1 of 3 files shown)


msullivan:
My goal is to get these up streamed, though. I've submitted one so far

aljazerzen (approved):

PR #1527, opened 2025-03-07
Update CLI to use new gel-dsn lib

mmastrac:
The `gel-dsn` library replaces the previous connection argument parsing. There are some minor API changes that we need to update here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -199,11 +199,3 @@ vec_init_then_push = 'allow'
 while_let_on_iterator = 'allow'
 too_many_arguments = 'allow'
 clone_on_copy = 'allow'
-
-[patch.crates-io]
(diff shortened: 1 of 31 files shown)


Cargo.toml
```diff
@@ -198,3 +199,12 @@ vec_init_then_push = 'allow'
 while_let_on_iterator = 'allow'
 too_many_arguments = 'allow'
 clone_on_copy = 'allow'
+
+[patch.crates-io]
```
elprans, line 203: Probably wasn't intended to get committed?

mmastrac: Need this locally until I can land the gel-rust changes :)

elprans (approved):

PR #1528, opened 2025-03-10
Minor testing improvements

mmastrac:
Extracted from #1527

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,7 +33,7 @@ jobs:
         with:
           submodules: true
       - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         with:
           submodules: true
       - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@d0e72ca3bfdc51937a4f81431ccbed269ef9f2a2
```
elprans, line 36: There are more uses of `actions-rs/toolchain` in this workflow. Would it make sense to switch those also?

elprans (approved):

PR #1535, opened 2025-03-11
Remove the edgedb-init scripts

msullivan:
They live in our website repo now.
(See https://github.com/geldata/website/blob/main/site/public/_install_cli.sh)


mmastrac (approved):

msullivan:
We should try to make this the source of truth automatically somehow, probably

PR #1536, opened 2025-03-12
Support installing musl server builds

### LINKED ISSUE #838: Add installation support for musl builds
We now have server builds that target musl libc, so the CLI should learn to recognize that it's running on a system with such libc and choose the correct artifact.

msullivan:
We build musl servers so we might as well actually support them.
I don't know if anyone will ever use this.

Tested on an alpine linux VM.

Fixes #838.

### DIFF
--- a/src/portable/platform.rs
+++ b/src/portable/platform.rs
@@ -1,4 +1,5 @@
 use std::fs;
+use std::process;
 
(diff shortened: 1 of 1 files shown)


mmastrac (approved):
Approved but is there a way to test this?

PR #1542, opened 2025-03-12
Use term=dumb for interactive tests

mmastrac:
These interactive tests fake a tty, but we really don't care about ANSI and it makes our lives so much harder.


mmastrac:
rustyline doesn't like TERM=dumb -- it fails to write the initial strings

PR #1548, opened 2025-03-17
Tweak writing of credentials in instance create

aljazerzen:
The recent migration to gel-dns introduced a "new" credentials.json format, which differs from the old in that it contains "applied defaults" instead of missing fields which imply defaults.

This PR makes sure that `gel instance create` writes `credentials.json` with the new format so user is not prompted to update the format right after creation.


mmastrac:
Good catch -- I think that was inadvertent but it's probably better to have the explicitly resolved credentials in here. It was just branch/database?

I'll see if I can add some tests for this.

mmastrac (approved):

aljazerzen:
It was missing `host`, `branch` and `tls_security`.

PR #1554, opened 2025-03-20
Rename edgedb-cli-derive to gel-cli-derive

mmastrac:
Pure rename + swapping out unmaintained crate (proc_macro_error -> proc_macro_error2).

`cargo fmt` adds a bit of noise.


scotttrinh (approved):
🪼

PR #1563, opened 2025-03-25
Fix credentials when upgrading beyond v5

mmastrac:
If the credentials file was pointed at database = edgedb previously, it needs to be updated to branch = main after the upgrade.

### DIFF
--- a/src/portable/instance/upgrade.rs
+++ b/src/portable/instance/upgrade.rs
@@ -10,7 +10,6 @@ use gel_cli_derive::IntoArgs;
 use gel_cli_instance::cloud::CloudInstanceUpgrade;
 
 use crate::branding::{BRANDING, BRANDING_CLI_CMD, BRANDING_CLOUD, QUERY_TAG};
-use crate::cloud;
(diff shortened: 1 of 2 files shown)


src/portable/instance/upgrade.rs
```diff
@@ -407,6 +407,15 @@ pub fn upgrade_incompatible(
                 paths.dump_path.join("edgedb.dump"),
                 paths.dump_path.join("main.dump"),
             )?;
+
+            if let Ok(mut creds) = credentials::read_sync(&paths.credentials) {
+                if creds.database == Some("edgedb".to_string()) {
+                    log::info!("Updating credentials file {:?}", &paths.credentials);
+                    creds.database = Some("main".to_string());
```
scotttrinh, line 414: Any drawback to unsetting the `database` property here?

mmastrac: We currently have a bug where the CLI wants to rewrite the credentials if we omit it. Ideally we'd only have one or the other, yeah.

scotttrinh (approved):

PR #1564, opened 2025-03-25
Don't spew stdin for failed docker builds

mmastrac:
When the docker builds failed, they spewed 100k lines to stdout.


mmastrac:
Requires a bit more work to fix.

PR #1572, opened 2025-04-01
Start migration to testing scripts

mmastrac:
This is a WIP while I work through options to figure out what will be most robust solution.

These scripts are written using a language of a tool that isn't published yet.

The goal with the scripts is to gain confidence that our built binaries:

 - Work functionally, as we expect, on all supported tier-1 systems (linux/macos/windows)
 - Behave as expected in complex or detailed situations, on our main supported platform Linux.
 - Are compatible with all operating systems we build binaries for.
 - Work as expected with all versions of the server that we publish: current versions, nightly and past versions up to a support window.

This means that not all functionality needs to be tested in each situation, however, we must ensure that sufficient coverage for each case exists.


tests/scripts/link/script.cli
```diff
@@ -0,0 +1,234 @@
...
+? ┌%{DATA}┬%{DATA}┬%{DATA}┐
+? │%{DATA}│%{DATA}│%{DATA}│
+repeat {
+    choice {
+        ? │ stable %{DATA}
+        ? │ nightly %{DATA}
+        ? │ testing %{DATA}
+    }
+}
+? └%{DATA}┴%{DATA}┴%{DATA}┘
```
aljazerzen, line 42: Woah, this is very cool.

I'll have to learn how to read grok patterns. Where is `ALLOW_FAILURE` defined?
I see that there is a definition for VERSION in the comment above, but there isn't anything for DATA or GREEDYDATA.

tests/scripts/link/script.cli
```diff
@@ -0,0 +1,234 @@
...
+        ? Created symlink %{GREEDYDATA}
+    }
+}
+! Instance inst1 is up and running.
+! To connect to the instance run:
+!   gel -I inst1
+
+$ gel project info --instance-name --project-dir=tests/proj/project1
+%EXIT 1
+! gel error: Project is not initialized. Run `gel project init`.
```
aljazerzen, line 66: What would you think about also placing contents of all needed files into these scripts?
Like have a header that says:
```
file "tests/proj/project1/gel.toml" {
[project]
schema-dir = "dbschema"
}
``` 
That would make these files more isolated and portable. We would have to make a mechanism that creates a tmp dir, writes the files and removes them after the test (or mock the file system directly).

tests/scripts/link/script.cli
mmastrac: This is a decent list of Grok patterns: https://docs.mezmo.com/telemetry-pipelines/using-grok-to-parse

For some reason these aren't well-documented outside a few pages on the web.

I need to update this script a bit but where you see "%ALLOW_FAILURE" that's basically "%EXIT any" (ie: any exit code)

tests/scripts/link/script.cli
mmastrac: Let me double-check, but heredocs might work for this...

aljazerzen (approved):

mmastrac:
The goal with the scripts is to gain confidence that our built binaries:

    Work functionally, as we expect, on all supported tier-1 systems (linux/macos/windows)
    Behave as expected in complex or detailed situations, on our main supported platform Linux.
    Are compatible with all operating systems we build binaries for.
    Work as expected with all versions of the server that we publish: current versions, nightly and past versions up to a support window.

.github/workflows/new-tests.yml
```diff
@@ -0,0 +1,164 @@
+name: New Tests
+
+on:
+  push:
+    branches: [ main ]
```
fantix, line 5:
```suggestion
    branches: [ 'master' ]
```

PR #1574, opened 2025-04-02
Support branching commands on cloud without a project

aljazerzen:
Branching commands all need to know the "current branch".
Current branch is either inferred from the project or
read from credentials.json.

This is a problem for cloud instances, becuase they do not
have a credentials.json so before this PR, we only supported
branching commands on cloud if there was an associated project.

This PR makes branching commands use --branch arg to override
current branch, which then removes the need for instances to have
credentials.json.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -24,9 +24,6 @@
     "edgedb": {
       "inputs": {
         "crane": "crane",
-        "fenix": [
-          "fenix"
-        ],
         "flake-parts": [
           "flake-parts"
         ],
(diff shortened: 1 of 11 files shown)


aljazerzen:
Actually, no. This is a half-assed patch of the symptom of the real problem, which is that cloud instances do not have a "current branch" concept.

We should fix that instead.

elprans:
> Actually, no. This is a half-assed patch of the symptom of the real problem, which is that cloud instances do not have a "current branch" concept.
> 
> We should fix that instead.

`gel branch wipe` currently takes a non-optional `<branch>` argument.  We should fix that at least before doing a more general Cloud "current branch" fix.

elprans:
Proposed alternative fix in https://github.com/geldata/gel-cli/pull/1575

aljazerzen:
So first, adding a "current branch" to cloud instances is more complicated than just a fix in the cli - we would have to update all bindings.

Your fix is better, because it just uses the branch from the wipe command.

But this is a problem also applies to `list`, `drop`, `rename`, since they do not really require the "current branch". Let me try another fix.

PR #1581, opened 2025-04-07
Improve credential load/save handing

mmastrac:
The CLI was failing to load certain outdated credentials files, mostly because we have so many different code paths that load credentials. This moves all of them to the new stored-info handlers in gel-dsn.

Major changes:

 - Credentials are now auto-upgraded in place when necessary. We no longer prompt for upgrades, though in the future we may wish to do opt-in major versioned upgrades.
 - `DatabaseBranch` is used in as many places as we can to avoid ambiguity.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1511,7 +1511,6 @@ dependencies = [
  "gel-cli-derive",
  "gel-cli-instance",
  "gel-derive",
- "gel-dsn",
  "gel-errors",
  "gel-jwt",
  "gel-protocol",
@@ -1639,9 +1638,15 @@ dependencies = [
 
 [[package]]
 name = "gel-dsn"
+<<<<<<< HEAD
 version = "0.2.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "299df7202234e7e15016fa6bcc9c3647dd14f58713472dd88fcaf44ddc80b30f"
+=======
+version = "0.2.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8c1104cd7b44da2116b86eadcbacf140a898635d46fcd7b5e0d5243c658087e3"
+>>>>>>> 842175e (.)
 dependencies = [
  "base64 0.22.1",
  "crc16",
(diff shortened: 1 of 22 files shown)


src/cli/install.rs
```diff
@@ -198,27 +198,20 @@ fn _run(cmd: &Command) -> anyhow::Result<()> {
...
     }
 
+    // We no longer support migration from the old layout.
     let base = home_dir()?.join(".edgedb");
     let new_layout = if base.exists() {
         eprintln!(
             "\
                 {BRANDING_CLI} no longer uses '{}' to store data \
                 and now uses standard locations of your OS. \
+                If you have migrated, you can safely ignore this message \
```
scotttrinh, line 208: I think it's not completely obvious that "migrated" here means "have run the `gel cli migrate` command in the past". Maybe "If you have used the CLI to update to the newer directory layout in the past..." or something like that?

src/cli/install.rs
```diff
@@ -198,27 +198,20 @@ fn _run(cmd: &Command) -> anyhow::Result<()> {
...
+                and remove the directory. If you have not migrated, you can \
+                download and run an older version of `edgedb-cli` to migrate. \
```
scotttrinh, line 210: I think it's pretty difficult for someone to "run an older version of `edgedb-cli`" without asking us directly how to do this. Not really sure what to say about this other than maybe we keep the old code around forever or consider this a breaking change and don't point someone at the impossible task of figuring out how to download an older version.

scotttrinh (commented):
I didn't do a very thorough code review, but a few comments about messaging. The other thing I want to mention is that I'm a little 😰 about automatically mutating credentials files. On the one hand, it's _kind of_ an implementation detail that _usually_ is something we generate and update, but it's also a public interface. Maybe at least still flagging to the user that the file has been updated seems worth doing?

src/cli/install.rs
mmastrac: I could potentially just remove this check as I think it's pretty unlikely anyone will end up in this situation.

src/cli/install.rs
mmastrac: I could potentially just remove this check as I think it's pretty unlikely anyone will end up in this situation.

scotttrinh: Yeah, I'm down with that.

PR #1588, opened 2025-04-10
Skip the update check if the binary was just written

mmastrac:
This potentially helps with getting flagged by Windows anti-virus. Instead of doing an update check after we freshly install the CLI, we check the CLI modification/creation dates and wait until the CLI binary is sufficiently old enough (1 day) before checking.

We don't skip the check if either date is in the future. 

The untested theory is that we are doing too much on first start. The heuristic/AI-based malware detectors use behaviour-on-first-boot as a way to categorize things as sketchy or not, but do not take into account the reputation of the domains we are communicating with.

This removes a network check on most first boots, which hopefully makes it look less suspicious.

### DIFF
--- a/src/version_check.rs
+++ b/src/version_check.rs
@@ -71,9 +71,44 @@ fn newer_warning(ver: &ver::Semver) {
     }
 }
 
+/// Check the modification and creation time of the CLI binary. If either of
+/// them is younger than 1 day, skip the version check. This prevents the
+/// version check from being run on the first download which _may_ trigger
+/// Windows anti-virus.
+fn _check_newly_installed() -> anyhow::Result<bool> {
+    let metadata = fs::metadata(std::env::current_exe()?)?;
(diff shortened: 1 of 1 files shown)


src/version_check.rs
```diff
@@ -71,9 +71,44 @@ fn newer_warning(ver: &ver::Semver) {
...
+    Ok(false)
+}
+
 fn _check(cache_dir: &Path, strict: bool) -> anyhow::Result<()> {
     let self_version = cli::upgrade::self_version()?;
     let channel = cli::upgrade::channel();
+
+    if _check_newly_installed().unwrap_or(true) {
+        return Ok(());
+    }
```
aljazerzen, line 110: This should probably go into `check` function, before we start looking for cache dirs and self versions.

https://github.com/geldata/gel-cli/blob/master/src/version_check.rs#L149

aljazerzen (approved):

aljazerzen:
... but I don't have an insight into how much this would help the antivirus problem.

PR #1592, opened 2025-04-14
Local instance restore 

mmastrac:
Adds `instance restore` to local instances. For the initial implementation of this feature, the server must be stopped.

We run `pg_verifybackup` and unpack the backup data and WAL files to a new server folder.

Note that we need to ship `pg_waldump` with the server binaries for this to work.

### DIFF
--- a/gel-cli-instance/src/local/localbackup.rs
+++ b/gel-cli-instance/src/local/localbackup.rs
@@ -5,6 +5,7 @@ use std::{
 use anyhow::bail;
 use futures::FutureExt;
 use gel_dsn::gel::InstanceName;
+use log::info;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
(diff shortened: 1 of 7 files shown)


gel-cli-instance/src/local/localbackup.rs
```diff
@@ -98,6 +127,14 @@ impl InstanceBackup for LocalBackup {
                 _ = std::fs::remove_dir_all(&temp_dir);
             });
 
+            // TODO: we should trigger the server to start if it's not running
```
elprans, line 130: Yeah, we should. There's no reason not to given that the CLI has complete control over the instance.

gel-cli-instance/src/local/localbackup.rs
```diff
@@ -116,6 +154,126 @@ impl InstanceBackup for LocalBackup {
...
+                    Ok::<_, std::io::Error>(())
+                }).await??;
+            }
+            if restore_tmpdir2.exists() {
+                bail!("Restore directory {} already exists from an incomplete restore, please remove it manually.", restore_tmpdir2.display());
+            }
+
+            if run_dir.join(".s.PGSQL.5432").exists() {
+                return Err(anyhow::anyhow!(
+                    "PostgreSQL socket found at {}. Stop the server before restoring.",
```
elprans, line 191: We should do this automatically.

gel-cli-instance/src/local/localbackup.rs
mmastrac: I want to do this in a follow-up -- coordination of server start/stop adds a bit of complexity and if I'm going to clean up the currently spread-out code that might take some time.

gel-cli-instance/src/local/localbackup.rs
mmastrac: Ditto, for a follow-up.

elprans (approved):

PR #1594, opened 2025-04-15
Add assert for system/instance paths

mmastrac:
This is a sanity check for the new path code in gel-dsn -- this will allow us to ensure we've correctly migrated to that code.


src/portable/local.rs
```diff
@@ -12,6 +12,7 @@ use fn_error_context::context;
 
 use gel_tokio::dsn::UnixPath;
 use gel_tokio::{Builder, Config, InstanceName};
+use libc::system;
```
aljazerzen, line 16: stray import?

aljazerzen (approved):

src/portable/local.rs
```diff
@@ -12,6 +12,7 @@ use fn_error_context::context;
...
```
mmastrac, line 15:
```suggestion
```

src/portable/local.rs
```diff
@@ -285,7 +286,25 @@ impl Paths {
...
+        assert_eq!(system_paths.cache_dir, cache_dir().ok());
+        assert_eq!(
+            system_paths.data_dir.unwrap().join("data"),
+            data_dir().unwrap()
+        );
+        assert_eq!(system_paths.config_dir, config_dir().ok());
+        assert_eq!(system_paths.home_dir, crate::platform::home_dir().ok());
+
+        assert_eq!(instance_paths.data_dir, res.data_dir);
+        assert_eq!(instance_paths.credentials_path, credentials::path(name)?);
```
mmastrac, line 304:
```suggestion
```

PR #1595, opened 2025-04-15
Bump ubuntu 20 -> 22

mmastrac:
Ubuntu 20 runners are kaput, we need to move to at least 22.

https://github.com/actions/runner-images/issues/11101


nutgood (changes requested):
Might as well do `24.04` while we're here, no ?

nutgood (approved):

PR #1599, opened 2025-04-22
Apply clippy::large_futures lint

aljazerzen:
`await` macro desugars into a state machine that holds all local
variables at each `.await` point a function. This state machine is
stored on the stack, which might get very large.

On Windows we were hitting stack overflow problems in latest attempts
to test hooks and watch command.

Solution is to wrap futures with `Box::pin`, which moves the state
machine to the heap.

Fortunately, there is a clippy lint that finds large futures.

This PR also applies a few other lints in Windows-only code.


src/migrations/upgrade_check.rs
```diff
@@ -197,7 +197,7 @@ async fn do_check(
         .with_fs()
         .build()
         .context("cannot build connection params")?;
-    let cli = &mut Connection::connect(&config, QUERY_TAG).await?;
+    let cli = &mut Box::pin(Connection::connect(&config, QUERY_TAG)).await?;
```
mmastrac, line 200: Rather than putting these in the caller, let's change the functions themselves to by non-async and return `impl Future + Unpin`. The changes will be less invasive.

aljazerzen: Do you mean this specific call or all of the new `Box::pin`s?

aljazerzen: I see, you mean just `Connection::connect`, because it is used in a lot of places.

mmastrac: I think `Connection::connect` would be a good candidate. Looking through the others, I don't really see any other commonly-used and large futures.

PR #1604, opened 2025-04-30
Add --skip-hooks and GEL_SKIP_HOOKS

aljazerzen:
Implements https://github.com/geldata/rfcs/pull/112

The majority of the work here was plumbing `crate::options::Options`
from top-level functions into commands that might trigger hooks.

I've also renamed a few params in the spirit of refactoring in #1465.

There are tests for --skip-hooks and GEL_SKIP_HOOKS, but only for
`branch switch` command.


mmastrac (approved):
LGTM: the argument needing to be passed down into so many functions suggests we might need to create a migration context at a higher level, but shouldn't block this from landing.

aljazerzen:
> we might need to create a migration context at a higher level

Well, hooks are not specific to migrations. They might be defined for `project.init`, `branch.switch` and a few others. So this would not be a "migration context" be a general "command context".

Which is what already is `crate::commands::Options`. We should be using that more instead of `crate::options::Options`, which contains *all* process args provided, not just the stripped down version.

(and we already do have a migration context, this PR does add a `skip_hooks` field to it)

PR #1610, opened 2025-05-01
Fix squashing of long migration histories

### LINKED ISSUE #1609: Migration squash results in edgedb error
- EdgeDB Version: 5.6+adb9e77
- EdgeDB CLI Version: 6.0.2+5ba8c2c
- OS Version: mac os 13.5.1 (22G90)

We have a few manually created migrations (that have all been successfully migrated) and many automatic migrations. Running  `edgedb migration create --squash` results in one fixup and reduces the amount of migrations. In the end, the process ends with  `edgedb error: No such file or directory (os error 2)`.

msullivan:
When deleting old migration scripts during squash, we first rename
them to `.old.edgeql` and then delete them.  While looping through the
directory contents, if we see an existing `.old.edgeql`, we register
that to be deleted too.

However, we are iterating through the directory as we do this, so it
is possible that we will see a file that was just *renamed* and thus
is already registered to be deleted, and which we will then try to
delete twice.

On linux, this happens if there too many files to fit in one call to
`getdents64` with whatever buffer size libc is using.  To make things
extra fun, I originally struggled to reproduce this because glibc uses
a much larger buffer size than musl does (32k vs 2k).

Fix by just dropping the deletion of existing `.edgeql.old` files. I
don't think they should be there for any reason of ours, and so we
don't want to mess with them?
I don't feel strongly, though; we could also skip duplicates.

Tested manually; I didn't include a test because the long migration
history is a pain.

Fixes #1609.

### DIFF
--- a/src/migrations/squash.rs
+++ b/src/migrations/squash.rs
@@ -265,8 +265,6 @@ impl TwoStageRemove<'_> {
             }
             if lossy_name.ends_with(".edgeql") {
                 self.rename(&item.path()).await?;
-            } else if lossy_name.ends_with(".edgeql.old") {
-                self.filenames.push(item.path());
             }
         }
         Ok(())
--- a/src/migrations/squash.rs
+++ b/src/migrations/squash.rs
@@ -292,3 +292,46 @@ impl TwoStageRemove<'_> {
         Ok(())
     }
 }
+
+#[tokio::test]
+async fn test_two_stage_remove() -> anyhow::Result<()> {
+    // Test for #1610, which fixed a bug when there were too many
+    // migration files for one call to getdents64.
+
+    // Put the temp dir under (unfortunately) the current directory.
+    // This is because /tmp is often a tmpfs, and the issue doesn't
+    // show up on tmpfs!
+    let tmp = tempfile::tempdir_in(".")?;
+
+    let ctx = Context {
+        schema_dir: tmp.path().to_path_buf(),
+        quiet: false,
+        project: None,
+    };
+
+    // Create test migration files
+    let migrations_dir = tmp.path().join("migrations");
+    fs::create_dir(&migrations_dir).await?;
+    for i in 0..2000 {
+        fs::write(migrations_dir.join(format!("{:05}-mXXXXXX.edgeql", i)), "").await?;
+    }
+
+    // Run rename + removal
+    let mut remover = TwoStageRemove::new(&ctx);
+    remover.rename_revisions().await?;
+    remover.commit().await?;
+
+    // Verify files are removed
+    let mut entries = fs::read_dir(&migrations_dir).await?;
+    while let Some(item) = entries.next_entry().await? {
+        let file_name = item.file_name();
+        let name = file_name.to_string_lossy();
+        assert!(
+            !name.ends_with(".edgeql") && !name.ends_with(".edgeql.old"),
+            "Found unexpected file: {}",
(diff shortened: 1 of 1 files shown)


fantix (approved):
LGTM

LLM-generated test here, but it doesn't fail without the fix somehow, even with `--target x86_64-unknown-linux-musl`:

```rust
#[test]
fn test_two_stage_remove() -> anyhow::Result<()> {
    use tokio::runtime::Runtime;
    let rt = Runtime::new()?;

    rt.block_on(async {
        let tmp = tempfile::tempdir()?;
        let ctx = Context {
            schema_dir: tmp.path().to_path_buf(),
            quiet: false,
            project: None,
        };

        // Create test migration files
        let migrations_dir = tmp.path().join("migrations");
        fs::create_dir(&migrations_dir).await?;
        for i in 0..1024 {
            fs::write(migrations_dir.join(format!("test{}.edgeql", i)), "").await?;
        }

        let mut remover = TwoStageRemove::new(&ctx);
        remover.rename_revisions().await?;

        // Commit changes and verify files are removed
        remover.commit().await?;
        let mut entries = fs::read_dir(&migrations_dir).await?;
        while let Some(item) = entries.next_entry().await? {
            let file_name = item.file_name();
            let name = file_name.to_string_lossy();
            assert!(
                !name.ends_with(".edgeql") && !name.ends_with(".edgeql.old"),
                "Found unexpected file: {}",
                name
            );
        }

        Ok(())
    })
}
```

msullivan:
> LGTM
> 
> LLM-generated test here, but it doesn't fail without the fix somehow, even with `--target x86_64-unknown-linux-musl`:
> 
> ```rust
> #[test]
> fn test_two_stage_remove() -> anyhow::Result<()> {
>     use tokio::runtime::Runtime;
>     let rt = Runtime::new()?;
> 
>     rt.block_on(async {
>         let tmp = tempfile::tempdir()?;
>         let ctx = Context {
>             schema_dir: tmp.path().to_path_buf(),
>             quiet: false,
>             project: None,
>         };
> 
>         // Create test migration files
>         let migrations_dir = tmp.path().join("migrations");
>         fs::create_dir(&migrations_dir).await?;
>         for i in 0..1024 {
>             fs::write(migrations_dir.join(format!("test{}.edgeql", i)), "").await?;
>         }
> 
>         let mut remover = TwoStageRemove::new(&ctx);
>         remover.rename_revisions().await?;
> 
>         // Commit changes and verify files are removed
>         remover.commit().await?;
>         let mut entries = fs::read_dir(&migrations_dir).await?;
>         while let Some(item) = entries.next_entry().await? {
>             let file_name = item.file_name();
>             let name = file_name.to_string_lossy();
>             assert!(
>                 !name.ends_with(".edgeql") && !name.ends_with(".edgeql.old"),
>                 "Found unexpected file: {}",
>                 name
>             );
>         }
> 
>         Ok(())
>     })
> }
> ```

Cute; how did you prompt it?

I probably won't add the test since it doesn't fail on the old version, though, and don't feel like debugging it

msullivan:
Ha! The reason why the test didn't fail is I think because my (and presumably your) `/tmp` is `tmpfs`, and it seems that tmpfs's directory listing behavior differs from `ext4` in ways that prevent the issue.

fantix:
> Cute; how did you prompt it?

I used your PR description, with a few back-and-forth revisions in details

PR #1611, opened 2025-05-02
Apply hooks in dev_mode

fantix:
`gel watch --migrate` and `gel migrate --dev-mode` should trigger `migration.apply.*` and `schema.update.*` hooks if the schema is changed.

Refs #1449

### DIFF
--- a/src/migrations/dev_mode.rs
+++ b/src/migrations/dev_mode.rs
@@ -10,7 +10,8 @@ use crate::async_try;
 use crate::branding::BRANDING;
 use crate::bug;
 use crate::commands::Options;
-use crate::migrations::apply::{apply_migrations, apply_migrations_inner};
+use crate::hooks;
+use crate::migrations::apply::{apply_migrations_ext, apply_migrations_inner};
 use crate::migrations::context::Context;
 use crate::migrations::create;
 use crate::migrations::create::{CurrentMigration, FutureMigration};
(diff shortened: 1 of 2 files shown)


src/migrations/dev_mode.rs
```diff
@@ -31,6 +32,12 @@ pub async fn migrate(cli: &mut Connection, ctx: &Context, bar: &ProgressBar) ->
     }
     let migrations = migration::read_all(ctx, true).await?;
     let db_migration = get_db_migration(cli).await?;
+
+    if let Some(project) = &ctx.project {
+        hooks::on_action("migration.apply.before", project).await?;
+        hooks::on_action("schema.update.before", project).await?;
+    }
```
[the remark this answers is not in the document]
fantix: Negative, I've reverted that and redone it in a different way, so that the hooks reflect what actually happened.

scotttrinh (approved):
Latest changes make sense to me, thanks for the effort here! 🎉

src/migrations/dev_mode.rs
```diff
@@ -163,24 +164,39 @@ async fn get_db_migration(cli: &mut Connection) -> anyhow::Result<Option<String>
...
-    }
+    let old_timeout = timeout::inhibit_for_transaction(cli).await?;
     async_try! {
         async {
-            _migrate_to_schema(cli, ctx).await
+            if matches!(cli.transaction_state(), NotInTransaction) {
+                execute(cli, "START TRANSACTION", None).await?;
+                async_try! {
+                    async {
+                        let apply_hooks = true;
```
aljazerzen, line 174: What's the intention of using a constant value here? Shouldn't this be pulled from `ctx`?

src/migrations/dev_mode.rs
```diff
@@ -163,24 +164,39 @@ async fn get_db_migration(cli: &mut Connection) -> anyhow::Result<Option<String>
...
+                    },
+                    except async {
+                        execute_if_connected(cli, "ROLLBACK").await
+                    },
+                    else async {
+                        execute_if_connected(cli, "COMMIT").await
+                    }
+                }
+            } else {
+                let apply_hooks = false;
```
aljazerzen, line 185: Ditto

src/migrations/dev_mode.rs
fantix: Oh, it was to give the value a name, in the lack of named parameters in Rust.

Issue #1607, opened 2025-05-01
`gel watch --migrate` should pick up _current_ changes as well


scotttrinh:
If you've made some changes outside of the watch loop, and start up the watcher, the watcher will not detect the existing changes until you `touch` the file again. This makes sense conceptually, but from the user's perspective, it should act like `gel migrate --dev-mode && gel watch --migrate`.

PR #1612, opened 2025-05-03
Make `gel watch --migrate` run migration on startup

fantix:


src/watch/migrate.rs
```diff
@@ -39,6 +39,11 @@ impl Migrator {
         mut input: UnboundedReceiver<ExecutionOrder>,
         matcher: Arc<Watcher>,
     ) {
+        // Run initial dev-mode migration on startup
+        if let Err(e) = self.migration_apply_dev_mode().await {
+            print::error!("{e}");
+        }
```
scotttrinh, line 45: question: is there a way to just poke the `input` here so that we don't have to keep this bit aligned with the side effects below?

fantix: Oh, that was my first attempt 👍 good question! It's just a bit tricky to rearrange the ownership of the sender `tx` here, so I just made another explicit call; the side effect could become a reused function if it grows larger in the future.

fantix: Sure! Now fixed to use a do..while loop.

scotttrinh (approved):
oh, this is _great_ 👏

aljazerzen (approved):
Looks good

aljazerzen:
Closes #1607

PR #1615, opened 2025-05-09
Implement incremental backup

mmastrac:
Performing a backup for an instance will now use an "automatic" mode where we perform an incremental backup if the current backup's "generation" is <= 5 (an arbitrary number to avoid incrementals becoming too deeply nested) or if the previous backup is older than a week.

We also terminate and start the instance appropriately for backup and restore to take place.

### DIFF
--- a/gel-cli-instance/src/cloud/instance.rs
+++ b/gel-cli-instance/src/cloud/instance.rs
@@ -3,11 +3,9 @@ use gel_dsn::gel::{CloudName, InstanceName};
 use std::time::Duration;
 
 use crate::instance::{
-    Instance, InstanceOpError, Operation,
     backup::{
-        Backup, BackupId, BackupStrategy, BackupType, InstanceBackup, ProgressCallback, RestoreType,
-    },
-    map_join_error,
+        Backup, BackupId, BackupStrategy, BackupType, InstanceBackup, ProgressCallback, RequestedBackupStrategy, RestoreType
+    }, map_join_error, Instance, InstanceOpError, Operation
 };
 
 use super::{CloudApi, CloudError, CloudHttp, schema};
@@ -23,11 +21,17 @@ struct CloudInstanceBackup<H: CloudHttp> {
 }
 
 impl<H: CloudHttp> InstanceBackup for CloudInstanceBackup<H> {
-    fn backup(&self, callback: ProgressCallback) -> Operation<Option<BackupId>> {
+    fn backup(&self, strategy: RequestedBackupStrategy, callback: ProgressCallback) -> Operation<Option<BackupId>> {
         let api = self.instance.api.clone();
         let name = self.instance.name.clone();
 
         tokio::spawn(async move {
+            if strategy != RequestedBackupStrategy::Auto {
+                return Err(CloudError::InvalidRequest(
+                    "cloud backups only support automatic backups".to_string(),
+                ));
+            }
+
             let operation = api.create_backup(&name).await?;
             api.wait_for_operation(operation, Duration::from_secs(30 * 60), callback)
                 .await?;
--- a/gel-cli-instance/src/cloud/instance.rs
+++ b/gel-cli-instance/src/cloud/instance.rs
@@ -3,9 +3,12 @@ use gel_dsn::gel::{CloudName, InstanceName};
 use std::time::Duration;
 
 use crate::instance::{
+    Instance, InstanceOpError, Operation,
     backup::{
-        Backup, BackupId, BackupStrategy, BackupType, InstanceBackup, ProgressCallback, RequestedBackupStrategy, RestoreType
-    }, map_join_error, Instance, InstanceOpError, Operation
+        Backup, BackupId, BackupStrategy, BackupType, InstanceBackup, ProgressCallback,
+        RequestedBackupStrategy, RestoreType,
+    },
+    map_join_error,
 };
 
 use super::{CloudApi, CloudError, CloudHttp, schema};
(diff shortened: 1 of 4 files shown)


src/portable/instance/backup.rs
```diff
@@ -203,6 +238,11 @@ pub async fn restore(cmd: &Restore, opts: &crate::options::Options) -> anyhow::R
         return Ok(());
     }
 
+    if let InstanceName::Local(inst_name) = &cmd.instance {
+        let inst_name = inst_name.clone();
+        tokio::task::spawn_blocking(move || super::control::do_stop(&inst_name)).await??;
```
1st1, line 243: We want to notify the user that the instance will be stopped (or maybe even confirm that?)

mmastrac: We currently back up and restore the whole DB, but I'll make sure this is worded properly.

msullivan:
Is there any dev documentation that should be written about the overall scheme?

src/portable/instance/backup.rs
mmastrac:
```rust

    let stop_warning = if let InstanceName::Local(_) = &cmd.instance {
        "This will stop the instance and restore all branches from the backup. Any data not backed up will be lost."
    } else {
        "This will restore all branches from the backup. Any data not backed up will be lost."
    };

    let prompt = format!(
        "Will restore {inst_name:#} from the specified backup. {stop_warning}\
        \n\nContinue?",
    );

```

mmastrac:
> Is there any dev documentation that should be written about the overall scheme?

Unsure: it's all self-contained in localbackup.rs. The backup scheme is stock pg_basebackup (including incremental) so everything is effectively just a helper for doing standard basebackup. Restore is literally just unpacking a basebackup.

src/portable/instance/backup.rs
1st1: > "This will stop the instance and restore all branches from the backup. Any data not backed up will be lost."

I'd also add roughly this: "... After the restore operation is completed, the instance will be restarted."

BTW, can restore theoretically error out? Should we spell that out or the probability is so low that we don't want to?

1st1:
> Is there any dev documentation that should be written about the overall scheme?

Yeah, I'd appreciate documenting it "reference" style in https://docs.geldata.com/reference/using/cli -- the source for that is in `gel/docs`

msullivan:
Also, did you do any testing with crashes in the middle of backup / before the json records are written?

src/portable/instance/backup.rs
mmastrac: The restore _could_ fail but we perform the entire restore in a tempdir and then do two renames to get it into the right spot. Basically we only do the final swap of old and new dir once pg_verifybackup confirms it's valid.

mmastrac:
> Also, did you do any testing with crashes in the middle of backup / before the json records are written?

Yeah, we use temp folders for backups until they've completed, and placeholder JSON records are written without a "completed_at" field so we can detect failed backups. Silently-crashed backups are detected by `last_updated_at` becoming stale (it's SystemTime-based which means it's immune to TZ/DST changes, but we set the stale time at 1 hr to avoid getting bit by clock drift).

gel-cli-instance/src/cloud/instance.rs
```diff
@@ -23,11 +24,21 @@ struct CloudInstanceBackup<H: CloudHttp> {
...
+    fn backup(
+        &self,
+        strategy: RequestedBackupStrategy,
+        callback: ProgressCallback,
+    ) -> Operation<Option<BackupId>> {
         let api = self.instance.api.clone();
         let name = self.instance.name.clone();
 
         tokio::spawn(async move {
+            if strategy != RequestedBackupStrategy::Auto {
```
elprans, line 36: Hm, manual cloud backups should be supported, is that not the case?

mmastrac: It should be supported -- in this case the strategy is full/incremental/auto, and I figured cloud was "auto" on. BackupType is where we get manual/automatic.

Auto for strategy just means "pick incremental if we have a full backup and it's not too far away".

Hard to name the enums, but type was already in place for manual/auto :)

mmastrac:
Note that I had to do a last-minute disabling of the backup/restore on WSL because it requires some additional work to run the backups inside of the WSL environment.

fantix (approved):
I've read through the code, LGTM! Cool test btw

msullivan (approved):

gel-cli-instance/src/local/localbackup.rs
```diff
@@ -367,30 +702,41 @@ impl<P: ProcessRunner> PgBackupCommands<P> {
...
-    // }
+    pub async fn pg_combinebackup(
+        &self,
+        target_dir: impl AsRef<Path>,
+        paths: &[impl AsRef<Path>],
+        callback: ProgressCallback,
+    ) -> Result<(), anyhow::Error> {
+        let pg_combinebackup = self.find_executable("pg_combinebackup")?;
+        let mut cmd = Command::new(pg_combinebackup);
+        cmd.arg("--output").arg(target_dir.as_ref());
```
elprans, line 713: Consider using `--clone` (works on MacOS and XFS on Linux).

elprans (approved):
This is very solid stuff!

gel-cli-instance/src/local/localbackup.rs
mmastrac: I originally tried adding per-platform flags here but `pg_combinebackup` doesn't do fallback and error recovery was challenging. We might need to do some sort of capability sniffing. I'm not totally sure why they don't do that internally in the tool.

PR #1616, opened 2025-05-12
Initial locking implementation

mmastrac:
We add a set of locks on instances and projects that control exclusive or shared access. This allows us to prevent failures when `gel watch` is running.

Most commands take an exclusive lock, but a few use a shared lock which allows them to run under gel project hooks. I might have missed some commands but it's easy enough to mark them as taking read-locks.

To simplify the implementation, we add locks into higher-level contexts in the CLI rather than individual commands where possible.

### DIFF
--- a/src/locking.rs
+++ b/src/locking.rs
@@ -8,6 +8,7 @@ use gel_tokio::InstanceName;
 use log::{debug, warn};
 use serde_json::json;
 
+use crate::cli::env::Env;
 use crate::portable::project::get_stash_path;
 
 const LOCK_FILE_NAME: &str = "gel-cli.lock";
@@ -69,6 +70,14 @@ pub enum LockError {
 }
 
 fn try_create_lock(lock_type: file_guard::Lock, path: PathBuf) -> Result<LockInner, LockError> {
+    // Special case: we allow "free" shared locks in hooks, since the write lock
+    // is presumed to exist in the parent process.
+    if lock_type == file_guard::Lock::Shared
+        && *Env::in_hook().unwrap_or_default().unwrap_or_default()
+    {
+        return Ok(LockInner::NoLock);
+    }
+
     let lock_file = OpenOptions::new()
         .create(true)
         .write(true)
--- a/src/locking.rs
+++ b/src/locking.rs
@@ -147,9 +147,6 @@ impl LoopState {
     fn error(&mut self, _e: LockError) -> Result<(), LockError> {
         if self.first {
             self.first = false;
-        }
-
-        if self.first {
             if let Ok((pid, cmd)) = self.load_lock_file() {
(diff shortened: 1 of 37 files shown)


vpetrovykh:
```
$ gel watch -m
Hint: --migrate will apply any changes from your schema files to the database. When ready to commit your changes, use:
1) `gel migration create` to write those changes to a migration file,
2) `gel migrate --dev-mode` to replace all synced changes with the migration.
```
We definitely expect `gel migration create` and `gel migration apply` to work while in `watch` mode. Currently the lock prevents this.

On the other hand we still expect a migration in progress (either `create` or `apply` mode) to stop `watch` from starting so this behavior should remain as is.

vpetrovykh:
I think that failing fast without retrying is preferable at this point. This is definitely a better option for `watch` since a lock means something is wrong and we'd rather fail fast to let the user figure it out than wait a bit and make it look like things are OK. This is especially true for integration with FastAPI. The `fastapi dev` will be running and doing other stuff as if everything is fine and then potentially crashing on a completely unrelated error (like failure to connect to gel and fetch data) rather than getting the error right away, which may be confusing as to what's causing the error.

src/locking.rs
```diff
@@ -0,0 +1,415 @@
...
+    lock_type: file_guard::Lock,
+    path: &PathBuf,
+) -> std::io::Result<Option<file_guard::FileGuard<Box<File>>>> {
+    // Special case: we allow "free" shared locks in hooks, since the write lock
+    // is presumed to exist in the parent process.
+    if lock_type == file_guard::Lock::Shared
+        && *Env::in_hook().unwrap_or_default().unwrap_or_default()
+    {
+        return Ok(None);
+    }
```
aljazerzen, line 130: This only holds if the parent really is holding the lock.
That might be the case when the script is being triggered by any (currently existing) hook - I've went over the list and they are seem to acquire the lock before executing.

~But from what I read, it is not the case if the script is being run by `gel watch`~ it does call `crate::portable::project::load_ctx`, which acquires the lock.

tldr: all is well, but I had to check

aljazerzen (commented):
I haven't read all the code yet, but it looks good.

aljazerzen (approved):

aljazerzen (approved):

fantix (approved):
Should we consider the scenario where the same instance is used (as remote instances for example) by multiple projects?

mmastrac:
> Should we consider the scenario where the same instance is used (as remote instances for example) by multiple projects?

I think that's a much bigger task -- we'd need to engage the database for those locks which could be challenging. I believe the biggest thing we're preventing is mostly branch switching while watching migrations which shouldn't be an issue across machines.

PR #1619, opened 2025-05-13
Migrate hook tests to scripts

mmastrac:
Migrate the hook tests to a testing script so we can decommission an entire test matrix.


tests/scripts/project/hooks.cli
```diff
@@ -0,0 +1,108 @@
...
+! hook project.init.after: true
+*
+! hook migration.apply.before: true
+*
+! hook schema.update.before: true
+*
+! hook migration.apply.after: true
+*
+! hook schema.update.after: true
+*
```
aljazerzen, line 30: I think this is not testing that there were *no other hooks* ran in-between two hooks, right?

tests/docker_test_wrapper.rs
```diff
@@ -109,7 +109,6 @@ fn dockerfile() -> String {
 }
 
 #[test_case("portable_smoke")]
-#[test_case("portable_project")]
 #[test_case("shared_client_tests")]
```
aljazerzen, line 112: Nice :D

aljazerzen (approved):

tests/scripts/project/hooks.cli
mmastrac: True - I can reject other types of hook updates. I'll tweak the test.

PR #1621, opened 2025-05-14
Run hooks outside transactions in dev-mode

fantix:
Reproduction of the issue fixed in this PR: Adding a `schema.update.after` hook to run `gel query -L sql 'select * from pg_constraint'` would reliably deadlock `gel migrate --dev-mode` (with any schema changes), because the SQL `pg_constraint` references a materialized view `edgedbsql_VER.virtual_tables`, which is locked (`AccessExclusiveLock` and many others) in an ongoing migration transaction.

Fixed so that the `after` hooks run after migration transactions.

Refs #1611

### DIFF
--- a/src/migrations/dev_mode.rs
+++ b/src/migrations/dev_mode.rs
@@ -164,15 +164,28 @@ async fn get_db_migration(cli: &mut Connection) -> anyhow::Result<Option<String>
 async fn migrate_to_schema(cli: &mut Connection, ctx: &Context) -> anyhow::Result<bool> {
     use gel_protocol::server_message::TransactionState::NotInTransaction;
 
-    let old_timeout = timeout::inhibit_for_transaction(cli).await?;
-    async_try! {
-        async {
-            if matches!(cli.transaction_state(), NotInTransaction) {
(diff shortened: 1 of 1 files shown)


aljazerzen:
An aftereffect of this change is that if `schema.change.after` or `migration.apply.after` fail, the migration would still succeed. But this is ok and intended behavior, right @vpetrovykh ?

aljazerzen (approved):

1st1:
> An aftereffect of this change is that if schema.change.after or migration.apply.after fail, the migration would still succeed. But this is ok and intended behavior, right @vpetrovykh ?

Yes, I think this is correct. `schema.change.after` by definition means "run after the migration has been applied". So even if we had a mechanism to rollback already applied migrations it would be a big question if we should do that at all.

@fantix please go ahead with this, trigger a nightly for @scotttrinh to test, and then we have to push a release asap.

aljazerzen:
This one should probably go into release asap too:

https://github.com/geldata/gel-cli/pull/1622

fantix:
> This one should probably go into release asap too:
> 
> #1622

Roger

PR #1626, opened 2025-05-23
Don't set force_database_error in watch

fantix:
`force_database_error` is breaking dev workflows, this is the first step of its deprecation and drop.


vpetrovykh:
It seems to clear the previous error state (from the older CLI version), which is important :+1:

PR #1630, opened 2025-05-26
localdev: auto backup before migration

fantix:
- [x] print backup messages
- [x] before `gel migrate`
- [x] before `gel migrate --dev-mode`
- [ ] within `gel watch`
- [ ] add environment variable to turn auto backup off


mmastrac:
I wonder if we should omit the full path to the backup and just list the backup ID, with a command to restore it?

gel-cli-instance/src/instance/backup.rs
```diff
@@ -106,4 +114,6 @@ pub trait InstanceBackup {
     ) -> Operation<()>;
     /// List backups.
     fn list_backups(&self) -> Operation<Vec<Backup>>;
+    /// Get backup details by ID.
+    fn get_backup(&self, backup_id: &BackupId) -> anyhow::Result<Backup>;
```
mmastrac, line 118: This should be an Operation as well.

gel-cli-instance/src/cloud/instance.rs
```diff
@@ -120,16 +120,22 @@ impl<H: CloudHttp> InstanceBackup for CloudInstanceBackup<H> {
...
+                    size: None,
+                    location: None,
                 })
                 .collect())
         })
         .map(map_join_error::<_, CloudError>)
         .boxed()
     }
+
+    fn get_backup(&self, _backup_id: &BackupId) -> anyhow::Result<Backup> {
```
mmastrac, line 132: Should this use list_backups for now until the cloud API exists?

gel-cli-instance/src/instance/backup.rs
mmastrac: Also: note that we could return the full `Backup` from the `backup` method as well to make this a single operation.

scotttrinh:
> > I wonder if we should omit the full path to the backup and just list the backup ID, with a command to restore it?
>
> wdyt?

I think the point of showing the path was to emphasize that this is a local file and there is a stable location where you can go find your backups. If that seems noisy (it doesn't feel too bad to me personally) here are some ideas we could explore:

1. Just show the path to the backup directory itself in the initial message: "Created incremental backup at {backup_location}"
2. Allow restore to take a full path
3. Add a `-v` flag

Personally, I think it's fine as is.

mmastrac (approved):

PR #1633, opened 2025-05-28
Add some more watch tests.

vpetrovykh:
Test that watch locks the project so that branch and migration editing commands cannot be executed.

Test that restoring an instance during active watch cannot be done.

Test that switching git branch during active watch terminates the watch.


fantix (approved):

fantix:
I assume we still need this one?

PR #1641, opened 2025-06-02
Add `--extend-gel-toml` to `gel watch` command

fantix:
This option allows specifying an extra TOML file with additional hooks to extend the default `gel.toml`.

Before hooks will run in order: `gel.toml` -> extended TOML
After hooks will run in order: extended TOML -> `gel.toml`

This is to support (3) in geldata/gel-python#647


mmastrac (approved):
Looks good, but could we add a CLI test?

PR #1643, opened 2025-06-02
Support CI=true (partially)

fantix:
It's partial because `--non-interactive` is not yet fixed to honor `CI=true`.

### DIFF
--- a/src/cli/env.rs
+++ b/src/cli/env.rs
@@ -135,6 +135,14 @@ define_env! {
     /// Whether we are running in a hook
     #[env(_GEL_IN_HOOK)]
     in_hook: BoolFlag,
+
+    /// Whether we are running in a CI environment
+    #[env(CI)]
+    in_ci: BoolFlag,
+
+    /// The auto backup mode
+    #[env(GEL_AUTO_BACKUP_MODE)]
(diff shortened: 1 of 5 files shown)


src/cli/env.rs
```diff
@@ -218,3 +218,9 @@ impl std::ops::Deref for BoolFlag {
         &self.0
     }
 }
+
+static IS_CI: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
```
mmastrac, line 222: We have `Env` in `env.rs` where you can add this. We shouldn't add random env vars around.

src/migrations/apply.rs
```diff
@@ -216,15 +216,20 @@ impl AutoBackup {
             );
             return Ok(None);
         }
+        let mut skipped = crate::cli::env::is_ci();
         match std::env::var("GEL_AUTO_BACKUP_MODE") {
```
mmastrac, line 220: This should also live in `env.rs`, missed in a previous review.

mmastrac:
Note that we'll probably need to set CI to "false" before running the new tests, since the tests are mostly wanting to test how the CLI works in non-CII cases.

PR #1645, opened 2025-06-02
Add some hints to backup actions

### LINKED ISSUE #1644: add a hint to `gel instance list-backups`
Currently:

Image

I'd add one sentence below the table: `To restore to a certain backup point run "gel instance restore <ID>".`

Also we should add `See the full backup list with "gel instance list-backups".` to our logging -- when we make an automatic backup before the migration -- it should be in the log message so that people know how to use the system without googling.

This should be a quick change, I hope

fantix:
Fixes #1644 

image

image

### DIFF
--- a/src/migrations/apply.rs
+++ b/src/migrations/apply.rs
@@ -241,7 +241,14 @@ impl AutoBackup {
                     }
 
                     if !quiet {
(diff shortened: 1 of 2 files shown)


scotttrinh (approved):
🚢

mmastrac (approved):

PR #1646, opened 2025-06-02
Make `gel init` non-interactive by default

fantix:
1. `gel init` is now non-interactive by default
2. So is `gel project init`
3. `--non-interactive` on the `init` command is now a no-op
4. To recover old behavior, use `gel init --interactive`


scotttrinh (approved):
I trust that this was mostly a find and replace mission here, and the logic seems sound.

PR #1650, opened 2025-06-05
Change some more strings to "gel"

msullivan:


mmastrac (approved):
We should try to use the branding strings but this is better than using the out of date name

mmastrac:
I'll fix that build bustage shortly..

PR #1651, opened 2025-06-05
Fix `gel watch` to cope with running outside of project root

elprans:


mmastrac (approved):

mmastrac:
Rebasing on #1652 should fix things.

elprans:
@mmastrac, this fix actually doesn't work and I don't quite understand why, I'm not very familiar with the watcher code.  Can you lend a hand here please?

PR #1654, opened 2025-06-07
generate: Call `gel-generate-py` for python generators

elprans:
Also, add support for calling `uv run`, controlled by the
`GEL_GENERATE_USE_UV` environment enum, where `auto` is the default --
search for `gel-generate-py` in `PATH` and run it, if not found, try
`uv run gel-generate-py`.  The `always` setting always attempts `uv` and
`never` disables the use of `uv`.

### DIFF
--- a/src/commands/generate.rs
+++ b/src/commands/generate.rs
@@ -1,9 +1,12 @@
 use std::env;
+use std::ffi::OsString;
 use std::io::Write;
 use std::process::Stdio;
 
+use anyhow::Context;
 use tempfile::NamedTempFile;
 use tokio::process;
+use which::which;
 
 use crate::commands::options::Options;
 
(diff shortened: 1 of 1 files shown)


src/commands/generate.rs
```diff
@@ -13,20 +16,76 @@ pub struct Command {
...
-        Some(v) => v,
-        None => anyhow::bail!("Generator should be of form <lang>/<tool>"),
-    };
+/// How we should invoke via `uv run` (or not).
+#[derive(Debug, Clone, Copy)]
+enum UseUv {
+    Always,
+    Auto,
+    Never,
+}
```
msullivan, line 25: Should we also make this a command-line flag? Not urgent.

src/commands/generate.rs
```diff
@@ -13,20 +16,76 @@ pub struct Command {
...
+                }
+            }
         }
-        // XXX: or should we use `uv run`?
-        vec!["gel-py"]
     } else if lang == "js" {
         vec!["npx", "@gel/generate", generator]
+            .into_iter()
+            .map(OsString::from)
+            .collect()
```
msullivan, line 88: rust! I left it as `Vec<&str>` because I didn't want to have to do this annoyingly verbose coercion, but also knew that any change might end up requiring a type change...

(Though I guess it wouldn't if we put the non-resolved names in the Vec)

msullivan (approved):

fantix (approved):

PR #1666, opened 2025-06-20
Unmute hook output

fantix:
This changes the color of hooks output from gray (muted) to plain/original


mmastrac (approved):
LGTM - should the proxy formatter receive a flag saying whether the string is stderr/stdout though?

PR #1675, opened 2025-06-27
Creating indexes with `build_concurrently`

aljazerzen:
Ref https://github.com/geldata/gel/pull/8747 and
https://github.com/geldata/gel/issues/8175

This is the command output:

```
Applying m174z6shgcp4npgznerch2iiyppimoi3ybrl3refoe3ruzbj65m7wa (00001-m174z6s.edgeql)
... parsed
... applied
Applying m1zieaot5h3ugqofzch2exrzxbs2mijqidzcbjgvuvs2356wzs6saa (00002-m1zieao.edgeql)
... parsed
... applied
Applying m1dhu7b2uuqssjt2i576s7enjs2qikyiomcgug4bjr35seddphxdkq (00003-m1dhu7b.edgeql)
... parsed
... applied
Activating index on 'default::B' with expr '.z'
... done
```

I've also added `--no-index-activate` flag:

```
      --no-index-activate
          Disable creation of concurrent indexes after applying migrations.

          Indexes that set `create_concurrently` to `true` are not activated when migration is
          applied. Instead, they are activated right after, unless this flag is set. Index
          activation will not acquire any locks that would prevent concurrent reads or writes of
          objects.
```

### DIFF
--- a/src/migrations/apply.rs
+++ b/src/migrations/apply.rs
@@ -840,11 +840,16 @@ async fn disable_ddl(conn: &mut Connection) -> Result<(), anyhow::Error> {
 
 async fn index_build_concurrently(conn: &mut Connection) -> Result<(), anyhow::Error> {
     let version = conn.get_version().await?;
-    let is_supported = version.specific()
-        >= ver::Specific {
-            major: 6,
-            minor: ver::MinorVersion::Minor(10),
-        };
+
+    // supported on 7.0-dev.9640 onward
+    let is_supported = if version.specific().major == 7 {
+        match version.specific().minor {
+            ver::MinorVersion::Dev(m) if m < 9641 => false,
+            _ => true,
+        }
+    } else {
+        version.specific().major > 7
+    };
     if !is_supported {
         return Ok(());
     }
--- a/src/migrations/apply.rs
+++ b/src/migrations/apply.rs
@@ -73,13 +73,13 @@ pub struct Command {
     #[arg(long)]
     pub dev_mode: bool,
 
-    /// Disable creation of concurrent indexes after applying migrations.
+    /// Disable building of concurrent indexes after applying migrations.
     ///
-    /// Indexes that set `create_concurrently` to `true` are not activated when migration is applied.
-    /// Instead, they are activated right after, unless this flag is set. Index activation will
+    /// Indexes that set `build_concurrently` to `true` are not built when migration is applied.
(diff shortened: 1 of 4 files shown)


src/migrations/apply.rs
```diff
@@ -819,6 +838,53 @@ async fn disable_ddl(cli: &mut Connection) -> Result<(), anyhow::Error> {
     Ok(())
 }
 
+async fn index_activate_concurrently(conn: &mut Connection) -> Result<(), anyhow::Error> {
+    let version = conn.get_version().await?;
+    let is_supported = version.specific()
+        >= ver::Specific {
+            major: 6,
+            minor: ver::MinorVersion::Minor(10),
+        };
```
msullivan, line 847: We aren't going to backport, so we shouldn't do a 6.10 version. Let's put in the actual nightly number once it is built.

src/migrations/apply.rs
```diff
@@ -192,6 +206,11 @@ pub async fn run(
         }
     }
     apply_migrations(conn, migrations, &ctx, cmd.single_transaction).await?;
+
+    if !cmd.no_index_activate {
```
msullivan, line 210: Let's change `activate` to `build` everywhere

src/migrations/apply.rs
```diff
@@ -819,6 +838,53 @@ async fn disable_ddl(cli: &mut Connection) -> Result<(), anyhow::Error> {
...
+        subject_name: String,
+    }
+
+    let inactive_indexes: Vec<IndexShort> = conn
+        .query(
+            "
+            select schema::Index {
+              id, expr, subject_name := .<indexes[is schema::ObjectType].name
+            }
+            filter .create_concurrently and not .active
```
msullivan, line 865: I've changed this to `build_concurrently`

src/migrations/apply.rs
```diff
@@ -819,6 +838,53 @@ async fn disable_ddl(cli: &mut Connection) -> Result<(), anyhow::Error> {
...
+        .await?;
+
+    for index in inactive_indexes {
+        print::msg!(
+            "Activating index on '{}' with expr '{}'",
+            index.subject_name.emphasized(),
+            index.expr
+        );
+        conn.execute(
+            &format!("administer concurrent_index_create(\"{}\")", index.id),
```
msullivan, line 878: And this to `concurrent_index_build`, with a required `<uuid>` cast

msullivan (approved):

PR #1678, opened 2025-06-27
Fix for broken backup setups in 6.x 

### LINKED ISSUE #1677: CLI should check for replication permission in pg_hba.conf
I'm not sure why, but some 6.x release installations are missing the second line in pg_hba.conf. The CLI should check to see that it exists before performing an automatic backup.

```
local all postgres trust
local replication postgres trust
```

mmastrac:
Closes #1677

### DIFF
--- a/gel-cli-instance/src/local/localbackup.rs
+++ b/gel-cli-instance/src/local/localbackup.rs
@@ -48,6 +48,7 @@ impl LocalBackup {
             ));
         }
 
+        // Some 6.x releases appear to have a broken pg_hba.conf. Check for that and disable backups.
         let pg_hba = handle.paths.data_dir.join("pg_hba.conf");
         let Ok(pg_hba_contents) = std::fs::read_to_string(&pg_hba) else {
             return Err(InstanceOpError::Unsupported(
--- a/gel-cli-instance/src/local/localbackup.rs
+++ b/gel-cli-instance/src/local/localbackup.rs
@@ -2,6 +2,7 @@ use std::{
     fs::File,
     path::{Path, PathBuf},
     process::Command,
+    sync::Arc,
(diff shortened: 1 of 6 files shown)


gel-cli-instance/src/local/localbackup.rs
```diff
@@ -30,13 +31,40 @@ use super::LocalInstanceHandle;
...
+        // Some 6.x releases appear to have a broken pg_hba.conf. Check for that and disable backups.
+        let pg_hba = handle.paths.data_dir.join("pg_hba.conf");
+        let Ok(pg_hba_contents) = std::fs::read_to_string(&pg_hba) else {
+            return Err(InstanceOpError::Unsupported(
+                "older releases: pg_hba.conf not found or unreadable".to_string(),
+            ));
+        };
+        if !pg_hba_contents.contains("local replication postgres trust") {
+            return Err(InstanceOpError::Unsupported(
+                "older releases: pg_hba.conf does not allow local replication connections (add `local replication postgres trust` to the file)".to_string(),
```
scotttrinh, line 60: Should we consider doing this ourselves? Maybe at least printing the path to the file would be helpful.

mmastrac: I considered doing out ourselves, but restarting postgres is tricky. I can print out the full path though.

mmastrac: New error:

```
Automatic backup is disabled.
Unsupported operation for older releases, pg_hba.conf does not allow local replication connections (add `local replication postgres trust` to "/home/matt/.local/share/edgedb/data/dump_restore2/pg_hba.conf")
```

scotttrinh (approved):

PR #1684, opened 2025-07-03
`gel configuration apply` and `gel.local.toml`

fantix:
Sample TOML:

```toml
[local.config]
cors_allow_origins = ["https://xxx", "https://yyy"]
allow_bare_ddl = "AlwaysAllow"

[[local.config."ext::auth::EmailPasswordProviderConfig"]]
require_verification = true
```

Executes like:

```
$ gel init
Found `edgedb.toml` in /Users/fantix/workspace/edgedb-cli
Synchronizing project...
Executing query: configure current branch set cfg::cors_allow_origins := {
	"https://xxx",
	"https://yyy"
};
	with args: {}
Executing query: configure current branch set cfg::allow_bare_ddl := <cfg::AllowBareDDL>"AlwaysAllow";
Executing query: configure current branch reset ext::auth::EmailPasswordProviderConfig;
Executing query: configure current branch insert ext::auth::EmailPasswordProviderConfig {
	require_verification := true
};
```

High priority:
- [x] Schema of config and TOML compilation/configuration
- [x] Remove config parameterization which doesn't work, just use injection for everything
- [ ] Add config description and sample value
- [x] Generate default gel.local.toml

Others:
- [x] Handle nested objects that should be compiled into an insert subquery instead of global configure...insert
- [ ] ~Add cache for search_schema_name()~
- [ ] Use JSON schema to replace hard-coded schema
- [ ] Add schema introspection 
- [ ] Add schemas for different versions of Gel
- [ ] Manage schema resources through remote HTTP
- [x] Add tests


aljazerzen:
@fantix Did you mean for `Schema` to be able to serialize into the initial `gel.local.toml`? Or did you plan on hard-coding it?

aljazerzen:
I've:
- fixed the "config objects are not being reset" (we composed the reset query, but never executed it).
- refactored the code a lot, it's now in multiple files, hopefully segmented in a sensible way.
- made is so that `gel init` now always applies `gel.local.toml` config to the database and if the file does not exist, creates a new one from hard-coded example.
- also improved messages a bit.

Found problems:
- some config options should be applied to the instance instead of the database: `gel error: ConfigurationError: 'http_max_connections' is a system-level configuration parameter; use "CONFIGURE INSTANCE"`,
- `gel error: ext::auth::MagicLinkProvider, Caused by: missing tname` ???

aljazerzen:
@fantix I've closed the PR that I've opened and just pushed my commits here.

aljazerzen:
Yes, we can add `--using-extension auth` to init, but in that case we must check that a new project is being initialized. Because `gel init` is also being used for linking/creating instance for projects. And also, now, to apply config. And I don't want to be editing schema files that are not empty.

fantix:
Awesome, thank you so much @aljazerzen ! LGTM

PR #1688, opened 2025-07-15
Fix inserting non-root-config singleton objects

fantix:
This fixes compiling this config:

```toml
# UI customization
[local.config."ext::auth::UIConfig"]
redirect_to                        = "http://localhost:8000/auth/callback"
redirect_to_on_signup              = "http://localhost:8000/auth/callback?isSignUp=true"
```

Also fixed:
 * Schemaless string literal quoting
 * Nested insert subquery, so this will be possible:

```toml
[[local.config.auth]]
priority = 100
user = ["gel"]
method = { _tname = "cfg::JWT", transports = ["HTTP"] }

[[local.config.auth]]
priority = 200
user = ["admin", "gel"]
method = { _tname = "cfg::SCRAM", transports = ["TCP", "HTTP"] }
```

(but no schema added yet, because `auth` is a system config)

### DIFF
--- a/src/portable/project/config/mod.rs
+++ b/src/portable/project/config/mod.rs
@@ -1,6 +1,7 @@
 mod schema;
 mod validation;
 
+use edgeql_parser::helpers::quote_string as ql;
 use gel_protocol::value::Value as GelValue;
 use indexmap::IndexMap;
 use std::collections::HashMap;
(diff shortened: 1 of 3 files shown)


fantix:
> > (but no schema added yet, because auth is a system config)
> 
> What's system config? Options that cannot be applied to branches, but only to whole instance?

Exactly, the ones with the `cfg::system := 'true'` annotation.

> I wonder if we should have [instance] and [branch] top-level tables just to be super clear about where and how the config is being applied?

Yeah, that sounds like a good idea to me :+1:

PR #1692, opened 2025-07-23
Add `gel sync` command

fantix:
It does the following, so that user can repeat `gel sync` until success:

1. `gel migrate --quiet`
2. `gel migration create --non-interactive --quiet`
3. `gel configure apply` (on error, `gel sync` may edit `dbschema/extensions.gel`)


mmastrac (approved):
Looks good. We should consider adding a CLI test for it

fantix:
clitest is now added

PR #1697, opened 2025-07-25
Add help to `gel generate`

### LINKED ISSUE #1694: `generate --help` should show list of first-party generators
When you run `gel generate --help`, at the moment, we just say you need to pass `<GENERATOR>` which is a string as an argument. This doesn't give enough direction to the caller as to what they can/should pass here.

In scope with this change is to hardcode the first-party known values that we support here. In the future, we can implement some discovery mechanism so that the language bindings can publish their own available generators.

fantix:
Fixes #1694


scotttrinh (approved):
`js/prisma` 😅

PR #1699, opened 2025-07-27
Replace `uv run` with `uvx`

### LINKED ISSUE #1696: `gel generate py/*` should run from the root of a project, even if the Python project is nested
In the common monorepo scenario:

```
.
├── backend
│   ├── app
│   ├── pyproject.toml
│   └── tests
├── dbschema
│   └── default.gel
├── frontend
│   ├── package.json
│   ├── src
│   └── tests
└── gel.toml
```

[Using `uv run` as our command runner](https://github.com/geldata/gel-cli/blob/49675fa3109544ee4b46ad9beee93d3d0a1091b2/src/commands/generate.rs#L56-L57) currently cannot find the generate script from the root of this setup. It currently fails because the actual tool lives in `backend`. However, if we swap `uv run` for `uvx`, it will use the `gel-generate-py` script from the globally managed `gel` package that `uvx` manages.

fantix:
This PR replaces `uv run` with `uvx` completely:
* Renamed `GEL_GENERATE_USE_UV` to `GEL_GENERATE_USE_UVX`
* Added env var `GEL_PYTHON_VERSION_SPEC` to be used in `uvx --from gel${GEL_PYTHON_VERSION_SPEC} ...`, defaults to `>=4.0.0b1`

Fixes #1696 

Qs:

- [ ] Should we still keep the `uv run` path and detect smartly?
- [ ] Should we try to search for the proper `GEL_PYTHON_VERSION_SPEC` in `pyproject.toml` etc.?

### DIFF
--- a/src/commands/generate.rs
+++ b/src/commands/generate.rs
@@ -62,25 +62,25 @@ fn print_help(short_help: bool) -> anyhow::Result<()> {
     Ok(())
 }
 
-/// How we should invoke via `uv run` (or not).
+/// How we should invoke via `uvx` (or not).
 #[derive(Debug, Clone, Copy)]
-enum UseUv {
+enum UseUvx {
     Always,
     Auto,
     Never,
 }
 
-impl UseUv {
-    /// Read from the `GEL_GENERATE_USE_UV` env var, defaulting to `auto`
+impl UseUvx {
+    /// Read from the `GEL_GENERATE_USE_UVX` env var, defaulting to `auto`
     fn from_env() -> Self {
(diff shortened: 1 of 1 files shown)


mmastrac (approved):

fantix:
Update to `gel generate`:

1. Detect venv first, run `gel-generate-py` from venv/bin or fail if `gel-generate-py` doesn't exist (and suggest to install gel)
2. Detect uv project, run`uv run gel-generate-py` if found, which may fail
3. Nothing is detected, fail with hint ("enter your python project dir?")
4. No more `uvx` for now

PR #1702, opened 2025-07-29
Rename a few instances of edgedb to gel

aljazerzen:


src/migrations/upgrade_check.rs
```diff
@@ -96,7 +96,7 @@ pub fn upgrade_check(_options: &Options, options: &UpgradeCheck) -> anyhow::Resu
     // This is run from windows to do the upgrade check
     if let Some(status_path) = &options.run_server_with_status {
         let server_path = info.server_path()?;
-        let mut cmd = process::Native::new("edgedb", "edgedb", server_path);
+        let mut cmd = process::Native::new("gel-server", "gel-server", server_path);
```
mmastrac, line 99: Maybe better if we make this a branding string so we can track them down in the future.

PR #1704, opened 2025-07-30
In-place instance upgrade

### LINKED ISSUE #1519: Support performing in-place upgrades
We want to support in-place upgrades between major versions of the server.

Specifically, IPUs are supported between versions with different catalog versions, so we should be able to build and test it using upgrades between nightly versions.
(We should then I guess also support trivial in-place upgrades between nightlies with *the same* catalog version?)

https://github.com/geldata/gel/blob/master/dev-notes/inplace-upgrades.md is docs on the upgrades

aljazerzen:
Closes #1519

### DIFF
--- a/tests/scripts/upgrade.cli
+++ b/tests/scripts/upgrade.cli
@@ -40,8 +40,10 @@ choice {
 ! To connect to the instance run:
 !   gel -I inst3
 
-$ gel instance backup --instance=inst3 --non-interactive
-? Successfully created a backup [a-z0-9-]+ for Gel instance 'inst3'
+if TARGET_OS != "windows" {
+    $ gel instance backup --instance=inst3 --non-interactive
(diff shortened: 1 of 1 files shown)


aljazerzen:
Ah, it didn't even occur to me that the regex might be wrong. Thanks.

aljazerzen:
I've added tests for backup/restore. As I suspected, it should not be affected, since postgres (major) version does not change with in-place upgrade.

What kind of configuration did you have in mind?

---

Also, I'll not add backup as a part of upgrade right now. That's because database might be large and doing it implicitly might confuse users. They should do it manually, before the running upgrade.

mmastrac:
Great, thanks for adding those tests.

I'm not sure what's been happening but there are cases where instances fail to work with autobackup because pg_hba.conf hasn't been updated to include replication permissions. My current guess is that we're keeping the same postgres major version, but not re-building the pg_hba.conf file.

PR #1712, opened 2025-08-01
Fix uv project detection for dynamic versions

fantix:
`uv version` doesn't work with dynamic version (astral-sh/uv#14137) yet, while there's no simple way to discover uv project (astral-sh/uv#13636), so let's just parse ~the `uv run -v` debug log first, and fallback to `uv version` if the debug log format changes someday.~ the `uv version -v` debug log.

~Also added `--no-sync` flag to `uv run` calls, so that they would deal the minimal changes to the project dir. Unfortunately, `.venv` would still get created if it doesn't exist yet, even with the `--no-sync` flag.~

### DIFF
--- a/src/cli/env.rs
+++ b/src/cli/env.rs
@@ -205,9 +205,8 @@ define_env! {
     #[env(GEL_AUTO_BACKUP_MODE)]
     auto_backup_mode: AutoBackupMode,
 
-    /// How we should invoke via `uv run` (or not)
+    /// How we should detect uv project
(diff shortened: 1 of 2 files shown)


fantix:
Yeah, there's also debug info available in `uv version -v`, I'll use that instead.

(because the actual `uv run --no-sync gel generate` may still create `.venv` regardless of the existence of the Python tool, so we prefer to find the existing `.venv` and run `.venv/bin/gel-generate-py`)

mmastrac (approved):

PR #1715, opened 2025-08-01
Limit the versions we return in list-versions

mmastrac:
`gel server list-versions` is pretty noisy, so let's filter to showing just the latest stable for now and give users the ability to change the channel as needed.

### DIFF
--- a/src/portable/server/list_versions.rs
+++ b/src/portable/server/list_versions.rs
@@ -1,95 +1,186 @@
 use std::collections::BTreeMap;
 use std::path::PathBuf;
 
+use color_print::cprintln;
 use gel_cli_derive::IntoArgs;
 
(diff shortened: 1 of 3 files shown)


src/portable/server/list_versions.rs
```diff
@@ -1,96 +1,187 @@
...
+            "".to_string()
+        };
+
+        let installed = if cmd.installed_only {
+            "installed"
+        } else {
+            "all"
+        };
+
+        cprintln!("Listing <bold>{filter}</bold> versions of <bold>{BRANDING}</bold>:");
```
elprans, line 95: When listing only current stable versions, can we also inject a note on how to expand the list, e.g `use --all to list all available versions or --channel=<channel> to list versions available in another release channel`).

PR #1722, opened 2025-08-07
Upgrade a few deps

aljazerzen:
Reason for this now is that zip version we had in
Cargo.lock was yanked


mmastrac:
@aljazerzen I pushed a fix to master to fix the semantic merge conflict here.

PR #1725, opened 2025-08-08
Restore truncated channel for rust-toolchain

mmastrac:


mmastrac:
> .git-blame-ignore-revs

Good idea, I pushed that one right to master.

PR #1726, opened 2025-08-08
Add `config.update.*` hooks

fantix:
Such hooks are called before or after `gel.local.toml` config application.


src/hooks/mod.rs
```diff
@@ -65,6 +65,8 @@ fn get_hook<'m>(action: &'static str, hooks: &'m project::manifest::Hooks) -> Op
         "migration.apply.after" => &hooks.migration.as_ref()?.apply.as_ref()?.after,
         "schema.update.before" => &hooks.schema.as_ref()?.update.as_ref()?.before,
         "schema.update.after" => &hooks.schema.as_ref()?.update.as_ref()?.after,
+        "config.update.before" => &hooks.config.as_ref()?.update.as_ref()?.before,
```
mmastrac, line 68: I think we should update project3/gel.toml and hooks.cli with some tests.

aljazerzen (approved):

PR #1732, opened 2025-08-20
gel migrate create --explicit

mmastrac:
Implements a more verbose, much more explicit form of gel migrate create. We could consider making this the default and adding a `--compact` version.

```
did you alter the type of link 'foo' of object type 'default::Base'?

If so, the following DDL statements will be applied:

    ALTER TYPE default::Base {
        ALTER LINK foo {
            SET TYPE default::Child2 USING (\(cast_expr));
        };
    };

Select an action:

"yes" (or "y"): Confirm the prompt
"no" (or "n"): Reject the prompt; server will attempt to generate another suggestion
"confirmed" (or "c"): List already confirmed EdgeQL statements for the current migration
"back" (or "b"): Go back a step by reverting latest accepted statements
"stop" (or "s"): Stop and finalize migration with only current accepted changes
"quit" (or "q"): Quit without saving changes

Which action do you want to take? [y,n,c,b,s,q,?]

Please specify a conversion expression to alter the type of link 'foo'.
If left blank, the migration will use the expression:
.foo[IS default::Child2]

Conversion expression: .foo[IS default::Child2]
Created migration ...
```

### DIFF
--- a/src/migrations/create.rs
+++ b/src/migrations/create.rs
@@ -65,6 +65,10 @@ pub struct Command {
     /// This safe default can be overridden with `--allow-unsafe`.
     #[arg(long)]
     pub non_interactive: bool,
+    /// Run the migration creation in explicit mode. In this mode, the user
+    /// will receive more detailed prompts.
+    #[arg(long)]
+    pub explicit: bool,
     /// Apply the most probable unsafe changes in case there are ones. This
     /// is only useful in non-interactive mode.
     #[arg(long)]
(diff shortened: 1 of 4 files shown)


src/migrations/create.rs
```diff
@@ -978,13 +1004,48 @@ fn get_input(req: &RequiredUserInput) -> Result<String, anyhow::Error> {
     }
 }
 
+fn get_input_explicit(req: &RequiredUserInput) -> Result<String, anyhow::Error> {
+    let prompt = "Conversion expression:";
```
1st1, line 1008: Maybe we have a more detailed message here (use your judgement) -- like 

* `Please specify an EdgeQL expression that will be used to populate the existing optional fields in the database to become "required"`

* or `Please specify an EdgeQL expression that will be used to change existing data in property "foo" from "std::int64" to "std::str"`

* etc?

PR #1735, opened 2025-08-25
Tweak the rules for in-place upgrades

msullivan:
Disable doing an IPU to 6.0-beta.1, because it has some bad brokenness
with the auth extension.

Also, only default to IPUs for RCs and release versions, not for other
pre-releases.

Maybe we should never default to it yet?


aljazerzen (approved):

aljazerzen:
We should default to it, because if we don't we will never really know it works.

PR #1752, opened 2025-11-07
Fix extension install

msullivan:
More nested tokio runtime fixes. Follow up to #1747.


aljazerzen:
I'm pretty sure this worked when I tweaked it, although I'm not 100%. It's also possible that repository state changed such that we are now hitting different code paths.

---

I'm looking for tests now and it seems we have none for extensions.

